PR #82, opened 2011-11-15
Patches related to SQS transport

dahlia:
### [af7bef294f3b4140e03beb3fabb37f210a2903aa] Proper handling of SQS queue name format

Unlike [`str.translate()`](http://docs.python.org/library/stdtypes.html#str.translate) method `unicode.translate()` method cannot take table encoded by [`string.maketrans()`](http://docs.python.org/library/string.html#string.maketrans) function. So I replaced `CHARS_REPLACE_TABLE` with a `dict` table of Unicode ordinals.

Before patched this, it could occur error like following in some cases:

```
  File "/Users/dahlia/Projects/styleshare-env/lib/python2.7/site-packages/gevent-0.13.6-py2.7-macosx-10.7-intel.egg/gevent/greenlet.py", line 390, in run
    result = self._run(*self.args, **self.kwargs)
  File "/Users/dahlia/Projects/styleshare-env/lib/python2.7/site-packages/kombu-1.4.3-py2.7.egg/kombu/transport/virtual/__init__.py", line 359, in queue_declare
    self._new_queue(queue, **kwargs)
  File "/Users/dahlia/Projects/styleshare-env/lib/python2.7/site-packages/kombu-1.4.3-py2.7.egg/kombu/transport/SQS.py", line 165, in _new_queue
    self.entity_name(queue),
  File "/Users/dahlia/Projects/styleshare-env/lib/python2.7/site-packages/kombu-1.4.3-py2.7.egg/kombu/transport/SQS.py", line 157, in entity_name
    return safe_str(name).translate(table)
TypeError: character mapping must return integer, None or unicode
```
### [d54737cdc909583c59e2b7c95b96bbdadbb3beb3] Made SQS able to be configured by URL

[`urlparse.urlparse()`](http://docs.python.org/library/urlparse.html#urlparse.urlparse) normalizes URL schemes into all lowercases, so SQS transport cannot be resolved by URL-style configuration e.g. `SQS://...`. Added a lowercase version of alias for `SQS` as well to resolve SQS transport by URL-style configuration e.g. `sqs://...`.

### DIFF
--- a/kombu/transport/__init__.py
+++ b/kombu/transport/__init__.py
@@ -76,6 +76,7 @@ TRANSPORT_ALIASES = {
     "memory": "kombu.transport.memory.Transport",
     "redis": "kombu.transport.pyredis.Transport",
     "SQS": "kombu.transport.SQS.Transport",
+    "sqs": "kombu.transport.SQS.Transport",
     "beanstalk": "kombu.transport.beanstalk.Transport",
     "mongodb": "kombu.transport.mongodb.Transport",
     "couchdb": "kombu.transport.pycouchdb.Transport",


cdram:
f

PR #148, opened 2012-07-27
Initial ZeroMQ Transport

dctrwatson:
Currently, the ZeroMQ transport does not provide routing via queues. The
queue name is sent along with the message, but dropped upon consumption.

Messages sent to direct/topic exchanges are sent round-robin to connected
peers regardless of the designated queue. However, since fanout queues
are based on PUB/SUB, some peers don't have to subscribe to all topics.


etseidler:
hi. i know this is _way_ after the fact, but i was wondering: why zmq was removed as a transport option in the latest major version of kombu? i have a [flask-socketIO](https://github.com/miguelgrinberg/Flask-SocketIO) app with zmq integrations, and I was really hoping to use zmq as a transport for the messaging queue that is needed to facilitate socketIO emits on non-server processes. the author of flask-socketIO suggested that I try using [this sample app](https://github.com/miguelgrinberg/flack) first with redis as a messaging queue to prove non-server socketIO emits work at all (which they do with redis). i attempted to switch over to zmq as a transport, but did not get things working. do you have any code snippets of using zmq as a kombu transport? any other suggestions?

Issue #173, opened 2012-11-09
SQS Long Polling


23doors:
SQS today announced long polling - which may finally mean that SQS becomes cost efficient and fast instead of expensive (due to a lot of repeated API calls that returned no messages), slow (time reaction dependent on how often polls were done.

From API perspective it all comes down to one new param for get_messages (ReceiveMessage) - WaitTimeSeconds. It's easy with boto to add this param:

instead of queue.get_messages(number_messages)

```
    params = {'MaxNumberOfMessages' : number_messages, 'WaitTimeSeconds': 20}
    messages = queue.connection.get_list('ReceiveMessage', params,
                         [('Message', queue.message_class)],
                         queue.id, queue)
```

WaitTimeSeconds is a configurable value between 1 and 20, but no reason to not use max value.

For queues created by Kombu, this value should be set on creation by:

```
queue.set_attribute('ReceiveMessageWaitTimeSeconds', 20)
```

I will submit an issue for botu to support get_messages new param, although it would be great if celery could make use of it already using methods I provided.

For reference on this topic:
http://docs.amazonwebservices.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-long-polling.html
http://docs.amazonwebservices.com/AWSSimpleQueueService/latest/APIReference/Query_QueryReceiveMessage.html

PR #198, opened 2013-01-11
Add long-polling support to SQS transport

jamesls:
We've just released a [new version of boto](http://pypi.python.org/pypi/boto/2.7.0), which adds long
polling support for SQS.

#173 already has a discussion about adding long polling support
to the SQS transport, which is what I've implemented here.

The polling interval was set to 0 and a new
transport option (wait_time_seconds) was added.  This
parameter specifies how long to wait for a message from
SQS, and defaults to 20 seconds, which is the maximum
value currently allowed by SQS.

For testing, I verified that the tests in `funtests/tests/test_SQS.py`
still pass. I also verified that I see the appropriate API calls
being made to SQS as well as 20 seconds between the `get_messages`
calls.

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -121,6 +121,8 @@ class Channel(virtual.Channel):
 
     default_region = 'us-east-1'
     default_visibility_timeout = 1800  # 30 minutes.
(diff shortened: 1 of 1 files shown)


ask:
SQS in master uses long polling and is rewritten to use async I/O.

PR #235, opened 2013-05-22
a callback function when Consumer.callbacks raises an exception

ttddyy:
Added `on_callback_error` on Consumer. The registered function is invoked when one of the consumer callback functions have failed(throw exception). When True is returned from the function, it will continue the rest of consumer callback functions. Otherwise, stop and raise the originally thrown exception.

I think with this global error function, queue consumer applications can cleanly separate main logic and error handling logic.

### DIFF
--- a/kombu/messaging.py
+++ b/kombu/messaging.py
@@ -326,6 +326,21 @@ class Consumer(object):
     #: that occurred while trying to decode it.
     on_decode_error = None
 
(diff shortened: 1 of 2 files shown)


wowkin2:
Looks like this issue uses Milestone that is already release, propose to update to appropriate.

auvipy:
thanks

PR #304, opened 2014-01-16
Added await_data param to tailable cursor in mongodb transport layer.

hellerstanislav:
If the cursor does not have any data, blocks for a while instead
of returning no data. Now we get more 'streaming' behaviour rather
than getting blocks of messages once per poll unit.


ask:
MongoDB as a broker is no longer supported as we don't have the resources to maintain it :(

PR #315, opened 2014-02-10
[WIP] Add a transport for OpenStack Marconi

alex:
This is still very early, and I think needs a bunch of work. But it now passes all the functional tests, so I'm ready to start getting feedback from other people :-)


auvipy:
renamed to https://github.com/openstack/zaqar

flaper87:
I don't think this is needed/desired anymore.

PR #352, opened 2014-05-07
revoke support for djkombu

danni:
This is a somewhat nasty way of implementing the revoke support for django-celery without needing to handle broadcasting. We just drop the jobs from the queue and tell Celery they were revoked.


ask:
Using SQL as a broker is no longer supported as we don't have the resources to support it :(

PR #372, opened 2014-06-20
Add a requeue_log_error method

harlowja:
In order to match how there is a ack_log_error
and a reject_log_error it seems appropriate to
also provide a requeue_log_error method that
is provided in the message class.


auvipy (changes requested):
please rebase and cross check the suggestion

PR #373, opened 2014-06-24
Add support for RabbitMQ extensions for librabbitmq.

rogerhu:
Requires PR https://github.com/celery/librabbitmq/pull/49 to enable extension support.

It can be tested by the following:

```
from kombu import Connection

conn = Connection(host="host", userid="user",
                  password="password", virtual_host="vhost",
                  client_properties={'capabilities': {'authentication_failure_close': True}})

conn.connect()
```

Normally an error message of "_librabbitmq.ConnectionError: Couldn't log in: a socket error occurred" is given.  If this capability is enabled, an error message of "_librabbitmq.ConnectionError: Couldn't log in: unexpected method received" is received (this part should be improved).

Open to making pyamqp with the same support but that can be a separate PR.

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -137,7 +137,7 @@ class Connection(object):
                  ssl=False, transport=None, connect_timeout=5,
                  transport_options=None, login_method=None, uri_prefix=None,
                  heartbeat=0, failover_strategy='round-robin',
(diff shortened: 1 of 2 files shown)


thedrow:
@auvipy Do we need this PR?
It's for 3.0 which is very old.

Issue #382, opened 2014-07-23
Django Transaction Behavior Changing


fongandrew:
Using the Django transport introduces the following deprecation warning due to changes in Django's transaction handling.

    /usr/local/lib/python2.7/dist-packages/kombu/transport/django/managers.py:50: RemovedInDjango18Warning: commit_manually is deprecated in favor of set_autocommit.
      @transaction.commit_manually

ask:
Using SQL as a broker is no longer supported, as we don't have the resources to maintain it :(

PR #383, opened 2014-07-23
Update to use new Django transaction handling

fongandrew:
See https://docs.djangoproject.com/en/dev/topics/db/transactions/. Changes were introduced as of Django 1.6. Old functions are due to be deprecated in Django 1.8.

Fixes https://github.com/celery/kombu/issues/382.


ask:
Using SQL as a broker is no longer supported as we don't have the resources to maintain it :(

PR #386, opened 2014-08-03
Fix MongoDB Broadcast Cursor when empty

benjamin9999:
Tailable cursors do not remain alive (and block in next()) if they
are empty.

An example of this behavior:
http://stackoverflow.com/questions/10321140/using-pymongo-tailable-cursors-dies-on-empty-collections

Adding an outer find() loop appears to be non-trivial since the
cursor init/find() and poll/next() are decoupled, so the
BroadcastCursor has no way to restart itself.

### DIFF
--- a/kombu/transport/mongodb.py
+++ b/kombu/transport/mongodb.py
@@ -262,6 +262,11 @@ class Channel(virtual.Channel):
         self._open()
         self._ensure_indexes()
 
+        # tailable cursors won't block if empty.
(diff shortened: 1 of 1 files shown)


ask:
Using MongoDB as a broker is no longer supported as we don't have the resources to maintain it :(

benjamin9999:
I saw the removal, but the commit did not indicate why.  Is there somewhere to read more detail about the removal and its reasoning?

PR #437, opened 2014-12-11
Connect to a random host in a multi-broker config

ginoledesma:
Add a feature to Kombu that allows it to connect to a random host in a list of
brokers instead of always using the first node. This allows kombu to distribute
connections to a cluster of brokers and minimize overloading any given broker.

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import
 
 import os
 import socket
+import random
 
 from contextlib import contextmanager
 from itertools import count, cycle
@@ -84,6 +85,9 @@ class Connection(object):
     :keyword heartbeat: Heartbeat interval in int/float seconds.
         Note that if heartbeats are enabled then the :meth:`heartbeat_check`
         method must be called regularly, around once per second.
+    :keyword random_host_on_connect: Connect to a random host instead of the
+      first in the URL if multiple brokers are specified. This does not
+      affect the failover strategy.
 
     .. note::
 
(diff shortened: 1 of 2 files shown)


ask:
Thanks!

Looks good, but there is already a way to do this in that you can set `failover_strategy='shufflecycle'`!

That uses shuffle, but you could also implement what you have here by doing:

`Connection(failover_strategy=random.choice)`

MrTomWhite:
@ginoledesma This is a nice PR!
Thanks

It's a bit unfortunate that the functionality was already in kombu. Do you mind if the PR is closed?

 If you're still willing to help or want some fun problems to solve, there's still plenty of [issues](https://github.com/celery/kombu/issues) you could pick up. :smiley_cat:

ginoledesma:
Gladly! This is PR can be closed and I'll see what issues I can help with.

joekohlsdorf:
For everyone else who is confused I just wanted to comment that setting the `failover_strategy` is not enough.

The first connection will always be opened to the first host in the list because a new host is only picked from the cycle when retrying a connection (calling `maybe_switch_next()`).

auvipy:
which version are you using?

joekohlsdorf:
Happens on master.

jason-kane:
I'm +1 on this, I just had a bad day from a bunch of identically configured nodes thundering connections at the same rabbit node.  I'm tempted to patch in a random.shuffle(alt) right after https://github.com/celery/kombu/blob/master/kombu/connection.py#L174

joekohlsdorf:
@jason-kane As a workaround you could implement round-robin of all your hosts at the DNS level, depending on your DNS server it should return them in random order.

auvipy (changes requested):
would be really great if you can rebase

PR #442, opened 2015-01-16
Added expiration time for each publishing message

andronat:


ask:
Thanks! We should probably convert float seconds to milliseconds here, as that's what is usually used in Python.

andronat:
OK, I am on it.

andronat:
Travis fails but I think it is not due to my changes

andronat:
Any news on the subject?

ask:

PR #444, opened 2015-01-21
Fix message priority conversion for Redis and Beanstalk transports

daevaorn:
[AMQP protocol](http://www.amqp.org/sites/amqp.org/files/amqp.pdf) defines message priority field as:

>This field contains the relative Message priority. **Higher numbers indicate higher priority** Messages. Messages with higher priorities MAY be delivered before those with lower priorities.

but current Redis and Beanstalk transports relay on opposite logic. 

Redis transport iterates over its priority steps from lower ones:

    for pri in PRIORITY_STEPS:
        item = client.rpop(self._q_for_pri(queue, pri))
so low priority values are handled in the first place.

[Beanstalk protocol](https://raw.githubusercontent.com/kr/beanstalkd/master/doc/protocol.txt) also says:

> <pri> is an integer < `2**32`. **Jobs with smaller priority values will be scheduled before jobs with larger priorities**. The most urgent priority is 0; the least urgent priority is 4,294,967,295.

### DIFF
--- a/kombu/tests/transport/test_redis.py
+++ b/kombu/tests/transport/test_redis.py
@@ -103,18 +103,18 @@ class Client(object):
             raise Empty()
 
     def brpop(self, keys, timeout=None):
-        key = keys[0]
-        try:
-            item = self.queues[key].get(timeout=timeout)
-        except Empty:
-            pass
-        else:
-            return key, item
+        for key in keys:
+            try:
+                item = self.queues[key].get_nowait()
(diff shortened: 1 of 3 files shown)


ask:
Thanks! Must have missed this part (and they are *wrong* :).  This won't actually be backwards compatible though, so not sure how to introduce the change.

ask:
Merged into master, so we can get this in 4.0!

PR #447, opened 2015-02-02
Add additional string to check for when connecting to qpid

beav:
When we connect to qpid, we need to ensure that we skip to the next SASL
mechanism if the current mechanism fails. Otherwise, we will keep retrying the
connection with a non-working mech.

This patch adds an additional string that is returned when PLAIN is not
available, but python-saslwrapper is installed.


beav:
@bmbouter this is the patch for the issue we discussed on IRC

bmbouter:
LGTM, merging to 3.0 branch. @beav thanks!

PR #448, opened 2015-02-02
Qpid raises a NotFound exception when a Queue doesn't exist

bmbouter:
Adjusts Qpid to use qpid.messaging exceptions instead of AMQP exceptions. The Qpid transport doesn't raise many exceptions so this is a pretty small fix. There were some lines that were too long, and a few variables and test names have been fixed up for clarity.

Now the behavior is like this:

```
>>> from kombu import Connection, Queue
>>> with Connection('qpid://localhost:5672') as connection:
...     with connection.channel() as channel:
...         q = Queue('dog', channel=channel)
...         q.purge()
... 
  File "<stdin>", line 4, in <module>
  File "/home/bmbouter/kombu/kombu/entity.py", line 583, in purge
    nowait=nowait) or 0
  File "/home/bmbouter/kombu/kombu/transport/qpid.py", line 798, in queue_purge
    return self._purge(queue)
  File "/home/bmbouter/kombu/kombu/transport/qpid.py", line 516, in _purge
    raise NotFound(code=404, text=error_text)
qpid.messaging.exceptions.NotFound: NOT_FOUND - no queue 'dog'(404)
```

### DIFF
--- a/kombu/tests/transport/test_qpid.py
+++ b/kombu/tests/transport/test_qpid.py
@@ -14,14 +14,15 @@ from functools import wraps
 from mock import call
 
 from kombu.five import Empty, keys, range, monotonic
-from kombu.transport.qpid import AuthenticationFailure, QoS, Message
-from kombu.transport.qpid import QpidMessagingExceptionHandler, Channel
-from kombu.transport.qpid import Connection, ReceiversMonitor, Transport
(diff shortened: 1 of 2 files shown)


beav:
@bmbouter  this looks good to me

PR #450, opened 2015-02-06
Adds pre 3.0 error classes and removes a custom Exception handler

bmbouter:
This removes the QpidMessagingExceptionHandler and adds a simple exception catch to the one place that it was used. Also celery uses the pre-3.0 kombu error classes so this transport needs to continue to support those.

With this change I tested the following scenarios:
1) celery starts and waits for qpidd and successfully connects when it becomes available
2) celery is connected and qpidd dies. celery waits with the expected backoff behavior and when qpidd becomes available again it reconnects. This can happen any number of times stably.
3) same as (2) but qpidd is gracefully stopped and started
4) celery cannot authenticate. This is raised as an unrecoverable exception and it exits. That is the expected behavior.

### DIFF
--- a/kombu/tests/transport/test_qpid.py
+++ b/kombu/tests/transport/test_qpid.py
@@ -16,7 +16,6 @@ from mock import call
 from kombu.five import Empty, keys, range, monotonic
 from kombu.transport.qpid import (AuthenticationFailure, Channel, Connection,
                                   ConnectionError, Message, NotFound, QoS,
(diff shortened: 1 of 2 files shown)


bmbouter:
closing and reopening so that travis will rerun the correct commit after a force push

bmbouter:
Travis is still using the wrong commit. Closing and opening an entirely new PR

PR #452, opened 2015-02-08
Use multi-host support built-in to kazoo

harlowja:
The kazoo client already has the ability to use multiple
zookeeper servers as a comma separated list so instead of
sending only a single server in; send all the possible servers
and alternatives in to the client to let it natively handle
the servers to connect to.


ask:
Thanks!

PR #453, opened 2015-02-12
fix json serializable error for Queue.as_dict()

tumb1er:
Fix for https://groups.google.com/forum/#!searchin/celery-users/binding/celery-users/qeUiHmR7xmk/fmWA9rvIsCAJ

When celery config has a Queue with not-empty bindings,  worker can't reply to ` celery inspect active_queues`.
It just raises an error
```
[2015-02-12 12:22:03,554: ERROR/MainProcess] Control command error: EncodeError(TypeError('<binding: exchange->queue_name> is not JSON serializable',),)
```
It's because `Queue.as_dict()` method returns a dict where bindings is a set of `kombu.binding objects`.

Fix is to recursively descent to `set` members and make `binding` implement `as_dict()`

### DIFF
--- a/kombu/abstract.py
+++ b/kombu/abstract.py
@@ -41,6 +41,8 @@ class Object(object):
         def f(obj, type):
             if recurse and isinstance(obj, Object):
                 return obj.as_dict(recurse=True)
+            if recurse and hasattr(obj, '__iter__'):
+                obj = [f(o, None) for o in obj]
             return type(obj) if type else obj
         return {
             attr: f(getattr(self, attr), type) for attr, type in self.attrs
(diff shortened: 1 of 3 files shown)


tumb1er:
Close besause of recursion error it Travis.

tumb1er:
Added another fix.

ask:

tumb1er:
@ask Seems that wasn't merged properly.
```Closed with unmerged commits
This pull request is closed, but the tumb1er:json_error branch has unmerged commits.```
celery:master now doesn't have changes from tumb1er:json_error

compare https://github.com/celery/kombu/blob/master/kombu/entity.py#L310
and https://github.com/tumb1er/kombu/blob/json_error/kombu/entity.py#L302

* binding must extend Object instead of object
* binding must have as_dict method

ask:
Weird, not sure what happened there.  Thanks! Merged into master and 3.0

PR #454, opened 2015-02-17
Add a more useful message __repr__

harlowja:
In cases where a message is written to a log it is quite
useful to have the message format itself in a somewhat useful
format versus the default python ``repr`` which isn't quite useful.

This commit adds such useful ``__repr__`` and makes it so that it
returns a string that has some useful information about the
internals of the message (which can be handy when it is needed).


ask:
Thanks! Merged into master

Issue #455, opened 2015-02-19
qpid transport does not close sockets when close() is called


beav:
The following code leaves a socket open:

```python
from kombu import Connection

conn = Connection("qpid://localhost:5672/")
conn.connect()
conn.close()
raw_input("Press enter to terminate python process")

```
output of `netstat -anp | grep <python pid>` while process is running:
```
tcp        0      0 127.0.0.1:46353         127.0.0.1:5672          ESTABLISHED 32373/python
```

Eventually, opening and closing the connection will result in the system running out of file handles.

bmbouter:
PR available at  #456

bmbouter:
This has been resolved.

bmbouter:
There was an additional regression that needed fixing with this issue. It has been fixed with PR #485

PR #456, opened 2015-02-20
Fixes close bug where the connection was not being closed

bmbouter:
The connection was never actually closed before, which wasn't good. This fixes it. I verified the fix against the reproducer script in celery/kombu#455


beav:
@bmbouter this worked for me in my simple test case. I will also test in my more complex case as well.

beav:
this works for me, thanks!

PR #458, opened 2015-02-26
Fix custom transport overriding

daevaorn:
Custom transport class that has `can_parse_url=False` is overrided in `params['transport']` with one that returns from `parse_url` call.


ask:

PR #461, opened 2015-03-17
allow custom regioninfo and port, is_secure options for SQS

peonone:
Add support for local SQS server, such as [fake_sqs](https://github.com/iain/fake_sqs).


ask:
Thanks! Merged into master

PR #464, opened 2015-03-29
Add django postgresql transport that replicates 'acks_late' 

aclowes:
By using by using postgresql's advisory locks, we can mark a row as 'locked' until either the worker's connection to the database is lost (and the locks are released) or the row is deleted because the task finished processing.

Inspired by [que](https://github.com/chanks/que) by @chanks

@ask I'd like to get your feedback on the approach, and I'd like your suggestions on what to improve before considering this pull request. Should I use the kombu.common.QoS if implementing a transport that acks correctly? Or follow the Redis `ack_emulation` somehow?

### DIFF
--- a/funtests/tests/test_django_postgres.py
+++ b/funtests/tests/test_django_postgres.py
@@ -5,8 +5,8 @@ from kombu.tests.case import redirect_stdouts
 from funtests import transport
 
 
-class test_django(transport.TransportCase):
-    transport = 'django'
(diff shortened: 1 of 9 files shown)


ask:
This looks good to me!

It looks like you also fixed locking so that two workers cannot reserve the same task?

aclowes:
Yes, it should prevent two workers from reserving the same task.

I'm closing this and have opened a new one rebased off the latest master: #536

PR #477, opened 2015-05-08
Fix #476, file descriptor leak in transport.qpid.Transport.

### LINKED ISSUE #476: The qpid transport - Transport class leaks (2) file descriptors
The *Transport* class provided by the qpid transport creates a pipe in __init__().  The pipe file descriptors are never closed.  As a result, each connect or re-connect causes these two file descriptors to be leaked.

kombu 3.0.20 on f21.

jortel:
https://github.com/celery/kombu/issues/476

errno = EBADF is raised when calling os.close() on an already closed file descriptor.

Thought hard about logging cases when the os.close() raised OSError with errno != EBADF but decided an invalid or already closed file descriptor is really the only reason the close() would fail.  So, in the interest of simplicity, I decided to just ignore it.

Also, base.Transport does not implement ```__del__()``` so did not do: super(Transport, self).```__del__()```.

### DIFF
--- a/kombu/tests/transport/test_qpid.py
+++ b/kombu/tests/transport/test_qpid.py
@@ -2023,3 +2023,26 @@ class TestTransport(ExtraAssertionsMixin, Case):
         my_transport = Transport(self.mock_client)
         result_params = my_transport.default_connection_params
         self.assertDictEqual(correct_params, result_params)
+
+    @patch('os.close')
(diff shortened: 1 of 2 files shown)


bmbouter:
I reproduced the file descriptor issue on the 3.0 branch, and then verified that it is no longer an issue with this PR. Merging!

PR #483, opened 2015-06-10
pools have limits, not connections

gdw2:
Not 100% sure this is correct...


gdw2:
Yeah, I think this is wrong.  Closing...

ask:
No, you're right, the connection does not have a limit!

PR #486, opened 2015-06-12
fixing htmlclean

gdw2:
Not sure what this was supposed to do since `$(SPHINX)` wasn't defined anywhere.  Changed to call clean within docs dir.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ docsclean:
 	-rm -rf "$(SPHINX_BUILDDIR)"
 
 htmlclean:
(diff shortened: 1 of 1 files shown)


ask:
Thanks! Won't `make html` actually build the docs?

gdw2:
yes, will fix.

MrTomWhite:
@gdw2 This looks like it's two commits to the same line (roughly). Do you think we could squash them down into one logical commit? It'll be a little bit nicer on our history that way.

ask:
Thanks! (squashed) and merged into master

PR #489, opened 2015-06-19
do_close_self after setting connection to None will not close previou…

mainred:
…s connection

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -444,7 +444,6 @@ class Connection(object):
                     if max_retries is not None and retries > max_retries:
                         raise
                     self._debug('ensure connection error: %r', exc, exc_info=1)
-                    self._connection = None
                     self._do_close_self()
                     errback and errback(exc, 0)
                     remaining_retries = None


mainred:
If the connection still alive, it may 1. be a waste of resource, 2. receive messages from the MQ server without being noticed by the client, which is very dangerous side effect.

thedrow:
I'm not sure about this one.
@ask @malinoff What do you guys think?

malinoff:
+1, but I don't know about why this was originally written in this way - I'd ask @ask about that.

MrTomWhite:
>"If the connection still alive, it may 1. be a waste of resource, 2. receive messages from the MQ server without being noticed by the client, which has a very dangerous side effect."

@mainred: Could we add some automated tests to validate these two invariants? When we have that testing and the build + tests are passing it's probably safe to merge and close the PR. :sunglasses:

mainred:
@MrTomWhite , thank you for your recommendation, I'll try to add such tests later.

ask:
AMQP close operation is actually synchronous, so client sends close, and then waits for close_ok, which means all sorts of things can happen, or not happen, between those two steps.

Using collect here is probably safe for now, we may look into properly handling the synchronous close, but it's rather tricky.

mainred:
@ask , may I get the kombu version that will fix this bug?   Thank you very much.

ask:
It's in the master branch (to be kombu 4.0)

PR #491, opened 2015-07-01
fix docs about content_type and serialization

tbarbugli:
if serializer is None and content_type is not specified, then serialization will happen (see messaging.Producer:_prepare)


thedrow:
Thanks!

PR #500, opened 2015-07-27
Fixes SASL mech selection to be from broker string

### LINKED ISSUE #498: Qpid broker uses broker PLAIN sasl mech unexpectedly
With some versions of saslwrapper, when using a broker string that does not contain a username@ or username:password@ PLAIN is unexpectedly used. The sasl mech should be driven by the broker string.

bmbouter:
Fixes SASL mech selection to be from broker string

Also fixes some small test runner errors in qpid
and redis transport related to a newer version
of mock running on Travis.

Fixes #498

### DIFF
--- a/kombu/transport/qpid.py
+++ b/kombu/transport/qpid.py
@@ -42,6 +42,7 @@ import threading
 import time
 
 from itertools import count
+from gettext import gettext as _
 
 import amqp.protocol
 
@@ -1167,7 +1168,6 @@ class Connection(object):
 
     # A class reference to the :class:`Channel` object
     Channel = Channel
-    SASL_ANONYMOUS_MECH = 'ANONYMOUS'
 
     def __init__(self, **connection_options):
         """Instantiate a Connection object.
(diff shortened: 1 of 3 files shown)


kombu/transport/qpid.py
```diff
@@ -1205,66 +1210,42 @@ def __init__(self, **connection_options):
...
+            logger.debug(msg)
+            self._qpid_conn = establish(**self.connection_options)
+            # connection was successful if we got this far
+            msg = _('Connected to qpid with SASL '
+                    'mechanism %s') % sasl_mech
+            logger.info(msg)
+        except ConnectionError as conn_exc:
+            # if we get one of these errors, do not raise an exception.
+            # Raising will cause the connection to be retried. Instead,
+            # just continue on to the next mech.
```
beav: It looks like we only do one mech now, instead of retrying another mech

beav:
one thing to check into, we may be able to rename the parameter from sasl_mechanisms to singular sasl_mechanism if we are just passing one mech in.

kombu/transport/qpid.py
bmbouter: That is correct we only do one now based on how the broker string is configured. Letting the client libraries auto-negotiate has been hazardous because they all behave a little differently.

bmbouter:
I looked into renaming sasl_mechanisms from plural to singular and ultimately this comes from the [qpid.messaging setting `sasl_mechanisms`](https://qpid.apache.org/releases/qpid-0.26/messaging-api/python/api/index.html). We should leave it matching the real setting value we pass into `qpid.messaging`.

dkliban:
LGTM

PR #510, opened 2015-09-11
Fix kombu.transport.django migration for Django 1.7+

craigds:
South doesn't support django 1.7+, so `kombu.transport.django` breaks things for users using supported django versions.

This PR uses the recommended approach for libraries [1], which is to mandate South 1.0+ for django<1.7 users, and stick south migrations in a different directory, and provide core migrations for django 1.7+ users.

[1] See https://south.readthedocs.org/en/latest/releasenotes/1.0.html


ask:
I'm sorry, seems I  worked on this in parallel and didn't see your pull request :(

This is already supported in kombu 3.0.27

PR #511, opened 2015-09-12
Remove surplus % from parameters of log.debug

jindongh:
There are 5 % with only 4 parameters, remove the surplus one.


ask:
Thanks! Merged into master

PR #512, opened 2015-09-13
Print access_key when SQS auth failed.

jindongh:
boto will try many scenarios to get access_key and secret key.
1. get from arguments.
2. get from .boto configuration file.
3. get from environment variables.
4. get from metadata service in elastic beanstalk environment.
So when we get access denied error, it is hard to identify whether we are using the wrong account or the access policy is wrong, this fix will show the access_key for access denied error, it will help developer diagnose problem.


thedrow:
Can you add tests that uses moto to mock AWS and verify that this is actually working as expected?

hankjin:
Hi, I added some test cases, does that looks ok to you?

ask:
Thanks! This was merged into master :)

PR #517, opened 2015-09-28
SentinelChannel with Redis Sentinel support

lmancini:
This PR adds a `Channel` class with support for Redis Sentinel. Inspired by this gist: https://github.com/celery/celery/issues/1800.  Not sure this is in line with current Kombu roadmap, so it's basically a discussion start.

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -959,3 +959,50 @@ class Transport(virtual.Transport):
     def _get_errors(self):
         """Utility to import redis-py's exceptions at runtime."""
         return get_redis_error_classes()
+
+
+class SentinelChannel(Channel):
+    """Channel with explicit Redis Sentinel knowledge.
+
+    Broker url is supposed to look like:
+
+    sentinel://0.0.0.0:26379;0.0.0.0:26380/...
+
+    where each sentinel is separated by a `;`. Multiple sentinels are handled
+    by :class:`kombu.Connection` constructor, and placed in the alternative
+    list of servers to connect to in case of connection failure.
+
+    Other arguments for the sentinel should come from the transport options
+    (see :method:`Celery.connection` which is in charge of creating the
+    `Connection` object).
+    """
+
+    from_transport_options = Channel.from_transport_options + (
+        'service_name',
+        'password',
+        'min_other_sentinels',
+        'sentinel_timeout')
+
+    @cached_property
+    def _sentinel_managed_pool(self):
+
+        connparams = self._connparams()
+
+        sentinel = redis.sentinel.Sentinel(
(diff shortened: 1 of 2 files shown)


lmancini:
Rebased against current master; @ask if you think the approach is fine, I can start writing tests for the new `Channel` subclass.

kombu/transport/redis.py
```diff
@@ -959,3 +959,50 @@ def on_readable(self, fileno):
...
         """Utility to import redis-py's exceptions at runtime."""
         return get_redis_error_classes()
+
+
+class SentinelChannel(Channel):
+    """Channel with explicit Redis Sentinel knowledge.
+
+    Broker url is supposed to look like:
+
+    sentinel://0.0.0.0:26379;0.0.0.0:26380/...
```
thedrow: I don't really like the semicolon separated urls.
This doesn't play nicely with how failover is already built in Celery.
Maybe this should be implemented as a failover strategy instead of a channel?

ask: @thedrow: Just beneath it says 'Multiple sentinels are handled by :class:`kombu.Connection` constructor, and placed in the alternative list of servers to connect to in case of connection failure`

So seems it's already using that?

AFAIR you need the scheme for all URLs in the list:

   `sentinel://0.0.0.0:26379;sentinel://0.0.0.0:26380/`

as kombu can actually failover to a different type of transport ;)

It doesn't have to use semicolon separated urls, it can pass a list instead:

   x = Connection(['sentinel://0.0.0.0:26379/', 'sentinel://0.0.0.0:26380/'])

ask:
Looks good to me, except for maybe the missing schemas in the urls.

@thedrow?

Would be nice if you also updated the `docs/getting-started/brokers/redis.html` document in Celery

ask:
Please also add unit tests!

thedrow:
I'm OK with this.

ask:
Please add unit tests ASAP so we can get this into kombu 4.0

gsmethells:
+1

ZuSe:
+1

xavierhardy:
any news?

thedrow:
If anyone is willing to rebase this and add tests we'll merge this.

xavierhardy:
Isn't that done? Shouldn't we close that then?
https://github.com/celery/kombu/blob/c627c6d9d1f29ae0bcb03931d5f24001e60f72a9/Changelog

Great feature, we'll test that on Monday, thanks!

thedrow:
@xavierhardy You are correct. It is done. Thanks for the heads up.

@lmancini Thank you very much for your contribution.

Issue #521, opened 2015-10-02
libSystem.dylib fails to load on OS X 10.11


ewang:
On OS X 10.11, the libSystem = ctypes.CDLL('libSystem.dylib') call in kombu/five.py on the v3.0 branch fails to load with image not found error.

I am using kombo v3.0.26 that's depended on by celery v3.1.18

However, if the above call was changed to ctypes.CDLL(find_library('libSystem.dylib')), the call succeeds.

Disabling SIP resolves the issue, so it's probably due to SIP not allowing access to the /usr/lib folder. However, passing in an absolute path /usr/lib/libSystem.dylib to ctypes.CDLL() or using find_library() works.

ask:
Closed by #522

PR #522, opened 2015-10-02
Fix libSystem.dylib import on OS X 10.11

ewang:
PR for #521 

This will also need to be ported in py-amqp eventually since the master simply inherits from amqp.five

### DIFF
--- a/kombu/five.py
+++ b/kombu/five.py
@@ -49,7 +49,7 @@ if sys.version_info < (3, 3):
 
     if SYSTEM == 'Darwin' and has_ctypes:
         from ctypes.util import find_library
-        libSystem = ctypes.CDLL('libSystem.dylib')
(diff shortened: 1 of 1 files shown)


ask:
Thanks!

Weirdly, this is not an issue for me on my 10.11 installation, but maybe I have a slightly different environment or something, deployment target or whatever.

ask:
Merged into kombu 3.0, pyamqp master+1.4, billiard master,3.3

ewang:
Do you think you would be able to do a  release on pypi that includes just this change?

I already saw an article on stackoverflow w/ this same issue and I imagine more will start encountering it.


Thanks!

sideboltconor:
Also wondering if there is an eta on this update. A number of our developers ran into this issue when upgrading to 10.11.

Much appreciated!

ask:
Kombu 3.0.27 was released on  2015-10-09 :)

PR #528, opened 2015-10-13
Update required redis version to match code

spjwebster:
...as a follow up to f07ca7d4a7e1fc62d6073b5791dbbdcabde39709


ask:
We don't actually need this any more, as I downgraded the version requirement after just testing for the existence of redis.TimeoutError instead.

PR #529, opened 2015-10-16
ability not to log dispatch errors

mihapro59:
I suggest to control when to log exception of handling dispatched method and when not to log. 
Anyway, negative answer containing exception details will return to the sender, so he will handle it properly.
Usage example:
from celery.app import app_or_default
from celery.app.control import Control
control = Control(app_or_default())
reply = control.pool_shrink(1, [worker_name], reply=True, no_logging=True)

### DIFF
--- a/kombu/pidbox.py
+++ b/kombu/pidbox.py
@@ -106,7 +106,9 @@ class Node(object):
         except SystemExit:
             raise
         except Exception as exc:
-            error('pidbox command error: %r', exc, exc_info=1)
+            no_logging = arguments.get('no_logging', False)
+            if not no_logging:
+                error('pidbox command error: %r', exc, exc_info=1)
             reply = {'error': repr(exc)}
 
         if reply_to:


ask:
I'm not so sure about having this as an argument to the message. Would it not be better to have this as configuration in the server?

mihapro59:
We may need to log dispatch errors for some messages and not to log for the others. Configuration in the server do not give us control of this.

ask:
It just seems like a weird design, the client does not even see the effect of passing in the argument, and disables a possibly security sensitive feature of the worker (someone sending weird requests to the worker).

Why do you need it? Is there some specific error you want to silence?

mihapro59:
I agree with you about danger of disabling security feature of the worker.
But I need it.
This is my situation: I try to shrink pool of some worker. I checked that worker has free processes to shrink, so in theory I will not see "Can't shrink pool. All processes busy!" in logs, but sometimes it happens. I guess this is because of time shifts - while I'm checking free processes, they already have a new job! I do not want to litter my log file in this case, I just try to shrink again.
Maybe I use wrong approach to shrink pool. If so, could you please give me advice.

thedrow:
I have to agree with @ask here.
Sorry it took me so long to review this PR.

I'm going to close this. If you still feel strongly about this enhancement and have another solution to propose, please do.

PR #531, opened 2015-10-21
Removes all threading usage from Qpid Transport code

bmbouter:
This was made possible thanks to a new callback that can
be registered with the Qpid messaging client 0.32+. The
necessary versions of those dependency libraries will be
available on PyPI soon.

closes celery/celery#2199
https://github.com/celery/celery/issues/2199


ask:
Sweet!!! Great work Brian!

bmbouter:
The dependencies are available on PyPI now! merging!

PR #536, opened 2015-11-07
Add django postgresql transport that supports 'acks_late' by using postgresql's advisory locks

aclowes:
- remove outdated funtest requirements
- add a note about using CELERY_ACKS_LATE


ask:
Using SQL as a broker is no longer supported as we don't have the resources to maintain it.

I'm sorry it took me so long to reply to this issue, but this is also the reason why we have to slim down supported transports :(

PR #537, opened 2015-11-08
MongoDB TTL support and refactorings

daevaorn:
I have this change set for about year and half and now it is ready for review:-)

Fixes https://github.com/celery/celery/issues/1047

I haven't find proper place in kombu documentation to describe available transport options and new capabilities:-(


ask:
Definitely missing that documentation, but some if it is described in the celery documentation:
http://docs.celeryproject.org/en/latest/getting-started/brokers/

Not sure how we can have a canonical location for this information, while still having user friendly celery documentation.

Btw, did you see I added a way to programmatically introspect features for the different transports?

e.g:
https://github.com/celery/kombu/blob/master/kombu/transport/SQS.py#L471-L474
https://github.com/celery/kombu/blob/master/kombu/transport/SQS.py#L471-L474
https://github.com/celery/kombu/blob/master/kombu/transport/SQS.py#L471-L474

ask:
Thanks!  Merged into master.

flake8 was complaining a lot about long lines, and I removed the use of deprecated `self.assertEquals` among some other minor stuff.

jonnycrunch:
This TTL approach seems reasonable, however, which branch of celery, amp does it correlate with?

lqez:
:+1:

PR #538, opened 2015-11-13
Set default priority for all messages in a queue

gtmanfred:
This allows for better prioritizing, if you always one one queue to be
empty first, then you don't have to set the priority on every task, just
make sure that that task gets routed to the high priority queue which puts
the correct priority on the message.

Thanks
Daniel


gtmanfred:
I think i tested that this actually worked wrong.

PR #540, opened 2015-11-17
Fix uuid shim in upcoming Python 3.5.1.

groner:
After a recent change in cpython, uuid._uuid_generate_random() no longer
exists.

https://hg.python.org/cpython/rev/70be1f9c9255
http://bugs.python.org/issue25515

The issue being worked around was resolved in 2007, so it might be
sensible to just remove the work around.

Tested with 3.5.0 and 3.5 tip at this point:
https://hg.python.org/cpython/shortlog/f928dcb448a9

### DIFF
--- a/kombu/utils/__init__.py
+++ b/kombu/utils/__init__.py
@@ -16,7 +16,11 @@ from contextlib import contextmanager
 from itertools import count, repeat
 from functools import wraps
 from time import sleep
-from uuid import UUID, uuid4 as _uuid4, _uuid_generate_random
+from uuid import UUID, uuid4
+try:
+    from uuid import _uuid_generate_random
+except ImportError:
+    _uuid_generate_random = None
 
 from kombu.five import items, reraise, string_t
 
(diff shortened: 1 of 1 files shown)


hannosch:
The same issue also happens with Python 2.7.11rc1.

malinoff:
This code exists due to bug which was fixed in python 2.5. @ask can we simply throw this code away, since `kombu` now supports python2.7+ only?

ask:
Yeah, that sounds reasonable :)

ask:
Thanks! Merged into kombu 3.0 and for master we now just use uuid.uuid4()

remik:
Hi,
Is it possible to release new 3.0.x version?
Python 2.7.11 has been released so more people would be happy to see fixed version:)

ask:
Kombu 3.0.30 is on PyPI!  Thanks for the fix

bhagany:
would it be possible to cut a release of of the 2.5.x branch with this fix?

PR #571, opened 2016-02-16
Updates Qpid transport with updated names of Qpid async interface

bmbouter:


bmbouter:
All Qpid tests pass, so these failures are not related to this PR.

Issue #506, opened 2015-08-20
Virtual transport does not support multiple bindings for queues


hochbergg:
The virtual transport does not allow binding a queue with different routing keys. When `bind_to()` is called for the same queue more than once, the `queue_bind()` function in the virtual `Channel` class discards the call silently if the queue is already in `state.bindings`.

I have verified this in the `memory` virtual broker. 

I believe the AMQP protocol does allow for multiple bindings with the same queue, and I know its supported in RabbitMQ. Also, its very useful to have the `memory` virtual broker support this to allow testing scenarios.

nazavode:
Trying to add this feature on a [fork](https://github.com/nazavode/kombu) of the latest stable release. All the tests seems to be fine up to now, still a bit of work to be carried out but it looks promising.

nazavode:
#574 is closed, maybe we need to close this one as well @ask?

ask:
Thanks for reminding me! :)

PR #574, opened 2016-03-15
Multiple queue bindings for virtual transports

nazavode:
This PR adds the ability to define multiple bindings from a queue to the same exchange when using virtual transports (for native transports it is already supported by AMQP itself) as requested in #506 by @hochbergg.
Since defining a bind is almost _free_ in terms of resources, in situations where you want to subscribe (and unsubscribe) to different topics dynamically, adding and removing bindings is _much_ more efficient than adding and removing dedicated exchanges (I can provide benchmarks for the latest RabbitMQ if it makes sense). A simple usage example could be:
```python
queue.bind_to(exchange, routing_key=key) for key in ['key_a', 'key_b', 'key_c']
```
You can find a fairly complete test in [this gist](https://gist.github.com/nazavode/97926f916798729c8e4e).

In the current `master` code base, running the code above using a virtual transport will discard all bindings declarations that follow the first one.

Note: I'm working on a optimized rewrite of `BrokerState` class that reduces the complexity of queue deletion operations. Since it needs a new structure for exchanges' routing tables, I'm keeping it as a [branch](https://github.com/nazavode/kombu/tree/feature/multibind).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ kombu/tests/cover
 kombu/tests/coverage.xml
 .coverage
 dump.rdb
+.idea/
--- a/kombu/tests/transport/virtual/test_base.py
+++ b/kombu/tests/transport/virtual/test_base.py
@@ -28,11 +28,9 @@ class test_BrokerState(Case):
     def test_constructor(self):
         s = virtual.BrokerState()
         self.assertTrue(hasattr(s, 'exchanges'))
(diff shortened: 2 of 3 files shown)


ask:
Nice work! Merged into master.

As `exchanges` is public, I renamed the private variables to be public.   BrokerState is already internal-ish.

PR #581, opened 2016-04-04
Add support for arbitrary json messages to SQS transport

juancferrer:
I needed to be able to connect to an SQS queue and listen for json messages NOT created with kombu/celery
Random json messages created without the use the celery/kombu libraries do not include all the json keys that kombu was looking for when parsing the message, and it would fail.
Here's an update that should solve that issue.

I've added a new test that checks that the _message_to_python method can handle arbitrary json messages.


ask:
Thanks!  Merged into master

PR #584, opened 2016-04-11
Add Consul K/V store Transport

wido:
Using the Key/Value store of Consul.io this commit adds a new
Transport.

It is a simple Transport which Messages as Keys in the K/V store
of Consul using the HTTP API.

The HTTP API is consumed using the python-consul client which exposes
the API in a Python way.

### DIFF
--- a/kombu/transport/consul.py
+++ b/kombu/transport/consul.py
@@ -1,0 +1,272 @@
+"""
+Consul Kombu Transport
+
+It uses Consul.io's Key/Value store to transport messages in Queues
+
+It uses python-consul for talking to Consul's HTTP API
+"""
+from __future__ import absolute_import, unicode_literals
+
+import uuid
+import socket
+
+from . import virtual
+from kombu.exceptions import ChannelError
+from kombu.log import get_logger
+from kombu.five import Empty, monotonic
+from kombu.utils.json import loads, dumps
+
+try:
+    import consul
+except ImportError:
+    consul = None
+
+LOGGER = get_logger('kombu.transport.consul')
+
+DEFAULT_PORT = 8500
+DEFAULT_HOST = 'localhost'
+
+
+class Channel(virtual.Channel):
+    """
+    Consul Channel class which talks to the Consul Key/Value store
+    """
+    prefix = 'kombu'
+    index = None
+    timeout = '10s'
+    session_ttl = 30
+    lock_name = '%s' % socket.gethostname()
+    queues = {}
+
+    def __init__(self, *args, **kwargs):
+        if consul is None:
+            raise ImportError('Missing python-consul library')
+
+        super(Channel, self).__init__(*args, **kwargs)
+
+        port = self.connection.client.port or self.connection.default_port
+        host = self.connection.client.hostname or DEFAULT_HOST
+
+        LOGGER.debug('Host: %s Port: %d Timeout: %s', host, port, self.timeout)
+
+        self.client = consul.Consul(host=host, port=int(port))
+
+    def _lock_key(self, queue):
+        return '%s/%s.lock' % (self.prefix, queue)
+
+    def _key_prefix(self, queue):
+        return '%s/%s' % (self.prefix, queue)
+
+    def _get_or_create_session(self, queue):
+        """
+        Try to renew the session if it exists, otherwise create a new session
+        in Consul
+
+        This session is used to obtain a lock inside Consul so that we achieve
+        read-consistency between the nodes
+
+        :param queue: The name of the Queue
+        :return: The ID of the session
+        """
+
+        session_id = None
+        try:
+            session_id = self.queues[queue]['session_id']
+        except KeyError:
+            pass
+
+        if session_id is not None:
+            LOGGER.debug('Trying to renew existing session %s',
+                         self.queues[queue]['session_id'])
+            session = self.client.session.renew(session_id=session_id)
+            session_id = None
+            try:
+                session_id = session['ID']
+            except KeyError:
+                pass
+
+        if session_id is None:
+            LOGGER.debug('Creating session %s with TTL %d', self.lock_name,
+                         self.session_ttl)
+            session_id = self.client.session.create(name=self.lock_name,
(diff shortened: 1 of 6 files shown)


wido:
Comments are welcome! This is my first PR for Kombu and I hope it's OK :)

ask:
Thanks, this is looking good!  But we need some tests to make sure it does not regress, would you have time to write some test cases?

wido:
@ask Yes, I wanted to do that, but I liked to get a first review first.

I see the tests are now failing on python-consul not being present on the Travis server.

Mocking and UnitTesting aren't my strongpoints in Python though :(

ask:
Please get the tests in, or this will not make the Kombu 4 release.  It can perfectly well be in 4.1 though, as it doesn't affect any other components.

wido:
@ask: The Unit Tests have been updated. Is this one still in time for 4.0?

kombu/transport/consul.py
```diff
@@ -0,0 +1,272 @@
...
+from kombu.log import get_logger
+from kombu.five import Empty, monotonic
+from kombu.utils.json import loads, dumps
+
+try:
+    import consul
+except ImportError:
+    consul = None
+
+LOGGER = get_logger('kombu.transport.consul')
```
ask: changed to use lowercase logger to be consistent

kombu/transport/consul.py
```diff
@@ -0,0 +1,272 @@
...
+
+class Channel(virtual.Channel):
+    """
+    Consul Channel class which talks to the Consul Key/Value store
+    """
+    prefix = 'kombu'
+    index = None
+    timeout = '10s'
+    session_ttl = 30
+    lock_name = '%s' % socket.gethostname()
```
ask: moved to cached_property to avoid calling socket.gethostname as a side effect of importing the module

kombu/transport/consul.py
```diff
@@ -0,0 +1,272 @@
...
+        in Consul
+
+        This session is used to obtain a lock inside Consul so that we achieve
+        read-consistency between the nodes
+
+        :param queue: The name of the Queue
+        :return: The ID of the session
+        """
+
+        session_id = None
```
ask: long function split into three functions

kombu/transport/consul.py
```diff
@@ -0,0 +1,272 @@
...
+        """
+        session_id = self._get_or_create_session(queue)
+        lock_key = self._lock_key(queue)
+
+        LOGGER.debug('Trying to create lock object %s with session %s',
+                     lock_key, session_id)
+
+        if self.client.kv.put(key=lock_key,
+                              acquire=session_id,
+                              value=self.lock_name) is False:
```
ask: Do you have a special reason for using `is False`?  I changed it to use `not self.client.kv.put(...)`, as unless there's a comment telling me otherwise, I'm going to expect that this is a mistake :)

kombu/transport/consul.py
```diff
@@ -0,0 +1,272 @@
...
+                raise Empty
+
+            LOGGER.debug('Removing key %s with modifyindex %s',
+                         data[0]['Key'], data[0]['ModifyIndex'])
+
+            self.client.kv.delete(key=data[0]['Key'],
+                                  cas=data[0]['ModifyIndex'])
+
+            return loads(data[0]['Value'])
+        except TypeError:
```
ask: Where is the TypeError here expected to come from?  The error is so generic you can easily end up silencing some important error.

kombu/transport/consul.py
```diff
@@ -0,0 +1,272 @@
...
+        size = 0
+        try:
+            key = '%s/msg/' % self._key_prefix(queue)
+            LOGGER.debug('Fetching key recursively %s with index %s', key,
+                         self.index)
+            self.index, data = self.client.kv.get(key=key, recurse=True,
+                                                  index=self.index,
+                                                  wait=self.timeout)
+            size = len(data)
+        except TypeError:
```
ask: Same with TypeError here, is that expected to be from the tuple unpacking?

kombu/transport/consul.py
```diff
@@ -0,0 +1,272 @@
...
+        port = connection.client.port or self.default_port
+        host = connection.client.hostname or DEFAULT_HOST
+
+        LOGGER.debug('Verify Consul connection to %s:%d', host, port)
+
+        try:
+            client = consul.Consul(host=host, port=int(port))
+            client.agent.self()
+            return True
+        except ValueError:
```
ask: Is the ValueError something raised by int(port)?

kombu/transport/consul.py
```diff
@@ -0,0 +1,272 @@
...
+            )
+        )
+
+        self.channel_errors = (
+            virtual.Transport.channel_errors + (
+                consul.ConsulException, consul.base.ConsulException
+            )
+        )
+
+    def verify_connection(self, connection):
```
ask: Does this establish a new connection?

Most transports will not be able to check that a connection is still connected, so it's perfectly
reasonable to just return True here.

ask:
Perfect! There's still time :)

I had to modify the tests quite a bit, since they weren't passing at all.
Also made some changes to the transport, but mostly cosmetic.

I merged it into master, but there are some unresolved issues
in the comments, hope you have time to clarify them for me!

wido:
@ask, thanks! I'll send a new PR to address those changes.

kombu/transport/consul.py
wido: Correct, that's where it comes from

kombu/transport/consul.py
wido: Yes, it does. I thought this was to check if the supplied data is sufficient to create a connection

kombu/transport/consul.py
wido: Yes, that's where it comes from. The consul client might not return a Tuple

kombu/transport/consul.py
wido: No, no special reason. I think the 'not' is just fine to use.

PR #588, opened 2016-05-06
Changed the timeouts on reserve calls  so that blockage on beanstalkd is not excessive

ChaksChigurupati:


ChaksChigurupati:
@dziegler : Can you please take a look at these diffs and provide your comments? Thx.

ChaksChigurupati:
@ask : Can you please take a look at this PR as well? Do you think this beanstalk interface needs to be adapted for Billiards? The Travis failure seems to be unrelated to the changes in the PR:

      test_do_restore_message (kombu.tests.transport.test_redis.test_Channel) ... FAIL

ask:
I'm not sure about this change, will this not make it spin excessively fast?  How much CPU is the worker using with this enabled?

The beanstalk transport is still using the non-async prefork pool, so it's in a threaded environment.

ask:
Using Beanstalk as a broker is no longer supported, as we don't have the resources to maintain this transport :(

PR #598, opened 2016-06-03
Fix funtests for pyamqp and django

alanjds:
The funtests are broken, at least for Python 3.4.

Tested on Python 2.7 and 3.4 on my machine

### DIFF
--- a/funtests/transport.py
+++ b/funtests/transport.py
@@ -21,6 +21,9 @@ if sys.version_info >= (2, 5):
 else:
     from sha import new as _digest  # noqa
 
+if not hasattr(string, 'letters'):
+    string.letters = string.ascii_letters
+
 
 def say(msg):
     print(msg, file=sys.stderr)
--- a/funtests/transport.py
+++ b/funtests/transport.py
@@ -168,6 +168,8 @@ class TransportCase(unittest.TestCase):
                 purged += self.purge_consumer(consumer)
 
     def _digest(self, data):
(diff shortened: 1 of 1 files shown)


funtests/transport.py
```diff
@@ -21,13 +21,19 @@
 else:
     from sha import new as _digest  # noqa
 
+if not hasattr(string, 'letters'):
```
ask: What version of Python does not have string.letters?  Kombu 3.0 supports 2.6 and up, and as far as I can tell it's present.

alanjds: At least Python 3.4.3 have it not, but both Python 2.7.11 and 3.5 have it!

```bash
$ python -V
Python 3.4.3
$ python -c "import string; string.letters"
  File "<string>", line 1, in <module>
AttributeError: 'module' object has no attribute 'letters'
```

auvipy:
@alanjds could you fix the build failures?

funtests/transport.py
ask: Ah, so it was removed in Python 3!  Seems the reasoning behind the change is that `string.letters` actually varies defined on the locale set (which is news to me), so string.ascii_letters is probably what we should use always.

ask:
I merged it manually, since I was so slow at responding to this, but only against master.  I'm not sure there is any rush merging with 3.0 and we really need to get this stuff out the door.

PR #599, opened 2016-06-06
Fixed queue lookup in virtual transports with multiple overlapping bindings

nazavode:
When a queue is bound to an exchange with multiple bindings whose patterns overlap, the AMQP standard says that a single message shall be delivered *once* to that queue. RabbitMQ behaves according to the standard. When using virtual transports in `kombu/master`, a message will be delivered *one time for each binding that matches the topic pattern*, even for overlapping ones.
For example, the correct (AMQP) behaviour is:
```
# Queue bindings: ['mytopic', 'mytopic.#', '#']
-> Sent 1 message with topic: 'mytopic'
-> Delivered 1 message to queue
```
`kombu/master` behaves like follows:
```
# Queue bindings: ['mytopic', 'mytopic.#', '#']
-> Sent 1 message with topic: 'mytopic'
-> Delivered same message 3 times to the same queue
```

A regression test has been added to the commit.

### DIFF
--- a/kombu/transport/virtual/exchange.py
+++ b/kombu/transport/virtual/exchange.py
@@ -11,7 +11,6 @@ from __future__ import absolute_import, unicode_literals
 from kombu.utils import escape_regex
 
 import re
-import collections
 
 
 class ExchangeType(object):
--- a/kombu/transport/virtual/exchange.py
+++ b/kombu/transport/virtual/exchange.py
@@ -11,6 +11,7 @@ from __future__ import absolute_import, unicode_literals
 from kombu.utils import escape_regex
 
 import re
+import collections
 
 
 class ExchangeType(object):
@@ -51,8 +52,14 @@ class DirectExchange(ExchangeType):
     type = 'direct'
 
     def lookup(self, table, exchange, routing_key, default):
-        return [queue for rkey, _, queue in table
-                if rkey == routing_key]
+        # Using an OrderedDict to purge queue duplicates while
+        # keeping the same order in which they appear in the table
+        return list(
+            collections.OrderedDict.fromkeys(
+                queue for rkey, _, queue in table
+                if rkey == routing_key
+            )
+        )
 
     def deliver(self, message, exchange, routing_key, **kwargs):
         _lookup = self.channel._lookup
@@ -75,8 +82,14 @@ class TopicExchange(ExchangeType):
     _compiled = {}
 
     def lookup(self, table, exchange, routing_key, default):
-        return [queue for rkey, pattern, queue in table
-                if self._match(pattern, routing_key)]
+        # Using an OrderedDict to purge queue duplicates while
+        # keeping the same order in which they appear in the table
+        return list(
+            collections.OrderedDict.fromkeys(
+                queue for rkey, pattern, queue in table
+                if self._match(pattern, routing_key)
+            )
+        )
 
     def deliver(self, message, exchange, routing_key, **kwargs):
         _lookup = self.channel._lookup
@@ -120,7 +133,13 @@ class FanoutExchange(ExchangeType):
     type = 'fanout'
 
     def lookup(self, table, exchange, routing_key, default):
-        return [queue for _, _, queue in table]
+        # Using an OrderedDict to purge queue duplicates while
+        # keeping the same order in which they appear in the table
+        return list(
+            collections.OrderedDict.fromkeys(
+                queue for _, _, queue in table
+            )
+        )
 
     def deliver(self, message, exchange, routing_key, **kwargs):
         if self.channel.supports_fanout:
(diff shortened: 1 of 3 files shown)


kombu/transport/virtual/exchange.py
```diff
@@ -51,8 +52,14 @@ class DirectExchange(ExchangeType):
     type = 'direct'
 
     def lookup(self, table, exchange, routing_key, default):
-        return [queue for rkey, _, queue in table
-                if rkey == routing_key]
+        # Using an OrderedDict to purge queue duplicates while
```
nazavode: I don't know whether keeping the delivery order is meaningful or not, but discarding it leads some other tests to failure. A simpler solution that avoids using an `collections.OrderedDict` (but *randomizes* the result of queues in the method result) would be:
```python
return list(
    set(
        queue for rkey, _, queue in table
        if rkey == routing_key
    )
)
```

kombu/transport/virtual/exchange.py
```diff
@@ -11,6 +11,7 @@
 from kombu.utils import escape_regex
 
 import re
+import collections
```
nazavode: Used a `collections.OrderedDict` because it is already in use throughout the project (for example in `kombu/connection.py`) even if it excludes python 2.6 from the compatible interpreters.

kombu/transport/virtual/exchange.py
```diff
@@ -75,8 +82,14 @@ class TopicExchange(ExchangeType):
     _compiled = {}
 
     def lookup(self, table, exchange, routing_key, default):
-        return [queue for rkey, pattern, queue in table
-                if self._match(pattern, routing_key)]
+        # Using an OrderedDict to purge queue duplicates while
+        # keeping the same order in which they appear in the table
+        return list(
```
ask: This is nice! I didn't know about .fromkeys.  You can probably remove `list(` here to avoid making a copy, as using a dict as a list will already use the keys.

nazavode:
I left the ``list`` copy because of the ``assertListEqual`` in tests that were failing since they expect a list instance. Trying to use a directly a ``set`` as a return value from ``lookup`` but some Redis tests are failig, lokking into this...

nazavode:
Ok, all tests are passing now. The biggest change here is that the queues returned by ``lookup`` methods of every virtual exchange type are unordered (used to be a ``queue``, now a ``set``). This reflects the AMQP behaviour since there are no guarantees on the actual dispatch order of a message. @ask do you think that makes sense?

ask:
Yeah, I think that makes sense.

Merged into master!  Really nice work!

nazavode:
Sorry @ask but I think I don't see these changes in master...

ask:
That's strange, the commits are in master and merging the branch again says it's already merged.

ask:
Let me go over this manually to make sure everything is in

ask:
Hey! When I fetch nazavode/master I don't get these commits at all... 

These are the most recent commits in the branch that I receive:

commit a6aa1a5e5019421a52b3719b25f5a2a58a864881
Date:   Tue Mar 15 11:31:55 2016 +0100
Removed useless bindings parameter from BrokerState constructor

commit 9dd0dc27d897b58fe5c3511ceec531c0f17bb539
Date:   Tue Mar 15 11:15:25 2016 +0100
Finished queue_unbind for virtual channel, tests passing

commit 4328bcd677c47d507aacad7092c12a86dd88ea05
Date:   Mon Mar 14 15:20:28 2016 +0100
Passing unit tests

commit 3416432667682827a7e18d8910d97b9031feb62b
Date:   Mon Mar 14 11:52:31 2016 +0100
Initial support for multiple queue bindings

commit 5172b1312aa9e6aafbb2429bb9fb25af36c82485
Date:   Mon Mar 14 09:58:23 2016 +0100
Added PyCharm stuff to .gitignore


I tried deleting my local copy of your branch, and refetching it again using:

```
   git branch -d nazavode/master
   gh fetch nazavode/master
```

but still not seeing the commits listed here :/

ask:
I found them now!  They aren't in the log for that branch, but I could still cherry-pick them by SHA.  Very strange :/

It's in master now!

nazavode:
Thank you so much @ask ! This behaviour looks very strange, I think it could be related to the fact that after a previous PR you reintegrated in master I deleted my fork, and then forked again to work on this PR. Looks like a GitHub bug...

ask:
Thanks!  That sounds reasonable.  The best practice is to use dedicated branches for every feature. We don't enforce that as I think the contribution is more important than following rules.    I usually cherry-pick patches to make sure I'm not taking in any unwanted diffs, so scary to think I could have brought in unreviewed patches by just merging blindly.

nazavode:
I think I've missed the [Celery contribution guide](http://docs.celeryproject.org/en/master/contributing.html) :)

Issue #301, opened 2014-01-09
Kafka support


rdiomar:
Any plans to add Kafka to the mix? mahendra implemented it [here](https://github.com/mahendra/kombu/tree/kafka). It would be amazing if that gets pulled in.

xtfxme:
@mahendra any chance you can expand? even if briefly?

we are investigating the viability of using kafka as a message broker, and then making celery behave as sort of stream processor, to unify our uses of kafka and celery.

since your kafka transport has not received any love since 2013, and since you seem to have some reservations, i would very much like to know what they are before committing more time!

rafaelhbarros:
@mahendra any updates?

ninowalker:
@rafaelhbarros @mahendra @xtfxme - got clarification on the async I/O question: 

"Just means we need to register the socket(s) in the event loop, i.e. at least consuming cannot be blocking" - Ask, https://twitter.com/asksol/status/609424548396707840

rafaelhbarros:
Is there any interest in the community to add support for Kafka?

ask:
Yeah, I think there's quite some interest.  I would really like that to happen!

mahendra:
@rafaelhbarros i was originally planning to use kafka with celery (batch) for collecting metrics coming from clients. Not sure if kafka is a good fit for non-batch operations with celery. To get this working with kombu/celery there are few patches that I have to merge into python-kafka. Give me a week or two, let me check on the status of the python-kafka patches. I will get back.

Apologies for going incommunicado on this :-(

rafaelhbarros:
@mahendra @ask I appreciate the responses, I'll keep watching this issue with hopes!

thedrow:
@mahendra That's awesome!

cerdman:
+1

davidmashburn:
+1

shazeline:
+1

lavr:
+1

cerdman:
anyone have thoughts on where i could start (in terms of building a broker interface/adapter for kafka) on at least developing a proof of concept for this? i am good with kafka / python ... but have not used celery a ton - i have an urgent need to use celery now to save myself from reinventing the wheel completely on a project.

thedrow:
@cerdman The first step is to pick a Python client for Kafka that can work for the versions that we support.
[...]

PR #600, opened 2016-06-08
Kafka support (PR based on #301)

alanjds:
Based on @erickt and @BenjaminDavison work detailed at #301, I am trying to get the Kafka transport in.

This PR is not really final and working right now, but should be updated until acceptable.

### DIFF
--- a/kombu/transport/kafka.py
+++ b/kombu/transport/kafka.py
@@ -1,0 +1,177 @@
+"""
+kombu.transport.kafka
+=====================
+
+Kafka transport.
+
+:copyright: (c) 2010 - 2013 by Mahendra M.
+:license: BSD, see LICENSE for more details.
+
+**Synopsis**
+
+Connects to kafka (0.8.x) as <server>:<port>/<vhost>
+The <vhost> becomes the group for all the clients. So we can use
+it like a vhost
+
+It is recommended that the queue be created in advance, by specifying the
+number of partitions. The partition configuration determines how many
+consumers can fetch data in parallel
+
+**Limitations**
+* The client API needs to modified to fetch data only from a single
+  partition. This can be used for effective load balancing also.
+
+"""
+
+from __future__ import absolute_import
+
+import socket
+
+from anyjson import loads, dumps
+
+from kombu.exceptions import StdConnectionError, StdChannelError
+from kombu.five import Empty
+
+from . import virtual
+
+try:
+    from pykafka import KafkaClient
+
+    KAFKA_CONNECTION_ERRORS = ()
+    KAFKA_CHANNEL_ERRORS = ()
+
+except ImportError:
+    kafka = None                                          # noqa
+    KAFKA_CONNECTION_ERRORS = KAFKA_CHANNEL_ERRORS = ()   # noqa
+
+DEFAULT_PORT = 9092
+
+__author__ = 'Mahendra M <mahendra.m@gmail.com>'
+
+
+class Channel(virtual.Channel):
+
+    _client = None
+    _kafka_group = None
+    _kafka_consumers = {}
+    _kafka_producers = {}
+
+    def _get_producer(self, queue):
+        """Create/get a producer instance for the given topic/queue"""
+
+        producer = self._kafka_producers.get(queue, None)
+        if producer is None:
+            producer = self.client.topics[queue].get_sync_producer()
+            self._kafka_producers[queue] = producer
+
+        return producer
+
+    def _get_consumer(self, queue):
+        """
+        Create/get a consumer instance for the given topic/queue
+        """
+        consumer = self._kafka_consumers.get(queue, None)
+        if consumer is None:
+            consumer = self.client.topics[queue].get_simple_consumer(consumer_group=self._kafka_group,
+                                                                     auto_commit_enable=True,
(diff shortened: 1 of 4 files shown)


ask:
Thanks for working on this!  I don't know much about Kafka, but I'll be reading this for sure.

kombu/transport/kafka.py
```diff
@@ -0,0 +1,336 @@
+"""
+kombu.transport.kafka
```
dpkp: Can this be renamed kombu.transport.pykafka to avoid naming conflicts w/ other libraries (i.e., kafka-python) ?

ask: We always add `from __future__ import absolute_import` to every file, the build will break if not, so it shouldn't be a problem

ask: also prefer the nice name, and refuse to give in to the insane implicit relative imports behavior.  

There was much rejoicing the day Python 2.5 was dropped as a supported version, and we gave everything proper names :o)

dpkp: I'm not refering to implicit relative imports. I'm referring to an alternate kombu.transport.* that provides a kafka driver using, for example, kafka-python. I dont see a way in this implementation to choose whether you want to use pykafka or kafka-python. how would I do that?

alanjds: I totally agree with naming it ``pykafka``. Would be nice to have a ``kafka_python`` driver in the future.

alanjds: Well, for the 1st time we have green tests! Now is time to refactor the thing if needed. Lets start with the futures `absolute_import` and `unicode_literals`?

ask: I don't know that much about the kafka ecosystem, but do we really want to have more than one kafka transport at the same time?  If kafka_python is better, then could it not replace the pykafka based one later?  I think it'd be easier for users if they don't have multiple choices.

I did the same executive decision for the cassandra result backend in Celery 4.0rc, where the new version was first suggested to be named 'new_cassandra'.

If there's a compelling reason for a small subset of the users to continue using the older version, it's possible to publish a compatibility extension on PyPI.

dpkp: I'm obviously biased (I maintain kafka-python), but yes the kafka world has 2 functional python clients. The client apis are definitely different, the implementations are different, and the driver designs are different.

Even if you decide to ship w/ pykafka here, I still would like to use kafka-python w/ my teams that use celery at work. And I would prefer to not ship a second-class plugin even if it is not included in the official distribution.

Dowwie:
Make that 3 options to choose from.  The members of this discussion missed confluent-kafka-client, which seems to now lead the group in terms of  performance.

see this blog post:  http://activisiongamescience.github.io/2016/06/15/Kafka-Client-Benchmarking/

kombu/transport/kafka.py
Dowwie: Make that 3 options to choose from. The members of this discussion missed confluent-kafka-client, which seems to now lead the group in terms of performance.

see this blog post: http://activisiongamescience.github.io/2016/06/15/Kafka-Client-Benchmarking/

kombu/transport/pykafka.py
```diff
@@ -0,0 +1,336 @@
+"""
+kombu.transport.pykafka
```
alanjds: Bad bad Github! After updating with the rename change, the past comments vanished. But are still being count.

Alright.... rename done.

Dowwie: Oh, so you didn't just drop the hammer on my comment for an unknown reason.  What I commented on was that from out of left field came a high performance *third* [python client for kafka](https://github.com/confluentinc/confluent-kafka-python) and a blog post that benchmarked the three solutions:  http://activisiongamescience.github.io/2016/06/15/Kafka-Client-Benchmarking/

alanjds: @Dowwie no, I don't. I guess this is because of the file rename :/ . I still have the comments cached on other browser tab. Should I paste the whole thing here?

alanjds:
Btw, funtests are passing on my machine.

auvipy:
2.6 and 2.7 are failing. shouldn't the support of 2.6 is dropped? @ask

alanjds:
@auvipy In my machine it passes. Is there a way to tell Travis to show the traceback?

ask:
The diff must be applied to master, not '3.0' as it is now!

If the 3.0 branch is target, then Python 2.6 will also be.

ask:
I think it's too late to get this into 4.0, but as it's a new transport it could easily be part of 4.1

AlJohri:
hey @ask @alanjds, what's left to get kafka support into celery? looks like its slated for the next release? https://github.com/celery/kombu/milestone/2

alanjds:
Well, I heard somewhere else that it works, and I remember the tests to be passing. So is just merging I guess.

However, a new major release was issued in the mean time...

auvipy:
sorry to see this didn't get response for long... all the core maintainers were busy fighting with bug fixing for critical issues. any one willing to volunteer this effort should start it rebasing on master.

georgepsarakis:
@auvipy should we start a new PR?

auvipy:
I believe yes but not for 4.2 release

PR #604, opened 2016-06-23
kombu 2.5: apply hotfix for Issue #540

megahall:
As requested by @bhagany in this comment: https://github.com/celery/kombu/pull/540#issuecomment-172126169 , because we needed the same bug fix.

### DIFF
--- a/kombu/tests/test_utils.py
+++ b/kombu/tests/test_utils.py
@@ -17,7 +17,7 @@ from kombu.utils.compat import next
 
 from .utils import (
     TestCase,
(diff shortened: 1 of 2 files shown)


megahall:
Pinging this PR for consideration.

ask:
I'm more than a little bit scared of releasing a new 2.5 version, but I guess the patch should be safe.
Just need to find the time to do this

megahall:
@ask That's a fair concern. I just want to keep it running right for people who still need it. It is very hard to achieve the right effect via monkey-patches or similar things, because the error blows up deep inside the launch code for Celery and Django Celery tasks and so forth.

PR #608, opened 2016-07-07
Connection deepcopy

cazino:
A Connection instance cannot be 'deepcopied' if it has alternates urls.


ask:
Thanks!  Merged into master and 3.0

PR #618, opened 2016-08-09
Removes some Qpid tests causing other test skips

bmbouter:
The Qpid tests were mocking pipe and fcntl and this
was causing a strange behavior with nose. When I
would run `nosetests` not all unit tests would run.

By removing these three tests we don't loose much
in test coverage and lots of other tests now run.


bmbouter:
The test failure is due to a missing test dependency. I verified that all of the Qpid tests ran so I'm merging this.

PR #629, opened 2016-09-26
Mailbox constructor got queue and reply_queue TTL and Expires options

alanjds:
New `Mailbox.__init__()` options: `queue_ttl`, `queue_expires`, `reply_queue_ttl` and `reply_queue_expires`

`utils.maybe_s_to_ms` and `utils.functional.dictfilter` were copied-over
from Celery implementation (from actual master)


ask:
Looks good to me!

alanjds:
@ask: Thank you very much. Will use this branch as the base for a PR to celery yet to be coded.

alanjds:
Just a thing: I had noticed that the tests not passing are the ones for Python 3. Is desirable to the next patches to be complaint with Python 3, probable via http://www.python-future.org/ ?

auvipy:
you need to get this patch pass against python3 too to be accepted.

alanjds:
Ok. Will do with python-future. Ping if this is not acceptable. Thanks

auvipy:
python3 support is very important at the moment :) It shouldn't be that difficult for adding python3 support

alanjds:
Done.

ask:
This patch should actually be applied to master, not 3.0.

We use `from kombu five import items`,  iteritems is soooo Python 2, doesn't seem like you need anything else from python-future here.

Next major version is Python 3 only so not sure changing compatibility interface is of priority.

kombu/pidbox.py
```diff
@@ -227,7 +243,18 @@ def get_queue(self, hostname):
         return Queue('%s.%s.pidbox' % (hostname, self.namespace),
                      exchange=self.exchange,
                      durable=False,
-                     auto_delete=True)
+                     auto_delete=True,
+                     queue_arguments=self._get_queue_arguments())
+
+    def _get_queue_arguments(self, ttl=None, expires=None):
```
ask: I'd suggest writing this in a different way, as there's a bit of duplication in these methods.
Currently it would be difficult for a subclass to override how the reply_queue_ttl is set,
and it's difficult to add new arguments.

I'd rather see something like a dispatch table of arguments and their types:

```py

class Node(object):
    queue_arguments = {
        'x-message-ttl': maybe_s_to_ms,
        'x-expires': maybe_s_to_ms,
    }

   def __init__(self, ...):
          self.queue_arguments = self._prepare_queue_arguments()
          self.reply_queue_arguments = self._prepare_reply_queue_arguments()

    def _prepare_queue_arguments(self):
        return self._to_queue_arguments([
              ('x-message-ttl', self.queue_ttl),
              ('x-expires, self.queue_expires),
        ])

    def _prepare_reply_queue_arguments(self):
        return self._to_queue_arguments([
              ('x-message-ttl', self.reply_queue_ttl),
              ('x-expires, self.reply_queue_expires),
        ])

     def _to_queue_arguments(self, args):
         return dictfilter({
             key: self.queue_arguments[key](value)
             for key, value in args
        })

```

I guess now we can argue if this rely belongs in the Queue class.

kombu/utils/__init__.py
```diff
@@ -451,3 +452,7 @@ def maybe_fileno(f):
         return fileno(f)
     except FILENO_ERRORS:
         pass
+
+
+def maybe_s_to_ms(v):
```
ask: Note that in master the utils package has been refactored into many subpackages instead, and I fear you need to
add a .utils.time at this point.

kombu/utils/functional.py
```diff
@@ -77,6 +78,12 @@ def maybe_list(l, scalars=(Mapping, string_t)):
     return l if l is None or is_list(l, scalars) else [l]
 
 
+def dictfilter(d=None, **kw):
+    """Remove all keys from dict ``d`` whose value is :const:`None`"""
+    d = kw if d is None else (dict(d, **kw) if kw else d)
+    return dict((k, v) for k, v in iteritems(d) if v is not None)
```
ask: In master you should use a dict comprehension here (we're Python 2.7 only):

```py
{k: v for k, v in items(d) if v is not None}
```

ask (changes requested):
- Needs to be applied against master, not 3.0. Sorry about that  :(
- Instead of future should use `from kombu.five import items`

kombu/utils/functional.py
alanjds: No problem. I prefer this idiom too, but just copied over. Will update the PR

alanjds: ...Wait. On 3.X I should support older versions too?

kombu/utils/__init__.py
alanjds: Done.

alanjds: Done.

kombu/pidbox.py
alanjds: As this this is very related to the queue, my 1st sight is that it belongs to the Queue class. 

What about a staticmethod or a utils package? The proposed change seems less readable and too coupled for me :/

auvipy (changes requested):
is it possible that we can retarget it to main branch?

Issue #628, opened 2016-09-26
Redis Transport: attemping to create an SSL connection raises an exception


bkolba:
Repro environment:
- `kombu==4.0.0rc4`
- `redis-server` running locally

Two bugs here in the Redis transport that break the SSL connection

1. `connection_class` should be `redis.SSLConnection`, not `redis.Connection` (which does not support any ssl parameters). There is a workaround for this by adding `connection_class` to the SSL dict (e.g. via `celery.conf.broker_use_ssl`) since that entire dict ends up being added to `connparams`. That said, the better fix is to explicitly set the `connection_class` here similar to the socket case.

2. Redis-py requires a very specific set of SSL parameters to be passed in to initialize a connection. The supported parameters are:

    - ssl_cert_reqs
    - ssl_ca_certs
    - ssl_certfile
    - ssl_keyfile

    The `ssl` flag that is appended to `connparams` on L883 of `kombu/transport/redis.py` causes `redis.Connection` and `redis.SSLConnection` to fail initialization (stack trace below) due to an unknown keyword. While the `StrictRedis` client understand this keyword, it is not necessary to pipe it through since `KombuRedis` ends up explicitly setting the `connection` property.

```python
>>> from kombu.connection import Connection
>>> with Connection('redis://', ssl=True) as conn:
...     t = conn.connect()

...
Traceback (most recent call last):
  File "kombu/transport/virtual/base.py", line 912, in create_channel
    return self._avail_channels.pop()
IndexError: pop from empty list

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "kombu/utils/objects.py", line 40, in __get__
    return obj.__dict__[self.__name__]
KeyError: 'client'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "redis/connection.py", line 895, in get_connection
    connection = self._available_connections.pop()
IndexError: pop from empty list

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "kombu/connection.py", line 258, in connect
    return self.connection
  File "kombu/connection.py", line 798, in connection
    self._connection = self._establish_connection()
  File "kombu/connection.py", line 751, in _establish_connection
    conn = self.transport.establish_connection()
  File "kombu/transport/virtual/base.py", line 932, in establish_connection
    self._avail_channels.append(self.create_channel(self))
  File "kombu/transport/virtual/base.py", line 914, in create_channel
[...]

PR #634, opened 2016-09-29
Fix Redis SSL support

bkolba:
Fixes celery/kombu#628

**Root cause:**
- setting `connparams[‘ssl’] = True` will cause an unexpected keyword
exception when initializing a redis `Connection` or `SSLConnection`
- connection_class must be `redis.SSLConnection`. The base `redis.Connection`
class does not support SSL parameters.

**Notes:**
- Updated the comment and tests to reflect the required format since it
is different than amqplib. Ideally the Celery documentation that talks about `BROKER_USE_SSL` should be updated to specify the keyname format for redis vs amqplib.
- If you attempt to use Redis-over-SSL as both a transport and results backend then Celery will fail since the `RedisBackend.connparams` are not patched to support SSL. These end up passed directly into `redis.ConnectionPool` which defaults to the base `redis.Connection` class if one isn't specified. I didn't dig deeply enough into the Celery plumbing to know the right way to patch that but this workaround enabled Redis-over-SSL end to end for both the broker and results backend:

```python
# force=True isn't required, but in my case I grab cert files from s3 if they're missing
# when setting `BROKER_USE_SSL` and need to ensure it runs now before querying for
# that property below.
from redis.connection import SSLConnection
app = Celery()
app.config_from_object('celeryconfig', force=True)

if app.conf.broker_use_ssl:
    app.backend.connparams.update(app.conf.broker_use_ssl)
    app.backend.connparams['connection_class'] = SSLConnection
```

@ask: Do you want me to open an issue over on the Celery side for the Results backend (at least to document the workaround)? I wasn't sure the relationship between the two projects and best practices for managing issues.

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -879,10 +879,11 @@ class Channel(virtual.Channel):
             'socket_keepalive_options': self.socket_keepalive_options,
         }
         if conninfo.ssl:
-            # Connection(ssl={}) can be a dict like in amqplib.
-            connparams['ssl'] = True
+            # Connection(ssl={}) must be a dict containing the keys:
+            # 'ssl_cert_reqs', 'ssl_ca_certs', 'ssl_certfile', 'ssl_keyfile'
             try:
                 connparams.update(conninfo.ssl)
+                connparams['connection_class'] = redis.SSLConnection
             except TypeError:
                 pass
         host = connparams['host']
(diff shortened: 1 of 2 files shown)


bkolba:
Test failures appear unrelated to this change. I show the same test failing in this build: https://travis-ci.org/celery/kombu/builds/163570757

Appears to have started with https://github.com/celery/kombu/commit/3dea21b3a254c330e549cc2086545d62543ca012 -- quick spot check of builds prior to that do not show these failures.

auvipy:
plz add a changlog/release note too

bkolba:
I believe this is already covered in the Changelog -- this patch simply makes the statement true :)

>- Redis: now supports SSL using the ``ssl`` argument to
  :class:`~kombu.Connection`.

Issue #657, opened 2016-11-16
'Transport' object has no attribute 'use_async_interface'


goosemania:
It looks like [`verify_runtime_environmentenvironment` is called](https://github.com/celery/kombu/blob/master/kombu/transport/qpid.py#L1429-L1431) before the attribute `use_async_interface` is defined.

```
Traceback (most recent call last):
  File "/home/vagrant/.virtualenvs/pulp/bin/celery", line 11, in <module>
    load_entry_point('celery', 'console_scripts', 'celery')()
  File "/usr/lib/python3.5/site-packages/celery-4.0.0-py3.5.egg/celery/__main__.py", line 14, in main
    _main()
  File "/usr/lib/python3.5/site-packages/celery-4.0.0-py3.5.egg/celery/bin/celery.py", line 326, in main
    cmd.execute_from_commandline(argv)
  File "/usr/lib/python3.5/site-packages/celery-4.0.0-py3.5.egg/celery/bin/celery.py", line 488, in execute_from_commandline
    super(CeleryCommand, self).execute_from_commandline(argv)))
  File "/usr/lib/python3.5/site-packages/celery-4.0.0-py3.5.egg/celery/bin/base.py", line 278, in execute_from_commandline
    return self.handle_argv(self.prog_name, argv[1:])
  File "/usr/lib/python3.5/site-packages/celery-4.0.0-py3.5.egg/celery/bin/celery.py", line 480, in handle_argv
    return self.execute(command, argv)
  File "/usr/lib/python3.5/site-packages/celery-4.0.0-py3.5.egg/celery/bin/celery.py", line 412, in execute
    ).run_from_argv(self.prog_name, argv[1:], command=argv[0])
  File "/usr/lib/python3.5/site-packages/celery-4.0.0-py3.5.egg/celery/bin/base.py", line 282, in run_from_argv
    sys.argv if argv is None else argv, command)
  File "/usr/lib/python3.5/site-packages/celery-4.0.0-py3.5.egg/celery/bin/base.py", line 365, in handle_argv
    return self(*args, **options)
  File "/usr/lib/python3.5/site-packages/celery-4.0.0-py3.5.egg/celery/bin/base.py", line 241, in __call__
    ret = self.run(*args, **kwargs)
  File "/usr/lib/python3.5/site-packages/celery-4.0.0-py3.5.egg/celery/bin/control.py", line 75, in run
    return self.do_call_method(args, **kwargs)
[...]

PR #658, opened 2016-11-16
Switch to using getattr in qpid.Transport.__del__

pcreech:
The \_\_init\_\_ function calls verify_runtime_environment before setting
the use_async_interface variable.  If it throws an exception in that
function, the \_\_del\_\_ method will be called, and raises an
AttributeError

using getattr in the __del__ method to guard against this

### DIFF
--- a/kombu/transport/qpid.py
+++ b/kombu/transport/qpid.py
@@ -1731,7 +1731,7 @@ class Transport(base.Transport):
 
     def __del__(self):
         """Ensure file descriptors opened in __init__() are closed."""
-        if self.use_async_interface:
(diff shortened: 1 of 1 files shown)


pcreech:
This should resolve #657

auvipy:
travis is failing and there isn't any tests for this

pcreech:
The travis failures don't appear to be related to this PR.  Based on what I see under the PR page, the test 'test_Channel.test_list_bindings' started failing sometime before I submitted this.

As far as tests for this edge case, if needed I can write one.  I initially didn't see the need as I was just changing the way an attribute was accessed, but that thought could be wrong.

PR #689, opened 2017-01-27
Virtual transport timeout does not supesede period polling

### LINKED ISSUE #688: Virtual transport timeout does not supesede period polling
If period polling is large and timeout is small, periodic polling should not exceed timeout.

arcivanov:
fixes #688

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -140,3 +140,4 @@ Vincent Driessen <vincent@datafox.nl>
 Wido den Hollander <wido@widodh.nl>
 Zach Smith <zmsmith27@gmail.com>
 Zhao Xiaohong <mrluanma@gmail.com>
+Arcadiy Ivanov <arcadiy@ivanov.biz>
--- a/kombu/transport/virtual/base.py
+++ b/kombu/transport/virtual/base.py
@@ -956,6 +956,8 @@ class Transport(base.Transport):
         time_start = monotonic()
         get = self.cycle.get
         polling_interval = self.polling_interval
+        if timeout and polling_interval and polling_interval > timeout:
+            polling_interval = timeout
         while 1:
             try:
                 get(self._deliver, timeout=timeout)
--- a/t/unit/transport/virtual/test_base.py
+++ b/t/unit/transport/virtual/test_base.py
@@ -4,6 +4,7 @@ import io
 import pytest
 import sys
 import warnings
+import socket
 
 from case import MagicMock, Mock, patch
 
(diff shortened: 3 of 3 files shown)


arcivanov:
@auvipy ping?

arcivanov:
@auvipy ping?

auvipy:
sorry. looks OK. but could you also include some tests to verify the change?

@ask @malinoff

arcivanov:
@auvipy will do

arcivanov:
@auvipy done

auvipy:
quite surprised this gone out off my radar

AlexHill:
Hi @arcivanov just looking at this issue for the release notes, and it's not clear to me what was actually done here.

Initially I thought that point of this change was to ensure that `socket.timeout` was NOT raised when the polling period exceeded the timeout - and yet, the test in this PR asserts that the exception IS raised.

Could you give a bit more detail on the issue you encountered and the fix?

arcivanov:
@AlexHill 
The purpose of a timeout is to make sure that drain_events doesn't get stuck forever. The issue that this fixes is that if timeout is small and period is huge, the blocking time in drain_events will exceed the timeout arbitrarily.

https://github.com/celery/kombu/pull/689/files#diff-7658f59b04660e1104a1a78ac2756788R562
https://github.com/celery/kombu/pull/689/files#diff-7658f59b04660e1104a1a78ac2756788R565

Here a `polling_interval` is 60s, but timeout is 0.5s. If the first `get` raises `Empty` instantaneously the thread will sleep for 60s, exceeding requested timeout by a factor of 120. This, obviously, defeats the purpose of `timeout` argument to the `drain_events` method.

The fix ensures that if a `timeout` is specified and `polling_period` is specified, the and the `polling_period` is greater than the `timeout` then the `polling_period` will be set to the `timeout` specified.

AlexHill:
Ok, I understand that, but I still don't quite see how this fix helps. In the situation you mention, we set the polling interval to 0.5s.

Then, we
  1. call `get()` - this takes some non-zero time
  2. catch the `Empty` exception - this takes some non-zero time
  3. sleep for 0.5s
  4. loop

By the time we get back around to checking if `monotonic() - time_start >= timeout`, won't `monotonic() - time_start` ALWAYS be greater than or equal to `timeout`, whether our polling interval is 60s or 0.5s, because the time elapsed is 0.5 + some non-zero amount of time?

arcivanov:
> By the time we get back around to checking if monotonic() - time_start >= timeout, won't monotonic() - time_start ALWAYS be greater than or equal to timeout

Yes, but timeouts are never exact actually. The point is to timeout reasonably close to deadline.

PR #676, opened 2016-12-27
Don't modify dict size while iterating over it

felixonmars:
This fixes #675.

PR #692, opened 2017-02-13
Avoid modifying the dictionary being iterated over.

moriyoshi:
This ends up with an exception on Python 3.6.  (Some tests fail.)


moriyoshi:
Seems the dupe of #676... Close if it bothers.

PR #710, opened 2017-03-26
Parse password from Redis transport URL

### LINKED ISSUE #677: Use passwords passed in "redis+socket://" URLs.
This is the same issue as #460 but happening in the 4.x series now. Steps to reproduce and workarounds are the same as in the earlier bug report, and I suspect the code fix is similar, if not identical.

### Versions

Kombu version: 4.0.2
OS: Debian testing up to date as of 2016-12-28.

georgepsarakis:
- [x] Add password in Redis transport connection parameters.
- [x] Add unit test that verifies the change.

Closes #677.

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -894,7 +894,7 @@ class Channel(virtual.Channel):
                 pass
         host = connparams['host']
         if '://' in host:
-            scheme, _, _, _, _, path, query = _parse_url(host)
+            scheme, _, _, _, password, path, query = _parse_url(host)
             if scheme == 'socket':
                 connparams = self._filter_tcp_connparams(**connparams)
                 connparams.update({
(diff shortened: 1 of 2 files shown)


auvipy:
could you also check the landscape errors?

georgepsarakis:
@auvipy I did. There are several `cyclic import` warnings, unrelated to this PR. I wanted to try and fix them on a separate branch, but it will take some time.

auvipy:
sure

georgepsarakis:
@auvipy I think we cannot fix easily the circular import warnings. I ran pylint to find them and one example is: 
- https://github.com/celery/kombu/blob/master/kombu/connection.py#L751
- https://github.com/celery/kombu/blob/master/kombu/simple.py#L10

As you can see the import is in the `SimpleBuffer` function so this is intentional.

PR #713, opened 2017-04-11
Fix byte string for kazoo.Queue in Python 3

koorgoo:
Using Python 3 delaying a task with ZooKeeper as a broker fails with `TypeError: value must be a byte string` because [kazoo.Queue.put()](https://github.com/python-zk/kazoo/blob/8f6bf35d/kazoo/recipe/queue.py#L111) requires `value` argument to be a byte string .

I just add forced message conversion to bytes.

Will be thankful if anybody helps me with writing a test for it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ dump.rdb
 htmlcov/
 test.db
 coverage.xml
+venv/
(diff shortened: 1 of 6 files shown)


koorgoo:
@auvipy could you please say whether next version v4.0.3 will be released?

koorgoo:
Or should I ask @ask to do it?

auvipy:
no I am also a core team member :) hangout on irc

PR #693, opened 2017-02-15
Switching to boto3 only

revmischa:
I have removed all traces of boto2 that I could find. I got async message signing and parsing more or less working. Probably still needs some cleanup and testing. 
Much farther along now on the journey to boto3.

PR #714, opened 2017-04-14
Bugfix: Support Python 3.4, decode from base64 when converting messag…

alukach:
Bugfix: Support Python 3.4, decode from base64 when converting message to Python.

This is cleanup from issues I found with #693.

### DIFF
--- a/kombu/async/aws/sqs/message.py
+++ b/kombu/async/aws/sqs/message.py
@@ -19,7 +19,7 @@ class AsyncMessage(BaseAsyncMessage):
 
     def encode(self, value):
         """Encode/decode the value using Base64 encoding."""
-        return base64.b64encode(value).decode('utf-8')
+        return base64.b64encode(value.encode()).decode()
 
     def __getitem__(self, item):
         """Support Boto3-style access on a message."""
--- a/kombu/async/aws/sqs/message.py
+++ b/kombu/async/aws/sqs/message.py
@@ -2,8 +2,10 @@
 """Amazon SQS message implementation."""
 from __future__ import absolute_import, unicode_literals
 
-from kombu.message import Message
 import base64
+
+from kombu.message import Message
+from kombu.utils.encoding import str_to_bytes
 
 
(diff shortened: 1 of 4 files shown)


georgepsarakis:
Thanks for opening this PR, I will try to look at it ASAP.

kombu/async/aws/sqs/message.py
```diff
@@ -19,7 +19,7 @@ class AsyncMessage(BaseAsyncMessage):
 
     def encode(self, value):
         """Encode/decode the value using Base64 encoding."""
-        return base64.b64encode(value).decode('utf-8')
+        return base64.b64encode(value.encode()).decode()
```
georgepsarakis: A suggestion here:

```
# Python 2/3 compatibility
from kombu.five import bytes_if_py2
# base64 results to always ASCII character output
return base64.b64encode(bytes_if_py2(value)).decode('ascii')
```

alukach: Reading the code, I think that would actually be the opposite of what we want:

```
    def bytes_if_py2(s):
        """Convert str to bytes if running under Python 2."""
        return s
```

We want to input [`string` data if using Python 2](https://docs.python.org/2/library/base64.html#base64.b64encode) and to input [`bytes` data if using Python 3](https://docs.python.org/3.4/library/base64.html#base64.b64encode).

`.encode()` returns `string` in Python 2 and `bytes` in Python 3.  I'm pretty new to Python 3, so I may be missing something.  Maybe there's a better way to do this?

alukach: @georgepsarakis will merge when I have your OK

georgepsarakis: You are right. I think there is a suitable function that also handles Python 2/3 compatibility:

```
from kombu.utils.encoding import str_to_bytes
```

alukach: @georgepsarakis Good catch, fixed.

PR #715, opened 2017-04-16
Allow endpoint URL to be specified for SQS transport

georgepsarakis:
- [x] Add a `endpoint_url` property in SQS channel class.
- [x] Initialization SQS connection using endpoint URL, if defined.
- [x] Re-enable test cases by removing skip due to missing dependency (`boto`).

Related Pull Requests:
- #693 
- #714


georgepsarakis:
@auvipy thanks for reviewing. Tests are now passing, so I am merging this.

PR #725, opened 2017-04-19
Process todos first before polling

23doors:
As todos can be chained like amqp frame 1-2-3 and as a result may affect what we are polling for, we can end up with unfinished frame and poll that times out after 0-2s. Which results in a delayed task execution.

Fixes https://github.com/celery/celery/issues/3978 and a few related issues that make tasks run very slowly with amqp and celery 4.x. Didn't encounter any issues so far as todos at least for amqp are simply reads with 0 timeout.

### DIFF
--- a/kombu/async/hub.py
+++ b/kombu/async/hub.py
@@ -276,10 +276,11 @@ class Hub(object):
             for tick_callback in on_tick:
                 tick_callback()
 
-            while todo:
-                item = todo.pop()
-                if item:
+            if todo:
+                while todo:
+                    item = todo.pop()
                     item()
+                continue
 
             poll_timeout = fire_timers(propagate=propagate) if scheduled else 1
             #  print('[[[HUB]]]: %s' % (self.repr_active(),))
--- a/kombu/async/hub.py
+++ b/kombu/async/hub.py
@@ -279,7 +279,8 @@ class Hub(object):
             if todo:
                 while todo:
                     item = todo.pop()
-                    item()
+                    if item:
+                        item()
                 continue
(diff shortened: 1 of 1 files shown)


thedrow:
That code hasn't changed in years. I have no idea why it doesn't affect 3.x with amqp.

23doors:
Main difference that I found right now is in how pyamqp drain_events work in 2.x. In 1.x it seemed to have process reads until message was ready. In 2.x it just reads one frame and kombu queues next read. This queueing if not processed fast enough, will be delayed until next poll returns.

I guess there may be some other way to fix it. 
In pyamqp we could: 
- have on_frame return True for complete messages
- in drain_events - check if blocking_read returned False (it propagates on_frame). If it's false, repeat blocking_read until it's True.

This would be similar to how pyamqp worked in 1.x and it wouldn't touch kombu codebase in this case. @thedrow What do you think?

23doors:
Actually, I think fixing it in amqp would be better in this case.
PR: https://github.com/celery/py-amqp/pull/144
I'll wait for your feedback, but if that PR is merged, we won't need any changes in kombu.

Tested by running 1000 tasks in a group, sleeping for 2 seconds, running 1 task
Before:
```
1.480212s # time to run 1000 group tasks
1.297438s # time for just 1 task
```

After:
```
1.431802s
0.004595s
```

As expected as it mostly affects per task delay and group of tasks is unaffected. Now if we run 10 tasks sequentially we get:
Before:
```
11.87755s
```

After:
```
0.033392s
```

PR #727, opened 2017-04-20
Fix last chroot character trimming for ZooKeeper

koorgoo:
Noticed that kombu removes last chroot character.

For example, you connect to ZooKeeper with chroot `/project` but celery produces/consumes tasks with chroot `/projec`.

### DIFF
--- a/kombu/transport/zookeeper.py
+++ b/kombu/transport/zookeeper.py
@@ -29,6 +29,7 @@ import socket
 from kombu.five import Empty
 from kombu.utils.encoding import bytes_to_str, ensure_bytes
 from kombu.utils.json import dumps, loads
+
 from . import virtual
 
 try:
@@ -66,8 +67,8 @@ try:
         socket.error,
     )
 except ImportError:
-    kazoo = None                                    # noqa
-    KZ_CONNECTION_ERRORS = KZ_CHANNEL_ERRORS = ()   # noqa
+    kazoo = None  # noqa
+    KZ_CONNECTION_ERRORS = KZ_CHANNEL_ERRORS = ()  # noqa
 
 DEFAULT_PORT = 2181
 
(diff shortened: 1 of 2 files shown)


kombu/transport/zookeeper.py
```diff
@@ -28,8 +28,7 @@
 
 from kombu.five import Empty
 from kombu.utils.encoding import bytes_to_str, ensure_bytes
-from kombu.utils.json import loads, dumps
-
+from kombu.utils.json import dumps, loads
```
koorgoo: Sorry, PyCharm goes a little bit crazy.

kombu/transport/zookeeper.py
```diff
@@ -141,7 +140,7 @@ def _has_queue(self, queue):
 
     def _open(self):
         conninfo = self.connection.client
-        self.vhost = os.path.join('/', conninfo.virtual_host[0:-1])
```
georgepsarakis: So `conninfo.virtual_host` is a list/tuple here?

kombu/transport/zookeeper.py
```diff
@@ -166,6 +165,13 @@ def _open(self):
         conn.start()
         return conn
 
+    @staticmethod
+    def _normalize_chroot(chroot):
+        chroot = chroot.rstrip('/')
```
georgepsarakis: I am not sure you need this function. I think this can be reduced to:

```
chroot = '/{}'.format(chroot.strip('/'))
```

kombu/transport/zookeeper.py
georgepsarakis: That's fine :smile: .

kombu/transport/zookeeper.py
koorgoo: In ZooKeeper domain `virtual_host` (aka `chroot`) must be a string. It will be a prefix of a path to set a value for (Ctrl+F at [this doc page](https://zookeeper.apache.org/doc/r3.2.2/zookeeperProgrammers.html)).

I suppose `[0:-1]` slicing had the aim to prevent from chroots with trailing slashes, e.g. `/chroot/`.

kombu/transport/zookeeper.py
koorgoo: Great! I will inline it and come up with a better test for chroot "preparing".

georgepsarakis:
Thanks! Feel free to add yourself in the [AUTHORS](https://github.com/celery/kombu/blob/master/AUTHORS) file.

koorgoo:
Thank you, @georgepsarakis , [already did](https://github.com/celery/kombu/blob/bf820b20/AUTHORS#L40) in last PR 😃

PR #728, opened 2017-04-23
adding support for exclusive consumers

plockaby:
This adds support to amqp for creating an exclusive consumer. I didn't write any tests because I wanted to see if there was support for merging this. I have successfully run this code and it's pretty trivial. I'd like to see this in 4.x if possible?

### DIFF
--- a/kombu/entity.py
+++ b/kombu/entity.py
@@ -707,7 +707,7 @@ class Queue(MaybeChannelBound):
                                         nowait=nowait) or 0
 
     def consume(self, consumer_tag='', callback=None,
-                no_ack=None, nowait=False):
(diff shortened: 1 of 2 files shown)


plockaby:
It sure is failing. But given that everyone's pull requests in the last few days are failing for exactly the same reason and given that the failure is a pydocstyle error in files that I didn't touch, I do not believe that this was caused by my pull request. Are you asking me to fix them anyway?

auvipy:
nope not at all

plockaby:
It has been more than three years since my pull request was submitted. I'll see if I can find time to put something together.

auvipy:
I want to know what purpose it will serve or what problem will this solve?

Issue #720, opened 2017-04-17
SimpleConsumer get() race condition causing exceptions/possibly inaccurate timeouts


c-nichols:
When supplying a timeout perimeter to the get() method of SimpleBase, it is possible for the local variable `remaining` to be assigned a negative value, which is then used. This causes the Python socket library to throw an exception.

Also, currently the `remaining` variable does not count down time from when the method was first called. In theory, it should be close as long as there Isn't much contention for the CPU. I think it might make sense, to rewrite it so it will behave closer to "real"-time.

I would be happy to submit a pull request for this

EDIT for ease of googling, the exception message I encountered was:
`ValueError: Timeout value out of range`

georgepsarakis:
Sounds right. Sure, feel free to provide a fix.

benoit-intrw:
I have this same problem, is there a workaround waiting the release of this fix ?

c-nichols:
You should be able to subclass SimpleConsumer and override the get() method with the version in my commit referenced above. That is what I have done in the meantime

benoit-intrw:
Thank you for your answer and you patch!

PR #729, opened 2017-04-24
Address shortcomings with timeout parameter described in issue #720

c-nichols:


c-nichols:
Seems like the  errors related to my changes, maybe  preexisting comment style issues or a misconfigured job? Not sure what the right way to fix is.

PR #733, opened 2017-05-04
Prevent SQS from defaulting to localhost hostname

alukach:
Commit [`cdbfe9a64e1884dc99fc9806a834f23cf8398a3b`](https://github.com/celery/kombu/commit/cdbfe9a64e1884dc99fc9806a834f23cf8398a3b#diff-3cc1ff6df0f488077057a603b8f5e5d4R487) added tooling to retrieve the `hostname` from the SQS Transport.  If no hostname is provided (for example, if you only configure an app with `broker_transport = 'sqs'`), the hostname is [set to a default value](https://github.com/alukach/kombu/blob/0c20e84c63489b6efb2dd93f607e00fcec6f0597/kombu/connection.py#L585-L599) from the Transport baseclass' [`default_connection_params` property](https://github.com/alukach/kombu/blob/0c20e84c63489b6efb2dd93f607e00fcec6f0597/kombu/transport/virtual/base.py#L1002-L1003).  This is problematic, as the default is `'localhost'`, which is likely not correct for an SQS queue and results in failed message retrieval.  Interestingly, a Celery worker is still able to successfully start, retrieve the list of queues, and retrieve messages being that Celery [provides the Transport a default `url` value of `None`](https://github.com/celery/celery/blob/master/celery/app/base.py#L746-L760).

This PR omits `hostname` from the SQS Transport's `default_connection_params` property, thereby letting Boto3 use its default logic if no `hostname` is manually provided to the SQS Transport.

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -516,3 +516,7 @@ class Transport(virtual.Transport):
         async=True,
         exchange_type=frozenset(['direct']),
     )
+
+    @property
+    def default_connection_params(self):
+        return {'port': self.default_port}
--- a/t/unit/transport/test_SQS.py
+++ b/t/unit/transport/test_SQS.py
@@ -182,6 +182,10 @@ class test_Channel:
         boto3_sqs = SQS_Channel_sqs.__get__(self.channel, SQS.Channel)
         assert boto3_sqs._endpoint.host == expected_endpoint_url
 
+    def test_none_hostname_persists(self):
+        conn = Connection(hostname=None, transport=SQS.Transport)
+        assert conn.hostname == conn.clone().hostname
+
     def test_new_queue(self):
         queue_name = 'new_unittest_queue'
         self.channel._new_queue(queue_name)


alukach:
Pinging the author of the first commit, @georgepsarakis.

georgepsarakis:
@alukach thanks, perhaps I missed this case. Would you mind expanding the unit tests, probably [here](https://github.com/celery/kombu/blob/1b878f4a67612c73ea19ddd561477994bb4ba987/t/unit/transport/test_SQS.py#L175) ?

georgepsarakis:
Also, do you think this is actually breaking backwards compatibility?

alukach:
@georgepsarakis I think the test best exemplifies the true core of the issue. While it could technically break backwards functionality, I don't think this should be an issue as:

* I can't think of any reason why a user would want a Boto connection to `localhost`
* Cloned connections maintaining the same hostname as its parent seems to be completely expected behaviour, hard to imagine people assuming otherwise

georgepsarakis:
I was mostly asking if you think that this will cause an issue for anyone that uses SQS and upgrading to the upcoming release of 4.0.3. If yes, then perhaps we should include this fix as well.

alukach:
@georgepsarakis Ah, I do think that this is an issue introduces by the commit referenced in the original issue which would break anyone's system that doesn't explicitly define the SQS endpoint (which, I'd be surprised if anyone actually does specify the endpoint rather than simply specifying the `region`).

Admittedly, I'm really only looking at this from the perspective of code from #693, I would recommend against deploying any release that includes #693 and #715 without this PR.

georgepsarakis:
@alukach thanks for your efforts! @thedrow @auvipy do you think we can merge just this fix for the 4.0.3 release?

georgepsarakis:
By the way, @alukach if you happen to have any comments for this [Pull Request](https://github.com/celery/celery/pull/3983), they will be more than welcome :smiley: !

alukach:
@auvipy is that something I could help out with or are those done in-house? Would it be a new entry in the CHANGELOG?

auvipy:
yes

auvipy:
from now on we shouldn't merged any pr without release note [if there is any]

alukach:
@auvipy The last entry in the `Changelog` is:

> 
> .. _version-4.0.2:
> 
> 4.0.2
> =====
> :release-date: 2016-12-15 03:31 P.M PST
> :release-by: Ask Solem
> 
> - Now depends on :mod:`amqp` 2.1.4
> 
>     This new version takes advantage of TCP Keepalive settings on Linux,
>     making it better at detecting closed connections, also in failover
>     conditions.
> 
> - Redis: Priority was reversed so, e.g. priority 0 became priority 9.
> 
> .. _version-4.0.1:

I'm assuming you're suggesting adding an entry for `4.0.3`. This PR is a bug fix for and issue raised on master.  It feels strange to report both the creation and resolution of a bug within the same release (we may as well just pretend it never happened).  There may be other features that will go into `4.0.3`, however I think organizing and recording those changes is a bit out of scope for this PR.

georgepsarakis:
@auvipy @alukach thanks :smile: !

PR #740, opened 2017-05-17
add connection kwargs for sentinel

strawposter:
........
class Sentinel(object):
"""
........
connection_kwargs are keyword arguments that will be used when
establishing a connection to a Redis server.
"""
........

#example config
BROKER_URL = 'sentinel://host1:26379;sentinel://host2:26379;sentinel://host3:26379'
RESULT_BACKEND = 'sentinel://host1:26379;sentinel://host2:26379;sentinel://host3:26379'
BROKER_TRANSPORT_OPTIONS = {
                                                             'master_name': 'mymaster',
                                                             'connection_kwargs': {'db': 10, 'password': 'secret'}
                                                            }

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -1066,7 +1066,8 @@ class SentinelChannel(Channel):
     from_transport_options = Channel.from_transport_options + (
         'master_name',
         'min_other_sentinels',
-        'sentinel_kwargs')
+        'sentinel_kwargs',
+        'connection_kwargs')
 
     connection_class = sentinel.SentinelManagedConnection if sentinel else None
 
@@ -1074,6 +1075,7 @@ class SentinelChannel(Channel):
         connparams = self._connparams(async)
 
         additional_params = connparams.copy()
+        additional_params = getattr(self, 'connection_kwargs', additional_params)
 
         additional_params.pop('host', None)
         additional_params.pop('port', None)
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -1075,7 +1075,9 @@ class SentinelChannel(Channel):
         connparams = self._connparams(async)
 
         additional_params = connparams.copy()
-        additional_params = getattr(self, 'connection_kwargs', additional_params)
+        additional_params = getattr(self,
+                                    'connection_kwargs',
+                                    additional_params)
 
         additional_params.pop('host', None)
         additional_params.pop('port', None)
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -1074,10 +1074,9 @@ class SentinelChannel(Channel):
     def _sentinel_managed_pool(self, async=False):
         connparams = self._connparams(async)
 
-        additional_params = connparams.copy()
         additional_params = getattr(self,
                                     'connection_kwargs',
-                                    additional_params)
+                                    connparams.copy())
 
         additional_params.pop('host', None)
         additional_params.pop('port', None)


kombu/transport/redis.py
```diff
@@ -1066,14 +1066,18 @@ class SentinelChannel(Channel):
...
+        'sentinel_kwargs',
+        'connection_kwargs')
 
     connection_class = sentinel.SentinelManagedConnection if sentinel else None
 
     def _sentinel_managed_pool(self, async=False):
         connparams = self._connparams(async)
 
         additional_params = connparams.copy()
+        additional_params = getattr(self,
```
georgepsarakis: Perhaps you should merge here instead of overriding, for better backwards compatibility. What do you think?

strawposter:
So?

kombu/transport/redis.py
```diff
@@ -1066,14 +1066,17 @@ class SentinelChannel(Channel):
...
 
     connection_class = sentinel.SentinelManagedConnection if sentinel else None
 
     def _sentinel_managed_pool(self, async=False):
         connparams = self._connparams(async)
 
-        additional_params = connparams.copy()
+        additional_params = getattr(self,
+                                    'connection_kwargs',
+                                    connparams.copy())
```
georgepsarakis: Perhaps I didn't explain my suggestion very well. So, I meant that perhaps the two dictionaries should be merged, since they both contain different but overlapping sets of parameters:

```
additional_params = connparams.copy()
connection_kwargs = getattr(self, 'connection_kwargs', {})
additional_params.update(connection_kwargs)
```

strawposter: You do not need to update the dictionary if the attribute exists. The attribute will contain all the necessary additional settings for the connection. And if it is not, then default parameters that are contained in connparams.copy () will be used. I think this is more logical.

'BROKER_TRANSPORT_OPTIONS': {
                                             'Master_name': 'mymaster',
                                             'Connection_kwargs': {'db': 10, 'password': 'secret'}
                                             },

'BROKER_TRANSPORT_OPTIONS': {
                                             'Master_name': 'mymaster',
                                             'Connection_kwargs': {'db': 10, 'password': 'secret'}
                                             },

georgepsarakis: OK, then perhaps should be clarified in the [documentation](https://github.com/celery/celery/blob/master/docs/userguide/configuration.rst#redis-backend-settings)?

strawposter: Of course, but I propose to wait a little before updating the documentation.
A few days ago, I sent a request for adding sentinel support for the backend:
https://github.com/celery/celery/pull/4053
The problem is that my backend code is not very good. I hope that someone
from the community will do it better than I, or tell me how to rewrite it, but the code for the kombu is not bad).

strawposter:
Hi, George.
Sentinel support added: "Adding redis sentinel backend # 4144".
Now nothing prevents to make merge for "add connection kwargs for sentinel # 740"

kombu/transport/redis.py
auvipy: so what's the resolution ?

strawposter:
Yes. it should be added, works

auvipy:
can you add test for this new feature?

strawposter:
please excuse me, but now I do not have free time, maybe later

auvipy:
@georgepsarakis should we include this now without test? i am in favor of getting this in though

georgepsarakis:
@auvipy perhaps I am wrong, but I cannot see why this change is required. From [here](https://github.com/celery/kombu/blob/488ed977e0345eb0f509f020a80bb88f42bb1315/kombu/transport/redis.py#L876) it seems that a [copy of all the necessary connection parameters](https://github.com/celery/kombu/blob/488ed977e0345eb0f509f020a80bb88f42bb1315/kombu/transport/redis.py#L1076) will be passed correctly.

auvipy:
closing then. reopen if it don't work

PR #743, opened 2017-05-29
Fixed broken link to AMQP documentation

atombrella:
The linked document was the first result on Google for the title of the link.

I'm not sure how the nice title for the replaced link was created (there's another bit.ly link with a hashed string pointing to Wikipedia).

### DIFF
--- a/kombu/entity.py
+++ b/kombu/entity.py
@@ -98,7 +98,7 @@ class Exchange(MaybeChannelBound):
 
 
                 .. _`AMQP in 10 minutes: Part 4`:
-                    http://bit.ly/amqp-exchange-types
+                    https://bit.ly/2rcICv5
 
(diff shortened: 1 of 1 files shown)


kombu/entity.py
```diff
@@ -98,7 +98,7 @@ class Exchange(MaybeChannelBound):
 
 
                 .. _`AMQP in 10 minutes: Part 4`:
-                    http://bit.ly/amqp-exchange-types
+                    http://bit.ly/2rcICv5
```
georgepsarakis: Could you change it to HTTPS too? I see that it responds to HTTPS as expected.

atombrella: Yep. Would it be fine to switch to HTTPS in the same commit for the other bit.ly link (Lamport's distributed mutex)?

PR #744, opened 2017-05-29
[amqp][ssl] Add ability to validate a server host

kbespalov:
To validate a servers hostname during ssl handshake according with SAN/SNI,  the next ssl option must be provided explicitly:

* ssl opt: 
    - `server_hostname=my-rabbitmq-host`

For example: 

```
Connection('amqp://myrabbit', ssl={server_hostname="myrabbit"})
```
In case then you are passing as kombu connection URI  the list of rabbitmq brokers runned under TLS:

```
Connection('amqp://guest:guest@host-a:5671,guest:guest@host-b:5671', ssl={server_hostname="??????"})
```

there are no any possibility to set the new `server_hostname` into the ssl options each time when kombu under the hood selects a new host and do reconnection according with failover strategy.

To make it possible, we can add `server_hostname` explicitly (if user is not set it) in the ssl options dict.

### DIFF
--- a/kombu/transport/pyamqp.py
+++ b/kombu/transport/pyamqp.py
@@ -114,6 +114,8 @@ class Transport(base.Transport):
                 setattr(conninfo, name, default_value)
         if conninfo.hostname == 'localhost':
             conninfo.hostname = '127.0.0.1'
+        if conninfo.ssl and 'server_hostname' not in conninfo.ssl:
+            conninfo.ssl['server_hostname'] = conninfo.hostname
         opts = dict({
             'host': conninfo.host,
             'userid': conninfo.userid,


thedrow:
@kbespalov ping?

inakimalerba:
Hi. 

Just bumped into this same problem. After a long debugging we realized we were missing the `server_hostname` parameter.

+1 to merge the proposed patch.

thedrow:
> Hi.
> 
> Just bumped into this same problem. After a long debugging we realized we were missing the `server_hostname` parameter.
> 
> +1 to merge the proposed patch.

There's no need for this kind of message. You can use the reactions to upvote.

If you'd like to advance this patch, please provide test coverage for it.

ccorbacho:
@thedrow @auvipy How would tests be added to this though?

Looking at the setup of the current Kombu tests in tox.ini, RabbitMQ is being run on port 5671 (no TLS). So the amqps path is not being tested at all that I can see - I imagine trying to do this would be quite a lot of overhead, and feels like a bigger job outside of this PR (figuring out self-signed CAs and deploying those in the tests to check this code path, and then figuring out how to run different host names to certs locally).

The other alternative would be something with mocks, to verify this is being set correctly. (As really all this change is doing is passing an extra parameter along through to amqp.transport.SSLTransport, which already supports this).

auvipy:
> The other alternative would be something with mocks, to verify this is being set correctly. (As really all this change is doing is passing an extra parameter along through to amqp.transport.SSLTransport, which already supports this).

I think this should be the way

PR #760, opened 2017-07-09
Fix infinite loop in create_loop

gabriel-amram:
fixes https://github.com/celery/celery/issues/3712 

Before handling the todo items we "freeze" them by copying them aside and clearing the list.
This way if an item in the todo list appends a new callable to the list itself it will be taken care of in the next iteration of the parent loop instead of producing an infinite loop by adding it to the list we're running on.

### DIFF
--- a/kombu/async/hub.py
+++ b/kombu/async/hub.py
@@ -283,7 +283,7 @@ class Hub(object):
             # we will break early and allow execution of readers and writers.
             current_todos = len(todo)
             for _ in itertools.repeat(None, current_todos):
-                if not todos:
+                if not todo:
                     break
 
                 item = todo.pop()
--- a/kombu/async/hub.py
+++ b/kombu/async/hub.py
@@ -269,13 +269,18 @@ class Hub(object):
         consolidate = self.consolidate
         consolidate_callback = self.consolidate_callback
         on_tick = self.on_tick
-        todo = self._ready
         propagate = self.propagate_errors
 
         while 1:
             for tick_callback in on_tick:
                 tick_callback()
-
+            
+            # To avoid infinite loop where one of the callables adds items to self._ready
+            # (via call_soon or otherwise), we copy the todo list aside and clear the _ready
+            # so items added don't affect this loop and instead they'll be taken care of on the
+            # next itermation of the parent-loop
+            todo = self._ready.copy()
+            self._ready.clear()
             while todo:
                 item = todo.pop()
                 if item:
--- a/kombu/async/hub.py
+++ b/kombu/async/hub.py
@@ -274,10 +274,11 @@ class Hub(object):
         while 1:
             for tick_callback in on_tick:
                 tick_callback()
-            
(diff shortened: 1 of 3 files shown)


gabriel-amram:
@thedrow

gabriel-amram:
The failing test in travis isn't clear
the test that fails is `t/unit/async/test_hub.py::test_Hub::test_loop__tick_callbacks` and the way it is considered successful is if the tick callbacks were called and then a `pop` was called on the `_ready` `Mock` object (which raises a `RuntimeError`, apparently to stop the test from continuing the execution of the loop). 
But with the new code suggested here, we shouldn't want that to happen, and so instead we might want to change the test to fail when `copy()` is called on the `_ready` `Mock` object.

I'm changing the test code as well.

kombu/async/hub.py
```diff
@@ -269,13 +269,18 @@ def create_loop(self,
...
 
         while 1:
             for tick_callback in on_tick:
                 tick_callback()
-
+            
+            # To avoid infinite loop where one of the callables adds items to self._ready
+            # (via call_soon or otherwise), we copy the todo list aside and clear the _ready
+            # so items added don't affect this loop and instead they'll be taken care of on the
+            # next itermation of the parent-loop
```
thedrow: itermation -> iteration

thedrow:
@gabriel-amram Check out https://github.com/gabriel-amram/kombu/pull/2
I haven't tested it yet. Let me know if it fixes the problem as well.

gabriel-amram:
@thedrow 
I've committed the changes you have suggested.
The landscape fails but doesn't show any problems with this pr
The apicheck fails, but also fails in the master branch, so it must be a configuration problem.

By requesting to add an entry to the changelog - I looked at previous PRs and I don't see people doing that so I couldn't use them as an example.. Should I just add to the Changelog file at the 4.0.3 section what I fixed?

Thanks!

thedrow:
We want to release as soon as we get access to PyPi so keeping the release notes in sync is crucial. 

Does my fix work as well? Have you tested it? 

Yes, please add an entry on 4.0.3.

gabriel-amram:
@thedrow 
Yes, the fix works as expected and resolves the problem.
Added an entry to the changelog as requested.

thedrow:
Thanks a lot for your contribution. Feel free to add yourself to the AUTHORS file.

floatdrop:
Unfortunately - this PR is not quite enough to fix infinite loop. Consider this scenario:

- Take first item from `todo`
- This task schedules itself back to `._ready`
- Since todo is reference to `._ready` we will take it again

@gabriel-amram @thedrow

auvipy:
thanks for investigating. it would be great if you could come up with a pr or 2 separate ones with your propsed fixes

floatdrop:
> thanks for investigating. it would be great if you could come up with a pr or 2 separate ones with your propsed fixes

Sure. We are still in investigation process, but fix will be in #923

PR #762, opened 2017-07-17
- Bugfix: when create a Queue object, can't create binding with Exchange object.

cha-messy:
If i create a Queue with list of Exchange object, it will cause exception, because two bind function have incompatible parameters.


thedrow:
Can you please provide us a test that reproduces this behaviour and a bit more background on what the problem is?

cha-messy:
Sorry for my poor feedback, I' ll finish it little bit later because I'm kind busy right now.

dralley:
@cha-messy , any update?

PR #769, opened 2017-07-18
Pass connection options from broker_options to default_channel 

### LINKED ISSUE #765: How to best set max_retries for system? AKA How to implement circuit breaker pattern?
I'm interested in implementing a [circuit breaker design pattern](https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern) with Celery in my codebase. This means that my application should be able to start if the broker is down and it should be able to raise an exception (within a reasonable window) when attempting to schedule a task. My setup is written so that there is a strong separation between task providers and task consumers (they are separate codebases). This issue is only about task providers, I accept that a task consumer should crash if it can't connect to a broker.

My task provider codebase is a Django server. I'm using SQS as a broker. It is configured to have a virtual topic exchange that schedules tasks to be the worker queues and a queue that's read from by a process responsible for recording scheduled tasks in a database. To ensure that the virtual exchange is registered at server startup, I run the following code:

```python
from django.apps import AppConfig


class TasksConfig(AppConfig):
    name = 'tasks'

    def ready(self):
        from .celery import app
        app.autodiscover_tasks(force=True)
        with app.producer_or_acquire() as P:
            # Ensure all queues are noticed
            for q in app.amqp.queues.values():
                P.maybe_declare(q)
```

This code goes through `kombu.connection.Channel.default_channel` which runs `self.ensure_connection()` with no `max_retries` specified:

```python
class Connection(object):
    # ...
    @property
    def default_channel(self):
        # make sure we're still connected, and if not refresh.
        self.ensure_connection()
        if self._default_channel is None:
            self._default_channel = self.channel()
        return self._default_channel
```

This allows `utils.functional.retry_over_time()` to retry indefinitely, causing the application to hang on startup. A very similar chain of actions occurs when calling `.delay()` on a task function. It feels like the simplest solution here is to provide a `max_retries` arguments.  However, **my question is "how should I best deliver this configuration to the channel?"**  I don't believe the `Channel` has access to the Celery app. It's feeling like providing it as a `transport_options` entry will be the simplest/cleanest method of configuration.  Any input on this before I make a PR to support this `transport_options` attribute?

alukach:
Not sure if this is the best implementation, however it gets the job done for my needs. My goal is to be able to start my application when the message broker is offline. Upon startup, the application attempts to establish queues, which requires the `Channel.default_connection` to be generated. When generating the `default_channel`, `Channel.ensure_connection` with no `max_retry` (or other options) provided. This caused the application to hang while infinitely attempting to connect to the broker.

Fixes #765

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -815,18 +815,20 @@ class Connection(object):
             a connection is passed instead of a channel, to functions that
             require a channel.
         """
-        ensure_connection_options = {}
-        if self.transport_options:
-            options_keys = [
-                'max_retries',
-                'interval_start', 'interval_step', 'interval_max',
-            ]
-            for k in options_keys:
-                if k in self.transport_options:
-                    ensure_connection_options[k] = self.transport_options[k]
+        conn_opts = {}
+        transport_opts = self.transport_options
+        if transport_opts:
+            if 'max_retries' in transport_opts:
+                conn_opts['max_retries'] = transport_opts['max_retries']
+            if 'interval_start' in transport_opts:
+                conn_opts['interval_start'] = transport_opts['interval_start']
+            if 'interval_step' in transport_opts:
+                conn_opts['interval_step'] = transport_opts['interval_step']
+            if 'interval_max' in transport_opts:
+                conn_opts['interval_max'] = transport_opts['interval_max']
 
         # make sure we're still connected, and if not refresh.
-        self.ensure_connection(**ensure_connection_options)
+        self.ensure_connection(**conn_opts)
         if self._default_channel is None:
             self._default_channel = self.channel()
         return self._default_channel
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -815,14 +815,18 @@ class Connection(object):
             a connection is passed instead of a channel, to functions that
             require a channel.
         """
-        options = {
-            k: v for k, v in (self.transport_options or {}).items()
-            if k in [
-                'max_retries', 'interval_start',
-                'interval_step', 'interval_max']
-        }
+        ensure_connection_options = {}
+        if self.transport_options:
+            options_keys = [
+                'max_retries',
+                'interval_start', 'interval_step', 'interval_max',
+            ]
+            for k in options_keys:
+                if k in self.transport_options:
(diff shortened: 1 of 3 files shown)


alukach:
@thedrow Thanks for the review. Please see changes.

kombu/connection.py
```diff
@@ -815,8 +815,18 @@ def default_channel(self):
             a connection is passed instead of a channel, to functions that
             require a channel.
         """
+        ensure_connection_options = {}
+        if self.transport_options:
+            options_keys = [
+                'max_retries',
+                'interval_start', 'interval_step', 'interval_max',
+            ]
+            for k in options_keys:
```
thedrow: There's still no need for a loop here.
Please hardcode the values as suggested in the last review.

alukach: Can you help me understand your reasoning for the aversion to using a loop?

I would recommend against using `transport_options.get('foo', None)`, being that the default response if no key is provided will override the defaults provided in the `ensure_connection` signature:

```python

    def ensure_connection(self, errback=None, max_retries=None,
                          interval_start=2, interval_step=2, interval_max=30,
                          callback=None, reraise_as_library_errors=True):
```

We could first check if the key is present in the options, however I feel that this is a bit repetitive. However, if you're against the loop for the sake of readability, then this is probably the best way to go.

```python
if 'max_retries' in self.transport_options:
    opts['max_retries'] = self.transport_options['max_retries']
if 'interval_start' in self.transport_options:
    opts['interval_start'] = self.transport_options['interval_start']
if 'interval_step' in self.transport_options:
    opts['interval_step'] = self.transport_options['interval_step']
if 'interval_max' in self.transport_options:
    opts['interval_max'] = self.transport_options['interval_max']
```

thedrow: You are correct. This is better.

sanjioh:
hi @thedrow, is by any chance a release of kombu 4.1.1 planned anytime soon, with this fix included? It would be nice to have the behaviour of kombu 4.0.2 restored, when used with Celery (see also #795).
thanks!

thedrow:
I'll consider it.

PR #774, opened 2017-07-27
EPOCH apply timezone

shn7798:
celery task ETA calc with this function, but it only works when timezone is 'UTC'

### DIFF
--- a/kombu/async/timer.py
+++ b/kombu/async/timer.py
@@ -41,7 +41,8 @@ def to_timestamp(d, default_timezone=utc, time=monotonic):
         if d.tzinfo is None:
             d = d.replace(tzinfo=default_timezone)
         diff = _time() - time()
-        return max((d - EPOCH).total_seconds() - diff, 0)
+        epoch = datetime.utcfromtimestamp(0).replace(tzinfo=default_timezone)
+        return max((d - epoch).total_seconds() - diff, 0)
     return d
 
 


georgepsarakis:
Before merging this we need to consider the discussion in https://github.com/celery/celery/issues/4160 .

kombu/async/timer.py
```diff
@@ -41,7 +41,8 @@ def to_timestamp(d, default_timezone=utc, time=monotonic):
         if d.tzinfo is None:
             d = d.replace(tzinfo=default_timezone)
         diff = _time() - time()
-        return max((d - EPOCH).total_seconds() - diff, 0)
+        epoch = datetime.utcfromtimestamp(0).replace(tzinfo=default_timezone)
```
rtpg: I don't believe that the timezone replacement here is proper. This won't convert the time to the new timezone, but simply replace the internal timezone.

So 9 AM UTC will "become" 9 AM JST.

Likewise, the epoch here might be transformed into "epoch + 9 hours". 

what you can go for instead is `default_timezone.normalize(datetime.utcfromtimestamp(0)`

shn7798: This patch make sure the `(d - epoch).total_seconds()` is calcing in same timezone

rtpg:
```

In [22]: from pytz import UTC

In [23]: from pytz.reference import Eastern

In [24]: from datetime import datetime

In [25]: datetime.utcfromtimestamp(0).replace(tzinfo=Eastern) - datetime.utcfrom
    ...: timestamp(0).replace(tzinfo=UTC)
Out[25]: datetime.timedelta(0, 18000)
```

`replace`ing here means that `epoch` is _no longer_ the the "proper" epoch. 

The whole point of storing timezone info in datetimes is to do math without worrying about timezones... so here if you need timezone-aware datetimes, you could just do `epoch = datetime.utcfromtimestamp(0).replace(tzinfo=UTC)` to get the "timezone-aware" epoch.

georgepsarakis:
@shn7798 can we close this now?

PR #783, opened 2017-08-13
Removed some unnecessary variable assignments

atombrella:
I added a comment in `kombu/connection.py`. I noticed a lot of unused parameters in functions and methods, and in a few cases the index-variable `i` in for-loops could be replaced with `_`.

### DIFF
--- a/kombu/compat.py
+++ b/kombu/compat.py
@@ -85,7 +85,6 @@ class Consumer(messaging.Consumer):
     durable = True
     exclusive = False
     auto_delete = False
-    exchange_type = 'direct'
     _closed = False
 
     def __init__(self, connection, queue=None, exchange=None,
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -173,7 +173,7 @@ class Connection(object):
                 # e.g. sqla+mysql://root:masterkey@localhost/
                 params['transport'], params['hostname'] = \
                     hostname.split('+', 1)
-                transport = self.uri_prefix = params['transport']
+                self.uri_prefix = params['transport']
             else:
                 transport = transport or urlparse(hostname).scheme
                 if not get_transport_cls(transport).can_parse_url:
--- a/kombu/utils/functional.py
+++ b/kombu/utils/functional.py
@@ -324,7 +324,6 @@ def retry_over_time(fun, catch, args=[], kwargs={}, errback=None,
         interval_max (float): Maximum number of seconds to sleep
             between retries.
     """
-    retries = 0
     interval_range = fxrange(interval_start,
                              interval_max + interval_start,
                              interval_step, repeatlast=True)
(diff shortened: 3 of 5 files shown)


kombu/compat.py
```diff
@@ -102,8 +100,8 @@ def __init__(self, connection, queue=None, exchange=None,
 
         self.queue = queue or self.queue
         self.exchange = exchange or self.exchange
-        self.exchange_type = exchange_type or self.exchange_type
-        self.routing_key = routing_key or self.routing_key
+        self.exchange_type = exchange_type or 'direct'
```
georgepsarakis: Is this change an enhancement for readability? This way the default values will be used from the class variables.

kombu/connection.py
```diff
@@ -173,7 +173,7 @@ def __init__(self, hostname='localhost', userid=None,
                 # e.g. sqla+mysql://root:masterkey@localhost/
                 params['transport'], params['hostname'] = \
                     hostname.split('+', 1)
-                transport = self.uri_prefix = params['transport']
+                transport = self.uri_prefix = params['transport']  # unused
```
georgepsarakis: I assume that `transport` is the variable not used elsewhere after this statement.

kombu/utils/functional.py
```diff
@@ -324,7 +324,6 @@ def retry_over_time(fun, catch, args=[], kwargs={}, errback=None,
         interval_max (float): Maximum number of seconds to sleep
             between retries.
     """
-    retries = 0
```
georgepsarakis: Seems unused indeed.

kombu/compat.py
atombrella: `self.exchange_type` was defined twice, I pushed an update to remove both. I think the proposed change is easier to read, because the default value is on the same line. If I'm pointed to a suitable place, I can add a test.

kombu/connection.py
atombrella: There's also the possibility that line 187: `params['transport'] = transport` is indented a level too much?

atombrella: @georgepsarakis If I revert this change, and add a `TODO` (or `FIXME`) note with my guess about indentation level of line 187, are you willing to merge the two other changes?

georgepsarakis: For your previous comment, I do not think that this is the case, because this function is regularly used to parse URLs, so it would need a failing test to prove that something is wrong.
Perhaps change the line to `self.uri_prefix = params['transport']` ?

atombrella: I updated that.

georgepsarakis:
@auvipy I do not see any testing scope for the applied changes, I believe it is fine. Regarding the Travis CI failures, hopefully they will pass on `master` after rebase.

PR #785, opened 2017-08-18
Make messages durable by default in qpid

daviddavis:
We are seeing messages disappear in durable queues when restarting
qpid. This coincides with the rabbitmq code (e.g. https://git.io/v7jNV).

### DIFF
--- a/kombu/transport/qpid.py
+++ b/kombu/transport/qpid.py
@@ -427,7 +427,8 @@ class Channel(base.StdChannel):
             rx.close()
         return message
 
-    def _put(self, routing_key, message, exchange=None, **kwargs):
+    def _put(self, routing_key, message, exchange=None, durable=True,
+             **kwargs):
         """Synchronously send a single message onto a queue or exchange.
 
         An internal method which synchronously sends a single message onto
@@ -460,6 +461,9 @@ class Channel(base.StdChannel):
             should be sent on. If no exchange is specified, the message is
             sent directly to a queue specified by routing_key.
         :type exchange: str
+        :keyword durable: whether or not the message should persist or be
+            durable.
+        :type durable: bool
 
         """
         if not exchange:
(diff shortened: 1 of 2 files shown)


thedrow:
Thanks for your contribution. Feel free to add yourself to the CONTRIBUTORS file.

bmbouter:
@thedrow we want to get this commit onto the 4.0 branch. Should we do that with a cherry pick? Also does it need to go on any branch besides 4.0 and master?

thedrow:
Master is good enough. We'll make a new release from it.

bmbouter:
@thedrow When you say "a new release" do you mean the very next 4.0.z release? I think the next one is 4.0.3. I want to be sure because we're about to apply the patch onto packaged (Fedora) versions of python-kombu, so I'm wondering how long this patch will need to be carried downstream.

Also more generally, do you just want all patches to land on master and we never have to merge to another branch? The branching model of both celery and kombu are confusing. I never know where things are supposed to land. There are branches like 3.0, 4.0, 5.0-devel, and also master. Some clarity here would help us contribute without asking every time if we've got it all on the right branches.

Thanks for your work on celery!

thedrow:
The next version is 4.2.0. We've already released 4.1.0.
We'll merge master to the 4.0 branch before the release.

bmbouter:
@thedrow Thanks for the info. So to confirm, there won't be a 4.1.1 and the next release is 4.2.0?

Also, do you have an idea of when the next Kombu release with this patch will be? We're trying to figure out when the next release is to see if its worth the effort to patch it in the packaged versions in Fedora.

thedrow:
I have no ETA. I don't have time to manage a new release.
Maybe @auvipy can release only Kombu 4.1.1. We have another fix in it so it could be worth it.

bmbouter:
@thedrow Thanks. Also if there is documentation on how we could help we could maybe help.

@auvipy If you are able to cut a 4.1.1 release, or if there is something we can do to help do that, please let us know. Thank you both.

thedrow:
@bmbouter We can give you permission to PyPi. See https://github.com/celery/celery/issues/4183

PR #786, opened 2017-08-20
Update links in documentation to either https or to new locations

atombrella:
This forwards automatically to the newest Python source, i.e., 3.6. 2.6 is not supported anymore (although it was introduced in 2.6, I've noticed), so I replaced it with 2.7.

`Python 2.7+`: https://docs.python.org/library/json.html

`apicheck` fails both locally, and remote. I'm not sure how to fix that.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,7 +6,7 @@
 
 :Version: 4.1.0
 :Web: http://kombu.me/
-:Download: http://pypi.python.org/pypi/kombu/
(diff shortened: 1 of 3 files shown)


thedrow:
We'll fix the apicheck failure soon. Disregard it for now.
Thanks for your contribution! Feel free to add yourself to the CONTRIBUTORS file.

Issue #784, opened 2017-08-18
Redis 2.10.4 or greater is required, not 2.10.0


colinjeanne:
Attempting to use Redis 2.10.3 with kombu will fail with `TypeError: can_read() got an unexpected keyword argument 'timeout'`. Redis only added the `timeout` parameter to this function in 2.10.4.

georgepsarakis:
The required version seems correct in the related Celery extras [requirements file](https://github.com/celery/celery/blob/master/requirements/extras/redis.txt). You are right, the [Kombu requirements file](https://github.com/celery/kombu/blob/master/requirements/extras/redis.txt) is wrong however.
Would you like to open a Pull Request for this change?

colinjeanne:
Sure, I'll get to this this week.

colinjeanne:
This was fixed by #788

PR #788, opened 2017-08-21
Require Redis 2.10.4 or greater

colinjeanne:
The Redis transport uses the `can_read`'s `timeout` parameter. This parameter was added in 2.10.4.

This fixes GitHub issue #784.


kombu/transport/redis.py
```diff
@@ -939,9 +939,9 @@ def _get_pool(self, async=False):
         return redis.ConnectionPool(**params)
 
     def _get_client(self):
-        if redis.VERSION < (2, 10, 0):
+        if redis.VERSION < (2, 10, 4):
```
georgepsarakis: Celery uses version `>=2.10.5`, so I would suggest to use the same here, we use since `kombu` is most frequently used from `celery`.

thedrow:
Thanks! Feel free to add yourself to the CONTRIBUTORS file :)

PR #790, opened 2017-08-28
Fix issue #789: Async http code not allowing for proxy config

### LINKED ISSUE #789: No proxy support using SQS
The async HTTP code is not allowing the proxy environment variables to be used by libcurl

It seems to be due to this code in `async/http/curl.py`

```
        if request.proxy_host:
            if not request.proxy_port:
                raise ValueError('Request with proxy_host but no proxy_port')
            setopt(_pycurl.PROXY, request.proxy_host)
            setopt(_pycurl.PROXYPORT, request.proxy_port)
            if request.proxy_username:
                setopt(_pycurl.PROXYUSERPWD, '{0}:{1}'.format(
                    request.proxy_username, request.proxy_password or ''))
        else:
            setopt(_pycurl.PROXY, '')
            curl.unsetopt(_pycurl.PROXYUSERPWD)
```
Notice that if the original request does not specify a `proxy_host` then it actually sets it to empty in the curl OPTs which makes libcurl NOT read the environment for it either.

And the caller in this case is the SQS transport which does NOT pass any proxy setting in either so when using SQS there doesn't appear to be a way to make Kombu use a proxy for the async HTTP. 

(Note: The calls through Boto DO work because Boto reads it from the environment variables by default)

jkoehl:
Fixes #789  

Don't reset the proxy config in the async curl based http request building unless that is specified by the caller

### DIFF
--- a/kombu/asynchronous/http/curl.py
+++ b/kombu/asynchronous/http/curl.py
@@ -231,9 +231,6 @@ class CurlClient(BaseClient):
             if request.proxy_username:
                 setopt(_pycurl.PROXYUSERPWD, '{}:{}'.format(
                     request.proxy_username, request.proxy_password or ''))
-        else:
-            setopt(_pycurl.PROXY, '')
-            curl.unsetopt(_pycurl.PROXYUSERPWD)
 
         setopt(_pycurl.SSL_VERIFYPEER, 1 if request.validate_cert else 0)
         setopt(_pycurl.SSL_VERIFYHOST, 2 if request.validate_cert else 0)
--- a/t/unit/asynchronous/http/test_curl.py
+++ b/t/unit/asynchronous/http/test_curl.py
@@ -1,4 +1,5 @@
-from unittest.mock import Mock, call, patch
+from io import BytesIO
+from unittest.mock import ANY, Mock, call, patch
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


jkoehl:
In theory if someone had setup their environment to use a proxy and this code used to ignore that and bypass the proxy. They may have had a reason they wanted to bypass the proxy for Celery but use a proxy for other things but this behavior is kind of not obvious.

For anyone actually needing to use a proxy to get to the queue this change is needed. In my case we use AWS SQS from a VPC and have to use a proxy to get to SQS.

auvipy:
I would be looking for a testcase to verify that this change wont introduce regression

auvipy:
please add a testcase for the change

aljets:
@auvipy can you take another look? Thanks!

auvipy:
oh sure. I have re started the CI! thanks for the push!

auvipy:
> @auvipy can you take another look? Thanks!

were you able to make it work in your project?

aljets:
> were you able to make it work in your project using this patch?

Yes, we've actually been using the patch for several years now via a fork for version 4.4.0 on python 2.7 and now for several weeks on version 5.2.2 on python 3.7. Attempting to get it into a release so we can stop using the work :)

auvipy (approved):

auvipy:
thanks a lot!

PR #792, opened 2017-08-28
Fixes #791: SQS Queue names getting mangles

### LINKED ISSUE #791: SQS queue name gets mangled
Using Celery/Kombu in an environment with _future_ strings support you end up with mangled queue names in SQS.

The code that tries to translate invalid SQS queue characters causes the byte string to get converted to a string representation such as b'my_queue_name' and then translated that to 'b_my_queue_name'

I am finding this because we are using Apache Airflow in a Python 2.7 environment using SQS queues as the broker.

jkoehl:
Fixes #791 by slightly changing the order of when the queue name is translated to SQS valid name and when the unicode to str encoding happens. Basically doing the translate() on the original string instead of doing it after we have translated the string into an encoded unicode string.

### DIFF
--- a/kombu/async/http/curl.py
+++ b/kombu/async/http/curl.py
@@ -218,9 +218,6 @@ class CurlClient(BaseClient):
             if request.proxy_username:
                 setopt(_pycurl.PROXYUSERPWD, '{0}:{1}'.format(
                     request.proxy_username, request.proxy_password or ''))
-        else:
-            setopt(_pycurl.PROXY, '')
-            curl.unsetopt(_pycurl.PROXYUSERPWD)
 
         setopt(_pycurl.SSL_VERIFYPEER, 1 if request.validate_cert else 0)
         setopt(_pycurl.SSL_VERIFYHOST, 2 if request.validate_cert else 0)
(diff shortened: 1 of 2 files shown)


kombu/async/http/curl.py
```diff
@@ -218,9 +218,6 @@ def _setup_request(self, curl, request, buffer, headers, _pycurl=pycurl):
             if request.proxy_username:
                 setopt(_pycurl.PROXYUSERPWD, '{0}:{1}'.format(
                     request.proxy_username, request.proxy_password or ''))
-        else:
-            setopt(_pycurl.PROXY, '')
```
georgepsarakis: Can you please explain this change? It seems unrelated to this PR.

kombu/transport/SQS.py
```diff
@@ -154,10 +154,10 @@ def entity_name(self, name, table=CHARS_REPLACE_TABLE):
         """Format AMQP queue name into a legal SQS queue name."""
         if name.endswith('.fifo'):
             partial = name.rstrip('.fifo')
-            partial = text_t(safe_str(partial)).translate(table)
+            partial = text_t(safe_str(partial.translate(table)))
```
georgepsarakis: Could you please a very simple test case to verify this change? You already have an example in https://github.com/celery/kombu/issues/791 , so you could use that.

kombu/async/http/curl.py
jkoehl: Sorry you are correct. That is a change for another PR. I branched from the wrong commit. I will fix that so this PR is only about this one issue.

jkoehl:
I am going

PR #794, opened 2017-08-30
Fixes #791 - SQS queue name gets mangled in Python 2.7 environment

### LINKED ISSUE #791: SQS queue name gets mangled

jkoehl:
This change alters the safe_str() function to handle the case when a python-future package "newstr" and "newbytes" are being used in Python 2.7


thedrow:
@jkoehl Any news on this one?

PR #797, opened 2017-09-06
fixed import path to SimpleQueue and SimpleBuffer in docs

wmatyskiewicz:


georgepsarakis:
Good catch! The Travis failures are unrelated so I merging this.

PR #798, opened 2017-09-08
Fix Pyro link

gordol:
it's pythonhosted.org not pythonhosting.org


thedrow:
The correct link is http://pythonhosted.org/Pyro4/ since that's the maintained version we use. Thanks for the heads up :)

Issue #746, opened 2017-05-31
Kombo + SQS + PyCurl, 'NoneType' object has no attribute 'call_repeatedly'


jonzlin95:
Hi. I'm running the lastest Kombu master (w/ stable 4.0.2 Celery), SQS Broker, with Django DB Backend, and I run into this error.

```
[2017-05-31 08:08:47,205: ERROR/MainProcess] Pending callback raised: AttributeError("'NoneType' object has no attribute 'call_repeatedly'",)
Traceback (most recent call last):
  File "/home/ubuntu/ies-league/venv/local/lib/python2.7/site-packages/celery/worker/consumer/consumer.py", line 233, in perform_pending_operations
    self._pending_operations.pop()()
  File "/home/ubuntu/ies-league/venv/local/lib/python2.7/site-packages/vine/promises.py", line 139, in __call__
    return self.throw()
  File "/home/ubuntu/ies-league/venv/local/lib/python2.7/site-packages/vine/promises.py", line 136, in __call__
    retval = fun(*final_args, **final_kwargs)
  File "/home/ubuntu/ies-league/venv/local/lib/python2.7/site-packages/kombu/message.py", line 130, in ack_log_error
    self.ack(multiple=multiple)
  File "/home/ubuntu/ies-league/venv/local/lib/python2.7/site-packages/kombu/message.py", line 125, in ack
    self.channel.basic_ack(self.delivery_tag, multiple=multiple)
  File "/home/ubuntu/ies-league/venv/local/lib/python2.7/site-packages/kombu/transport/SQS.py", line 380, in basic_ack
    self.asynsqs.delete_message(message['sqs_queue'],
  File "/home/ubuntu/ies-league/venv/local/lib/python2.7/site-packages/kombu/transport/SQS.py", line 436, in asynsqs
    region=self.region
  File "/home/ubuntu/ies-league/venv/local/lib/python2.7/site-packages/kombu/async/aws/sqs/connection.py", line 27, in __init__
    **kwargs
  File "/home/ubuntu/ies-league/venv/local/lib/python2.7/site-packages/kombu/async/aws/connection.py", line 178, in __init__
    **http_client_params)
  File "/home/ubuntu/ies-league/venv/local/lib/python2.7/site-packages/kombu/async/aws/connection.py", line 151, in __init__
    self._httpclient = http_client or get_client()
[...]

PR #817, opened 2017-12-08
Proposed fix for SQS transport acks and deletes

jay-johnson:
Thanks for putting together such a great project. Kombu is awesome!

I hit this issue and believe this PR can help:
https://github.com/celery/kombu/issues/746

I noticed ``hub._current_http_client`` was not defined on master and the async SQS client was never deleting messages in my SQS test queue. When I called ``message.ack()`` it would respond (after fixing ``get_client()``) but messages would go to **In Flight** instead of being deleted. 

The changes below seem to work, but I do not know the internals of the ``kombu.async.hub`` or how this could impact other SQS actions.

Thanks again!

### DIFF
--- a/kombu/async/http/__init__.py
+++ b/kombu/async/http/__init__.py
@@ -25,14 +25,8 @@ def get_client(hub=None, **kwargs):
 
         return hub._current_http_client
     except AttributeError:
-        
-        # What should happen on error?
-        # It appears to impact the message.ack() going
-        # back to SQS for issues like:
-        # https://github.com/celery/kombu/issues/746
         if not hub:
             hub = Hub()
-            set_event_loop(hub)
 
         client = hub._current_http_client = Client(hub, **kwargs)
         set_event_loop(hub)
--- a/kombu/async/http/__init__.py
+++ b/kombu/async/http/__init__.py
@@ -20,10 +20,10 @@ def get_client(hub=None, **kwargs):
     try:
         if not hub:
             hub = Hub()
+            client = hub._current_http_client = Client(hub, **kwargs)
             set_event_loop(hub)
 
-        client = Client(hub, **kwargs)
-        return client
+        return hub._current_http_client
     except AttributeError:
         
         # What should happen on error?
@@ -34,5 +34,6 @@ def get_client(hub=None, **kwargs):
             hub = Hub()
             set_event_loop(hub)
 
-        client = Client(hub, **kwargs)
-        return client
+        client = hub._current_http_client = Client(hub, **kwargs)
+        set_event_loop(hub)
+        return hub._current_http_client
--- a/kombu/async/http/__init__.py
+++ b/kombu/async/http/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
-from kombu.async import get_event_loop
+from kombu.async import get_event_loop, set_event_loop
+from kombu.async.hub import Hub
 
 from .base import Request, Headers, Response
 
@@ -17,7 +18,21 @@ def get_client(hub=None, **kwargs):
     """Get or create HTTP client bound to the current event loop."""
     hub = hub or get_event_loop()
     try:
-        return hub._current_http_client
+        if not hub:
+            hub = Hub()
+            set_event_loop(hub)
+
+        client = Client(hub, **kwargs)
+        return client
     except AttributeError:
-        client = hub._current_http_client = Client(hub, **kwargs)
+        
+        # What should happen on error?
+        # It appears to impact the message.ack() going
+        # back to SQS for issues like:
+        # https://github.com/celery/kombu/issues/746
+        if not hub:
+            hub = Hub()
+            set_event_loop(hub)
+
+        client = Client(hub, **kwargs)
         return client
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -371,14 +371,16 @@ class Channel(virtual.Channel):
         return super(Channel, self)._restore(message)
 
     def basic_ack(self, delivery_tag, multiple=False):
+        message = {}
+        sqs_message = {}
         try:
             message = self.qos.get(delivery_tag).delivery_info
             sqs_message = message['sqs_message']
         except KeyError:
             pass
         else:
-            self.asynsqs.delete_message(message['sqs_queue'],
-                                        sqs_message['ReceiptHandle'])
+            self._sqs.delete_message(QueueUrl=message['sqs_queue'],
+                                     ReceiptHandle=sqs_message['ReceiptHandle'])
         super(Channel, self).basic_ack(delivery_tag)
 
     def _size(self, queue):


kombu/async/http/__init__.py
```diff
@@ -17,7 +18,16 @@ def get_client(hub=None, **kwargs):
...
+            hub = Hub()
+            client = hub._current_http_client = Client(hub, **kwargs)
+            set_event_loop(hub)
+
         return hub._current_http_client
     except AttributeError:
+        if not hub:
+            hub = Hub()
+
         client = hub._current_http_client = Client(hub, **kwargs)
```
georgepsarakis: It seems that this line would actually set the `_current_http_client` attribute, as it was before. Could you please provide some details on the changes?

kombu/transport/SQS.py
```diff
@@ -371,14 +371,16 @@ def _restore(self, message,
...
     def basic_ack(self, delivery_tag, multiple=False):
+        message = {}
+        sqs_message = {}
         try:
             message = self.qos.get(delivery_tag).delivery_info
             sqs_message = message['sqs_message']
         except KeyError:
             pass
         else:
-            self.asynsqs.delete_message(message['sqs_queue'],
```
georgepsarakis: So the asynchronous client does not work in this case?

jay-johnson:
Hi @georgepsarakis 

I admit this PR needs to be reviewed with implications to downstream clients wanting the async SQS support, but this seems to be broken every single time I try the SQS transport just like the other users on the github issue. The PR fixes were made because there was never a valid, instantiated hub or client object coming out of the ``__init__.py``. The async class was changed to the boto client because the async never deleted the message in SQS (it looked like the needed callback on the async never returned). These might be due to having to install pycurl ``with-nss`` or just how the class I am using is trying to consume a message, but with both file fixes off this PR the test client can ack and delete messages in SQS.

Here's a quick way to reproduce the error and the fix.

Note you'll need to set a valid 
``SQS_AWS_SECRET_KEY``
and
``SQS_AWS_ACCESS_KEY``

Show the current behavior which is SQS does not receive an ack or delete:

```
virtualenv py2venvbroken && source py2venvbroken/bin/activate && pip install celery-connectors && pip uninstall -y pycurl; pip install pycurl --compile --global-option="--with-nss" pycurl && export SQS_AWS_SECRET_KEY=${SQS_AWS_SECRET_KEY} && export SQS_AWS_ACCESS_KEY=${SQS_AWS_ACCESS_KEY} && kombu_sqs_publisher.py && kombu_sqs_subscriber.py
```

Logs:
```
INFO:botocore.vendored.requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): queue.amazonaws.com
INFO:botocore.vendored.requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): queue.amazonaws.com
INFO:kombu-subscriber:kombu-sqs-subscriber - kombu.subscriber queues=test1 consuming with callback=handle_message
INFO:kombu-sqs-subscriber:callback received msg body={u'subscription_id': 222, u'created': u'2017-12-09T14:32:01.076154', u'stripe_id': 333, u'product_id': u'DEF', u'account_id': 111}
INFO:kombu-subscriber:kombu-sqs-subscriber - kombu.subscriber consume hit exception='NoneType' object has no attribute 'call_repeatedly' queue=test1
INFO:kombu-sqs-subscriber:End - kombu-sqs-subscriber
Restoring 1 unacknowledged message(s)
```

Demo the fix - notice the first SQS subscribe consumes the message (ack + delete in the AWS console) and the second attempt waits for a message and then stops after 10 seconds

```
deactivate && virtualenv py2venvwithpr && source py2venvwithpr/bin/activate && pip install celery-connectors && export SQS_AWS_SECRET_KEY=${SQS_AWS_SECRET_KEY} && export SQS_AWS_ACCESS_KEY=${SQS_AWS_ACCESS_KEY} && rm -rf /tmp/prfixbranch; git clone https://github.com/jay-johnson/kombu.git /tmp/prfixbranch && pushd /tmp/prfixbranch && git checkout sqs-http-get-client && pip uninstall -y kombu; pip uninstall -y pycurl; pip install pycurl --compile --global-option="--with-nss" pycurl && python setup.py develop && popd && kombu_sqs_subscriber.py && kombu_sqs_subscriber.py
```

Logs
```
INFO:botocore.vendored.requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): queue.amazonaws.com
INFO:botocore.vendored.requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): queue.amazonaws.com
INFO:kombu-subscriber:kombu-sqs-subscriber - kombu.subscriber queues=test1 consuming with callback=handle_message
INFO:kombu-sqs-subscriber:callback received msg body={u'subscription_id': 222, u'created': u'2017-12-09T14:32:01.076154', u'stripe_id': 333, u'product_id': u'DEF', u'account_id': 111}
INFO:kombu-sqs-subscriber:End - kombu-sqs-subscriber
INFO:kombu-sqs-subscriber:Start - kombu-sqs-subscriber
INFO:kombu-subscriber:setup routing
INFO:botocore.vendored.requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): queue.amazonaws.com
INFO:botocore.vendored.requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): queue.amazonaws.com
INFO:kombu-subscriber:kombu-sqs-subscriber - kombu.subscriber queues=test1 consuming with callback=handle_message
INFO:kombu-sqs-subscriber:End - kombu-sqs-subscriber
```

auvipy:
it would be great if we could get this in 4.2

georgepsarakis:
@jay-johnson sorry for the delayed reply. Will you perhaps be available to review this again?

MoxCheese:
Are there any updates on getting this PR reviewed and merged? I have tested this patch on top of latest Kombu master, using Celery 4.2.0, and it appears to have resolved #746.

I can confirm that the change to not use the regular sqs client, rather than asynsqs to delete messages appears to be necessary for messages to be removed from the queue, but I can't speculate as to why.

nsonic001:
s this issue resolved, or which version is stable?
I am experiencing the same issue.

kombu/async/http/__init__.py
```diff
@@ -17,7 +18,16 @@ def get_client(hub=None, **kwargs):
     """Get or create HTTP client bound to the current event loop."""
     hub = hub or get_event_loop()
```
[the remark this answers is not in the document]
ikonst: https://github.com/celery/kombu/pull/981

galCohen88:
@thedrow messages are being acked with more recent versions, I'm not sure we need to merge this. I will dig into this soon

galCohen88:
I'm closing this PR, since we are in a different point in time of the project. Messages are acknowledged just fine, accessing the `self._sqs` attribute is redundant here, since `def sqs(self, queue=None)` function is responsible exactly for that (introduced in a later point in time)

Issue #781, opened 2017-08-06
Can't shrink pool when in use


arjun27:
Some of our celery tasks throw an RuntimeError on `apply_async`. The error is

```
Can't shrink pool when in use: was=1 now=10
```

Relevant stacktrace:

```
File "./events/utils.py" in event_post_creation_tasks
  850.     process_sdk_events.delay(event_ids, device_id, user_agent)

File "/usr/local/lib/python2.7/site-packages/celery/app/task.py" in delay
  412.         return self.apply_async(args, kwargs)

File "/usr/local/lib/python2.7/site-packages/celery/app/task.py" in apply_async
  535.             **options

File "/usr/local/lib/python2.7/site-packages/celery/app/base.py" in send_task
  734.         with self.producer_or_acquire(producer) as P:

File "/usr/local/lib/python2.7/site-packages/celery/app/base.py" in producer_or_acquire
  863.             producer, self.producer_pool.acquire, block=True,

File "/usr/local/lib/python2.7/site-packages/celery/app/base.py" in producer_pool
  1232.         return self.amqp.producer_pool

File "/usr/local/lib/python2.7/site-packages/celery/app/amqp.py" in producer_pool
  610.             self._producer_pool.limit = self.app.pool.limit

File "/usr/local/lib/python2.7/site-packages/celery/app/base.py" in pool
  1147.             pools.set_limit(limit)

File "/usr/local/lib/python2.7/site-packages/kombu/pools.py" in set_limit
  137.             pool.resize(limit)

File "/usr/local/lib/python2.7/site-packages/kombu/resource.py" in resize
  178.                         limit, self._limit))

Exception Type: RuntimeError at /api/v1/sdk_events/bulk/
Exception Value: Can't shrink pool when in use: was=1 now=10
```

Haven't been able to troubleshoot this through docs or other issues. What am I missing?

arjun27:
Changing `broker_pool_limit` to 10 inside celery config seems to have fixed it

PR #826, opened 2018-01-19
Prevent spurious exception on Resource.resize(0)

AlexHill:
`Resource.resize()` raises an exception if the pool is in use and the new size is smaller than the old size. However, it also raises this exception when the new size is zero, which should correspond to disabling the pool. Instead of shrinking the pool to zero and releasing all resources, we can simply dequeue all resources and forget about them.

Refs #781

### DIFF
--- a/kombu/resource.py
+++ b/kombu/resource.py
@@ -171,11 +171,11 @@ class Resource(object):
 
     def resize(self, limit, force=False, ignore_errors=False, reset=False):
         prev_limit = self._limit
-        if (self._dirty and limit < self._limit) and not ignore_errors:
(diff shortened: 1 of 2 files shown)


AlexHill:
Hi @auvipy & @thedrow - unit test added for this case. Let me know if this needs anything else!

AlexHill:
Question about the Celery release process - are Kombu releases synchronised with Celery releases? i.e. will this patch be released with Celery 4.2?

auvipy:
yes

thedrow:
@AlexHill We need someone to write the release notes.
Do you have time to do so?

AlexHill:
@thedrow, I've opened a PR to this repo with release notes for 4.2.0.

PR #827, opened 2018-01-26
Removed an always true if-check in t/integration/transport.py

atombrella:


thedrow:
Thanks! Feel free to add yourself to the AUTHORS file.

PR #828, opened 2018-01-26
Replaced a dict-construct with a dictionary comprehension.

atombrella:
It should be a bit faster to use dict comprehensions.


thedrow:
Thanks. Feel free to add yourself to the AUTHORS file.

Issue #761, opened 2017-07-13
Permission problem with SQS


Fiedzia:
AWS allows to give permission to read from specific queue, but not to list available queues.
In that case attempt to use kombu fails, because it is trying to list all queues and fails with permission error.
There is no workaround for that (other than monkeypatching).

auvipy:
whats your suggested workaround for that?

georgepsarakis:
@auvipy @Fiedzia the queue listing based on *prefix* happens [here](https://github.com/celery/kombu/blob/master/kombu/transport/SQS.py#L107).

As an alternative the [get_queue_url](http://boto3.readthedocs.io/en/latest/reference/services/sqs.html#SQS.Client.get_queue_url) API call could be used, but the queue prefix support must be dropped (if it was ever intentionally supported).

Fiedzia:
> whats your suggested workaround for that?

Main goal for me was to use celery, not Kombu directly, so I'd prefer solution that allows that.
If the current behavious is desired , one way would be to make it controllable via transport options,
another would be to have  single-queue variant of a transport.

However I don't really understand the problem that having  queue cache is trying to solve (and I don't believe it even solves it, taking into account that you may have more than one application using sqs).
The comment in code states that:

> # SQS blows up if you try to create a new queue when one already
> # exists but with a different visibility_timeout.  This prepopulates
> # the queue_cache to protect us from recreating
> # queues that are known to already exist.

couldn't this condition be simply handled by raising an exception when it will happen
and let developer deal with that? I am not sure what "blows up" mean, but does it happen in a way that allows to know the reason for it?

georgepsarakis:
From the AWS SQS [API reference](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html):

> QueueAlreadyExists
A queue already exists with this name. Amazon SQS returns this error only if the request includes attributes whose values differ from those of the existing queue.
HTTP Status Code: 400

georgepsarakis:
@Fiedzia your thought on letting the developer handle the error by raising a proper exception is correct, however what happens when the transport is initialized from a Celery worker? Perhaps the queue listing should be somehow converted to a transport option as you initially suggested.

georgepsarakis:
@eruvanos #834 has been opened that addresses this issue. Can you please have a look at the solution?

auvipy:
closing as per https://github.com/celery/kombu/pull/1156

PR #834, opened 2018-02-09
Add ability to use a pre-existing SQS queue, without requiring sqs:ListQueues permission

aeijdenberg:
In our environment one team who manages AWS resources makes available certain data sources (such as RDS instances, and SQS queues) available to other teams.

In this case, a team has been given full access to a specific SQS queue, but does not have permission to ListQueues (which allows listing of all queues in an account).

This patch provides 2 additional transport options to be specified:

`never_list_queues` (defaults to `False`) that if set, won't attempt to list queues.

`predefined_queue_urls` (defaults to `None`) that if set, defines a map of already existing queue names, to URLs.

When applied, we can use as follows:

```python
import celery
import urllib

c = celery.Celery(
    queue_name,
    broker='sqs://{}:{}@'.format(urllib.quote_plus(key), urllib.quote_plus(secret)),
    config_source={
        'broker_transport_options': {
            'region': region,
            'never_list_queues': True,
            'predefined_queue_urls': {
                queue_name: 'https://%s.queue.amazonaws.com/%s/%s' % (region, queue_account, queue_name),
            },
        },
        'task_default_queue': queue_name,
    },
)
```

Thank you for your consideration.

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -78,6 +78,14 @@ def maybe_int(x):
         return x
 
 
+class UnknownQueueException(Exception):
+    """We are using predefined queues and this one is not defined."""
+
+
+class PredefinedQueueException(Exception):
+    """We are using predefined queues, this one should not be deleted."""
+
+
 class Channel(virtual.Channel):
     """SQS Channel."""
 
@@ -104,6 +112,9 @@ class Channel(virtual.Channel):
         self.hub = kwargs.get('hub') or get_event_loop()
 
     def _update_queue_cache(self, queue_name_prefix):
+        if self.predefined_queue_urls:
+            self._queue_cache.update(self.predefined_queue_urls)
+            return
         resp = self.sqs.list_queues(QueueNamePrefix=queue_name_prefix)
         for url in resp.get('QueueUrls', []):
             queue_name = url.split('/')[-1]
@@ -170,12 +181,21 @@ class Channel(virtual.Channel):
         # _queue_cache population.
         queue = self.canonical_queue_name(queue)
 
+        # If we have the URL cached, return it
+        rv = self._queue_cache.get(queue, None)
+        if rv:
+            return rv
+
+        # If we don't have it cached, and are using predefined queue URLs,
+        # raise an error, don't try to create it.
+        if self.predefined_queue_urls:
+            raise UnknownQueueException
+
         # The SQS ListQueues method only returns 1000 queues.  When you have
         # so many queues, it's possible that the queue you are looking for is
         # not cached.  In this case, we could update the cache with the exact
         # queue name first.
-        if queue not in self._queue_cache:
-            self._update_queue_cache(queue)
+        self._update_queue_cache(queue)
         try:
             return self._queue_cache[queue]
         except KeyError:
@@ -190,6 +210,14 @@ class Channel(virtual.Channel):
 
     def _delete(self, queue, *args, **kwargs):
         """Delete queue by name."""
+        # If we are using predefined queue URLs,
+        # raise an error, don't try to delete it.
+        # At this time, the following doesn't actually
+        # try to delete it anwyway, but we won't rely on
+        # that behavior.
+        if self.predefined_queue_urls:
+            raise PredefinedQueueException
+
         super(Channel, self)._delete(queue)
         self._queue_cache.pop(queue, None)
 
@@ -449,6 +477,22 @@ class Channel(virtual.Channel):
     def visibility_timeout(self):
         return (self.transport_options.get('visibility_timeout') or
                 self.default_visibility_timeout)
+
+    @cached_property
+    def predefined_queue_urls(self):
+        """Map of queue_name to queue URL.
+
+        For example:
+        {"my-queue":
+         "https://ap-southeast-2.queue.amazonaws.com/1234567890/my-queue"}
+
+        This is useful when you have been supplied an already existing SQS
+        resource, but not permissions to create, delete, or list your own.
+
+        When set, this libary will not attempt to list SQS queues, nor will
+        it try to create or delete SQS queues.
+        """
+        return self.transport_options.get('predefined_queue_urls', None)
 
     @cached_property
     def queue_name_prefix(self):


georgepsarakis:
@aeijdenberg this is a valid concern, and has been raised in the past. Thanks for submitting the patch I will have a look shortly.

kombu/transport/SQS.py
```diff
@@ -450,6 +454,29 @@ def visibility_timeout(self):
         return (self.transport_options.get('visibility_timeout') or
                 self.default_visibility_timeout)
 
+    @cached_property
+    def predefined_queue_urls(self):
```
georgepsarakis: It is useful to add this to the transport documentation as well. See [here](http://docs.celeryproject.org/en/latest/getting-started/brokers/sqs.html).

kombu/transport/SQS.py
```diff
@@ -450,6 +454,29 @@ def visibility_timeout(self):
...
+
+        This is useful when you have been supplied an already existing SQS resource, but not
+        permissions to create your own, or list your own.
+
+        Often used with never_list_queues.
+        """
+        return self.transport_options.get('predefined_queue_urls', None)
+
+    @cached_property
+    def never_list_queues(self):
```
georgepsarakis: I don't think that you need this setting. I believe it is preferrable to just skip the queue URL listing request, when the queue URL mapping has been given with `predefined_queue_urls`. Seems clearer from a user's point of view.

kombu/transport/SQS.py
```diff
@@ -104,6 +104,10 @@ def __init__(self, *args, **kwargs):
         self.hub = kwargs.get('hub') or get_event_loop()
 
     def _update_queue_cache(self, queue_name_prefix):
+        if self.predefined_queue_urls:
```
georgepsarakis: You will possibly need one small test to verify the change in the behavior of `_update_queue_cache` depending on the transport settings. Let me know if you need help with that!

aeijdenberg:
@georgepsarakis - thanks for the comments. I'll take a look on Monday (just finishing up for Friday here in Australia).

kombu/transport/SQS.py
aeijdenberg: Created https://github.com/celery/celery/pull/4534 for documentation change.

kombu/transport/SQS.py
aeijdenberg: I probably will need some help with that - I haven't contributed to Python projects in a while, so not up to speed on the latest testing...

georgepsarakis:
@aeijdenberg a separate Pull Request is not necessary for documentation changes. Actually I think it is better to be placed here, along with the accompanying source code changes. If you can, just transfer these commits here too.

kombu/transport/SQS.py
georgepsarakis: The unit tests for the SQS transport can be found [here](https://github.com/celery/kombu/blob/master/t/unit/transport/test_SQS.py). I will try to construct a small test case.

aeijdenberg:
@georgepsarakis - I'm not sure exactly what you mean by transferring the commit for the doc change here.  The reason I opened a separate PR is that the documentation, as far as I can tell, is in a separate GitHub repo: https://github.com/celery/celery/pull/4534

thedrow:
Kombu and Celery have separate documentation. We need to add it to this repository as well.

kombu/transport/SQS.py
```diff
@@ -450,6 +478,22 @@ def visibility_timeout(self):
...
                 self.default_visibility_timeout)
 
+    @cached_property
+    def predefined_queue_urls(self):
+        """Map of queue_name to queue URL.
+
+        For example:
+        {"my-queue":
+         "https://ap-southeast-2.queue.amazonaws.com/1234567890/my-queue"}
+
```
georgepsarakis: I think that the comments from here onwards, however useful, belong to the documentation and not the docstring.

kombu/transport/SQS.py
```diff
@@ -190,6 +210,14 @@ def _new_queue(self, queue, **kwargs):
 
     def _delete(self, queue, *args, **kwargs):
         """Delete queue by name."""
+        # If we are using predefined queue URLs,
```
georgepsarakis: I don't think that an exception needs to be raised here, because it may be backwards incompatible. An early return would probably suffice.

kombu/transport/SQS.py
```diff
@@ -170,12 +181,21 @@ def _new_queue(self, queue, **kwargs):
...
         # The SQS ListQueues method only returns 1000 queues.  When you have
         # so many queues, it's possible that the queue you are looking for is
         # not cached.  In this case, we could update the cache with the exact
         # queue name first.
-        if queue not in self._queue_cache:
-            self._update_queue_cache(queue)
+        self._update_queue_cache(queue)
         try:
             return self._queue_cache[queue]
         except KeyError:
```
georgepsarakis: I would suggest to replace the modifications in this function with a simple additional condition in the `except` block:

```
if self.predefined_queue_urls:
    raise UnknownQueueException("Queue '{}' has not been declared in Predefined Queue URL mapping.".format(queue))
```

georgepsarakis:
@aeijdenberg sorry for taking long time to reply again. I will try and look into the test case.

thedrow:
@aeijdenberg Ping?
Are you going to continue to work on this PR?

thedrow:
@georgepsarakis Can we merge this? What's missing?

georgepsarakis:
@thedrow I think this is a very useful feature. It will also possibly resolve #761, by allowing users to define the queues, thus not requiring additional permissions for listing them.

Before merging I would suggest that we work on the following:
1. Address open items from the latest [review](https://github.com/celery/kombu/pull/834#pullrequestreview-116600149)
2. Document the options of the feature
3. Add unit tests

@aeijdenberg thanks for the contribution :+1: !

ayushthedev:
Any Updates on this

auvipy:
probably in next version of kombu... 4.7.x

ayushthedev:
When will it be out? Will that be supported for python 2.7 as well , Now that itll be going obsolete

auvipy:
celery and kombu 4.x.x versions will be 2.7 forever. and will get bug fix until December 2020. this is because of celery & kombu 5.x.x will adapt lots of modern python 3.6+ features & newer tools & architecture. This will take longer to be stable in production. from the milestone, you can see the celery 5.0 is planned to be a stable release by June 2020. Definitely, if we can make it earlier we will try. for kombu, 4.7.x expect 2-3 months more at least. after celery 4.4.0 stable version is released by August 30.

auvipy:
celery and kombu 4.x.x versions will be 2.7 forever. and will get bug fix until December 2020. this is because of celery & kombu 5.x.x will adapt lots of modern python 3.6+ features & newer tools & architecture. This will take longer to be stable in production. from the milestone, you can see the celery 5.0 is planned to be a stable release by June 2020. Definitely, if we can make it earlier we will try. for kombu, 4.7.x expect 2-3 months more at least. after celery 4.4.0 stable version is released by August 30.

PR #850, opened 2018-04-20
Index in sqlalchemy transport for quick sorting

mu5h3r:


kombu/transport/sqlalchemy/models.py
```diff
@@ -37,7 +37,10 @@ def messages(cls):
 class Message(object):
     """The message class."""
 
-    __table_args__ = {'sqlite_autoincrement': True, 'mysql_engine': 'InnoDB'}
+    __table_args__ = (
+        Index('ix_kombu_message_timestamp_id', 'timestamp', 'id'),
```
georgepsarakis: Can you please explain why you chose the two columns together? If it is only for sorting, I guess you only need the timestamp, right?

mu5h3r: This may be actual only in my case but postgresql planner do this for ~6000000 messages:
with index
default
without index
default

georgepsarakis: @mu5h3r thanks for the feedback, yes this makes sense.

thedrow:
@georgepsarakis Is this a breaking change?
Do we provide migrations?

georgepsarakis:
@thedrow I don't believe this would break anything, the only issue is that users must recreate the table to take advantage of the index (unrelated to this PR of course).

As far as I can see, unfortunately no migrations exist. Just as a note, I think for SQLAlchemy the proper migration tool is [alembic](https://bitbucket.org/zzzeek/alembic).

PR #859, opened 2018-05-01
Fix mongodb+srv:// uri parsing, and add the schema to the transport m…

### LINKED ISSUE #858: Kombu fails with tracebacks when you try to connect with mongodb+srv uris
Similar to https://github.com/celery/celery/issues/4668, kombu falls over when you try to pass in a mongodb+srv uri as described here: https://docs.mongodb.com/manual/reference/connection-string/#dns-seedlist-connection-format

I'll provide a Pull Request.

josephtate:
…atrix. Add tests.

Fixes #858

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -169,7 +169,8 @@ class Connection(object):
             if ';' in hostname:
                 alt.extend(hostname.split(';'))
                 hostname = alt[0]
-            if '+' in hostname[:hostname.index('://')]:
+            scheme = hostname[:hostname.index('://')]
+            if 'mongodb+srv' not in scheme and '+' in scheme:
                 # e.g. sqla+mysql://root:masterkey@localhost/
                 params['transport'], params['hostname'] = \
                     hostname.split('+', 1)
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -170,11 +170,23 @@ class Connection(object):
                 alt.extend(hostname.split(';'))
                 hostname = alt[0]
             scheme = hostname[:hostname.index('://')]
-            if 'mongodb+srv' not in scheme and '+' in scheme:
-                # e.g. sqla+mysql://root:masterkey@localhost/
-                params['transport'], params['hostname'] = \
-                    hostname.split('+', 1)
-                self.uri_prefix = params['transport']
+            if '+' in scheme:
+                tran, _, host = hostname.partition('+')
+                # Override the parsed transport with any provided
+                if transport and not isinstance(transport, string_t):
+                    params['transport'] = transport
+                    cls = transport
+                else:
+                    params['transport'] = tran
+                    cls = get_transport_cls(tran)
+                if getattr(cls, 'expects_url_munging', True):
+                    # e.g. sqla+mysql://root:masterkey@localhost/
+                    params['hostname'] = host
+                    self.uri_prefix = tran
+                else:
+                    # e.g., mongodb+srv://user:pass@host/db?params
+                    # Use the original value
+                    params['hostname'] = hostname
             else:
                 transport = transport or urlparse(hostname).scheme
                 if not get_transport_cls(transport).can_parse_url:
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -169,8 +169,7 @@ class Connection(object):
             if ';' in hostname:
                 alt.extend(hostname.split(';'))
                 hostname = alt[0]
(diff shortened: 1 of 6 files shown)


josephtate:
Please don't merge this yet. It's also broken in get_backend_by_url. I'll have a fix shortly.

kombu/connection.py
```diff
@@ -169,7 +169,8 @@ def __init__(self, hostname='localhost', userid=None,
             if ';' in hostname:
                 alt.extend(hostname.split(';'))
                 hostname = alt[0]
-            if '+' in hostname[:hostname.index('://')]:
+            scheme = hostname[:hostname.index('://')]
+            if 'mongodb+srv' not in scheme and '+' in scheme:
```
amol-: wonder if instead adding `mongodb+srv` explicitly here, it wouldn't make sense to check if scheme is in `TRANSPORT_ALIASES` (or better if you can get a transport for it) before splitting by `+`.

If you can't get a transport for the whole scheme, then you split by `+` and check if you can get a transport for the part before `+`.

That would allow to support any transport that has `+` in the scheme without having to explicitly list them here.

kombu/transport/mongodb.py
```diff
@@ -243,13 +243,15 @@ def _parse_uri(self, scheme='mongodb://'):
         client = self.connection.client
         hostname = client.hostname
 
-        if not hostname.startswith(scheme):
+        if '://' not in hostname:
             hostname = scheme + hostname
+        urischeme, rest = hostname.split('://', 1)
```
amol-: `scheme` and `urischeme` seem pretty confusing. As it's not really used maybe name it `_` or just use `scheme` itself as it's the scheme?

t/unit/test_connection.py
```diff
@@ -54,6 +54,11 @@ def test_as_uri_when_mongodb(self):
         x = Connection('mongodb://localhost')
         assert x.as_uri()
 
+    @skip.unless_module('pymongo')
+    def test_as_uri_when_mongodb_srv(self):
+        x = Connection('mongodb+srv://localhost')
+        assert x.as_uri()
```
bertjwregeer: Shouldn't this assert that the uri is `mongodb+srv://localhost`? Just asserting it doesn't return `Falsey` doesn't seem like a great test.

But I see that the other mongodb test does the same thing...

bertjwregeer:
:+1:

This looks great for supporting the issue I brought up in https://github.com/celery/celery/issues/4668

kombu/connection.py
```diff
@@ -170,10 +170,22 @@ def __init__(self, hostname='localhost', userid=None,
...
             if '+' in hostname[:hostname.index('://')]:
-                # e.g. sqla+mysql://root:masterkey@localhost/
-                params['transport'], params['hostname'] = \
-                    hostname.split('+', 1)
-                self.uri_prefix = params['transport']
+                tran, _, host = hostname.partition('+')
+                # Override the parsed transport with any provided
+                if transport and not isinstance(transport, string_t):
+                    params['transport'] = transport
+                    cls = transport
```
amol-: Uhm, other parts of the code (IE https://github.com/celery/kombu/pull/859/files#diff-8617041076de8665cfa306742f9b9163L235 ) seem to take for granted that `params['transport']` is a string.

Here it gets set to the provided class, which might have side effects to other parts of the code.

As the transport class doesn't seem to know its name in any way and looking it up in `TRANSPORT_ALIASES` would be fragile for custom transports I'm not sure supporting non-string transports is expected. 
The other branches of the if seem to take for grnated that `transport` is a string and call `get_transport_cls` on it to grab the actual class

josephtate: Running the tests proves otherwise. The test suite (especially the mongo tests) use a passed in transport class to mock out the connection object. NOT using it causes failures when the Connection tries to call out to a real mongodb.

In the original patch set, we bypassed this block if it was mongo, so it went down the other branch (lines 189+). We need a way to use a provided Transport as an override whether or not there's a plus sign. There aren't many transports that support the + schema extension system, so I think we're ok.

thedrow:
@georgepsarakis Should we close this PR?

bertjwregeer:
What is left to be done? If we are going to use @georgepsarakis approach then this PR can get closed, but how can we move this forward?

thedrow:
Since Celery isn't using MongoDB as a transport and this PR requires some work I'm going to close this PR.

We should start a new one if the MongoDB broker ever get's used.

Issue #863, opened 2018-05-08
Queues that are exclusive should not imply they also auto-delete


danpoland:
Queues that are exclusive should not imply they are also auto-deleted: https://github.com/celery/kombu/blob/master/kombu/entity.py#L582

The AMQP specification does not make the assumption that exclusive queues are also auto-deleted.

Section 2.1.4.1
> exclusive - if set, the queue belongs to the current connection only, and is deleted when the connection closes.

Section 3.1.4
> Message queues may be durable, temporary, or auto-deleted. Durable message queues last until they are deleted. Temporary message queues last until the server shuts-down. Auto-deleted message queues last until they are no longer used.

Per the AMQP specification a queue's exclusiveness only pertains to the connection. Auto-deleted deals with consumers and deletes a queue when the queue has had at least one consumer and the last consumer unsubscribes.

Also see RabbitMQ's AMQP explanation of queues: https://www.rabbitmq.com/tutorials/amqp-concepts.html. Note that it is possible to create exclusive non-auto-deleted queues in RabbitMQ.

danpoland:
PR incoming.

PR #864, opened 2018-05-08
Exclusive autodelete

danpoland:
Fixes https://github.com/celery/kombu/issues/863


danpoland:
I am not sure I know what you mean. 

If people coded against the current incorrect assumption that exclusive also meant auto-deleted, they could  run into unexpected results. However, that assumption is incorrect and out of AMQP spec and should be corrected.

clokep:
It isn't an assumption when using the kombu library though, it was a documented behavior. This seems like something that would need to be documented as changing and have a deprecation warning thrown before being changed in a major version update.

PR #865, opened 2018-05-09
"Reentrant" version, that allows a `celery.bin.celery.main` worker to run multiple times

alanjds:
The actual version does not allow `celery.bin.celery.main` to run, raise a SystemExit, catch it and then run again.

To allow it, the `kombu.async.hub.Hub.poller` became a @property that regenerates on access. And the Redis backend tries to recreate it before acessing, if needed.

Indeed, `celery.worker.state.should_{stop,terminate}` should be reset after every SystemExit for it to work, but this part is a patch not for Kombu repo ;) .

### DIFF
--- a/kombu/async/hub.py
+++ b/kombu/async/hub.py
@@ -99,19 +99,25 @@ class Hub(object):
 
         self._create_poller()
 
+    @property
+    def poller(self):
+        if not self._poller:
+            self._create_poller()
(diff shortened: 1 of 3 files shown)


alanjds:
Closing as it got targeted on `master`, but no `v410` target exists. Will reopen on the right base.

PR #866, opened 2018-05-09
"Reentrant" version, that allows a `celery.bin.celery.main` worker to run multiple times

alanjds:
The actual version does not allow `celery.bin.celery.main` to run, raise a SystemExit, catch it and then run again. (See: https://github.com/alanjds/celery-serverless/commit/a23f79d3cfe17ed8a91003707af4a5dd1e2aa542#diff-b4826aeb276ca699cf1adda1903fe3eaR55)

To allow it, the `kombu.async.hub.Hub.poller` became a @property that regenerates on access. And the Redis backend tries to recreate it before acessing, if needed.

Indeed, `celery.worker.state.should_{stop,terminate}` should be reset after every SystemExit for it to work, but this part is a patch not for Kombu repo ;) .
(See: https://github.com/alanjds/celery-serverless/commit/a23f79d3cfe17ed8a91003707af4a5dd1e2aa542#diff-b4826aeb276ca699cf1adda1903fe3eaR57)

### DIFF
--- a/kombu/asynchronous/hub.py
+++ b/kombu/asynchronous/hub.py
@@ -99,19 +99,25 @@ class Hub(object):
 
         self._create_poller()
 
+    @property
+    def poller(self):
+        if not self._poller:
+            self._create_poller()
+        return self._poller
+
     def reset(self):
         self.close()
         self._create_poller()
 
     def _create_poller(self):
-        self.poller = poll()
-        self._register_fd = self.poller.register
-        self._unregister_fd = self.poller.unregister
+        self._poller = poll()
+        self._register_fd = self._poller.register
+        self._unregister_fd = self._poller.unregister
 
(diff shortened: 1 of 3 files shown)


alanjds:
To raise the coverage, I can think of a case closing the Redis pooler an then trying to reuse it again. Will take a look on what file it would fit...

alanjds:
@auvipy I am trying to bake a test case for "Redis got closed then reused again", but could not find a way: The actual tests never set `MultiChannelPoller.poller = None`, the precondition for the error to occur.

I could use some help here, if possible.

alanjds:
Looks like Travis failed only on Python 3.4, for some unknown reason. Can you please issue a retry there please?

georgepsarakis:
@alanjds failed builds restarted.

PR #878, opened 2018-05-25
Handle librabbitmq fileno socket not connected raising ValueError

### LINKED ISSUE #860: ValueError: Socket not connected
Hello,
the following error happens sometimes when publishing :

```
  File "/foo/bar/lib/python2.7/site-packages/kombu/messaging.py", line 181, in publish
    exchange_name, declare,
  File "/foo/bar/lib/python2.7/site-packages/kombu/connection.py", line 506, in _ensured
    self.collect()
  File "/foo/bar/lib/python2.7/site-packages/kombu/connection.py", line 350, in collect
    gc_transport(self._connection)
  File "/foo/bar/lib/python2.7/site-packages/kombu/transport/librabbitmq.py", line 148, in _collect
    os.close(connection.fileno())
ValueError: Socket not connected

```

kombu==4.1.0

bshelton229:
Hi, I believe this will solve #860. Because librabbitmq connection's fileno method will raise `ValueError` (https://github.com/celery/librabbitmq/issues/39) if the socket has unexpected been closed, this was escaping being caught when running `os.close`, preventing publish connection retries if the socket disappeared.

Closes #860.

### DIFF
--- a/t/unit/transport/test_librabbitmq.py
+++ b/t/unit/transport/test_librabbitmq.py
@@ -125,6 +125,15 @@ class test_Transport(lrmqCase):
             self.T._collect(conn)
             close.assert_called_with(conn.fileno())
 
+    def test_collect__with_fileno_raising_value_error(self):
+        conn = Mock(name='connection')
+        conn.channels = {1: Mock(name='chan1'), 2: Mock(name='chan2')}
+        with patch('os.close') as close:
+            self.T.client = self.client
+            conn.fileno.side_effect = ValueError("Socket not connected")
+            self.T._collect(conn)
+            close.assert_not_called()
+
     def test_register_with_event_loop(self):
         conn = Mock(name='conn')
         loop = Mock(name='loop')
--- a/t/unit/transport/test_librabbitmq.py
+++ b/t/unit/transport/test_librabbitmq.py
@@ -133,6 +133,7 @@ class test_Transport(lrmqCase):
             conn.fileno.side_effect = ValueError("Socket not connected")
             self.T._collect(conn)
             close.assert_not_called()
+        conn.fileno.assert_called_with()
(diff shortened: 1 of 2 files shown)


bshelton229:
@auvipy I was curious about the coverage issue, I'll look into it. Is it expecting coverage to be a positive change in order to pass? I wasn't totally clear on what would cause that check to pass.

t/unit/transport/test_librabbitmq.py
```diff
@@ -125,6 +125,17 @@ def test_collect__with_conn(self):
...
             close.assert_called_with(conn.fileno())
 
+    def test_collect__with_fileno_raising_value_error(self):
+        conn = Mock(name='connection')
+        conn.channels = {1: Mock(name='chan1'), 2: Mock(name='chan2')}
+        with patch('os.close') as close:
+            self.T.client = self.client
+            conn.fileno.side_effect = ValueError("Socket not connected")
+            self.T._collect(conn)
+            close.assert_not_called()
```
georgepsarakis: Can you please also add an assertion that `conn.fileno` has actually been called once?

I believe that your patch and test case is 100% correct, but perhaps there is an issue with the test that also causes Code Coverage to be not reported as expected. What do you think?

bshelton229: Thanks @georgepsarakis I gave asserting fileno was called a try and that didn't seem to work, but was definitely worth a shot.

georgepsarakis: @bshelton229 thanks for trying this, I believe this is ready for merge.

georgepsarakis:
Feel free to add yourself to the list of contributors: https://github.com/celery/kombu/blob/master/AUTHORS

PR #884, opened 2018-06-04
don't cache declaration of expiring queues

mattbennett:
Queues that are declared with a TTL should also be excluded from the cache in case they expire between publishes on the same channel.

### DIFF
--- a/kombu/entity.py
+++ b/kombu/entity.py
@@ -804,7 +804,11 @@ class Queue(MaybeChannelBound):
 
     @property
     def can_cache_declaration(self):
-        return not self.auto_delete
+        if self.queue_arguments:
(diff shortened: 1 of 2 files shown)


thedrow:
Looks good but I'm wondering what exactly are you trying to fix here?
Why hasn't anyone noticed this problem before?

mattbennett:
The existing `can_cache_declaration` implementation forces channels to always redeclare queues that are declared as "auto-delete", presumably because the queue may be subsequently removed during the lifetime of the  channel. In other words, you can only safely omit the declaration if you _know_ the queue already exists, which is not the case for queues that are "auto-deleted". The same is true for queues declared with a TTL.

PR #887, opened 2018-06-12
Fixed SQS queue name parsing for queue names ending with an "f"

Sp1tF1r3:
Due to improper use of 'rstrip' queue names ending with an 'f' will not be parsed properly

```
>>> queue_name_ending_with_an_f = 'queue_name_ending_with_an_f.fifo'
>>> queue_name_ending_with_an_f.rstrip('.fifo')
'queue_name_ending_with_an_'
>>> queue_name_ending_with_an_f.split('.')[:1][0]
'queue_name_ending_with_an_f'
>>>
```

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -153,7 +153,9 @@ class Channel(virtual.Channel):
     def entity_name(self, name, table=CHARS_REPLACE_TABLE):
         """Format AMQP queue name into a legal SQS queue name."""
         if name.endswith('.fifo'):
-            partial = name.rstrip('.fifo')
+            # SQS allows no dots in queue names with one exception being FIFO queues
+            # For FIFO queues only one dot is allowed in '.fifo' suffix
+            partial = name.split('.')[:1][0]
             partial = text_t(safe_str(partial)).translate(table)
             return partial + '.fifo'
         else:


Sp1tF1r3:
will do

kombu/transport/SQS.py
```diff
@@ -153,7 +153,9 @@ def _reset_cycle(self):
     def entity_name(self, name, table=CHARS_REPLACE_TABLE):
         """Format AMQP queue name into a legal SQS queue name."""
         if name.endswith('.fifo'):
-            partial = name.rstrip('.fifo')
+            # SQS allows no dots in queue names except FIFO queues.
+            # For FIFO queues only one dot is allowed in '.fifo' suffix
+            partial = name.split('.')[:1][0]
```
georgepsarakis: Nice finding! A suggestion here, I think this is equivalent and perhaps simpler:

```python
partial = name.split('.')[0]
```

Sp1tF1r3: makes sense, but I initially thought SQS may have multiple '.' in queue name..

Sp1tF1r3: you are correct... I over complicated it for no reason

georgepsarakis: Is the dot character valid in queue names? If that is the case and perhaps accounts better for all possible cases:

```python
>>> 'a.b.c.fifo'.rsplit('.', 1)
['a.b.c', 'fifo']
```

kombu/transport/SQS.py
```diff
@@ -153,7 +153,9 @@ def _reset_cycle(self):
     def entity_name(self, name, table=CHARS_REPLACE_TABLE):
         """Format AMQP queue name into a legal SQS queue name."""
         if name.endswith('.fifo'):
-            partial = name.rstrip('.fifo')
+            # SQS allows no dots in queue names except FIFO queues.
+            # For FIFO queues only one dot is allowed in '.fifo' suffix
+            partial = name.split('.')[0]
```
jainraj: Why can't we use 
`partial = name.rsplit('.fifo')[0]`? 
Splitting on dot seems a bit scary. 
Or even `partial = name.split('.fifo')[0]`

Please note **split** instead of **strip**

jainraj: And moreover note that they are handling the **no dots in queue names** by the `CHARS_REPLACE_TABLE` setting. This would just violate DRY principle and bring in scary changes.

ikonst: Why can't you simply do:
```py
partial = name[:-(len('.fifo')]
```

jainraj: Yeah. This is the better method. Slicing is better than splitting. 👍

jainraj: according to https://doeidoei.wordpress.com/2011/10/18/python-slicing-vs-splitting-vs-endswith/

PR #889, opened 2018-06-13
fix librabbitmq link error

faycheng:
we couldn't find `https://pypi.python.org/librabbitmq/`  on pypi, and i modify it to `https://pypi.python.org/pypi/librabbitmq/`


thedrow:
Thanks!

PR #891, opened 2018-06-18
Add transports based on Azure PaaS

c-w:
This pull request adds two new transport implementations:

- `azurestoragequeues` is implemented on top of [Azure Storage Queues](https://azure.microsoft.com/en-us/services/storage/queues/). This offers a simple but scalable and low-cost PaaS transport for Celery users in Azure. The transport is intended to be used in conjunction with the [Azure Block Blob Storage backend](https://github.com/celery/celery/pull/4685).

- `azureservicebus` is implemented on top of [Azure Service Bus](https://azure.microsoft.com/en-us/services/service-bus/) and offers PaaS support for more demanding Celery workloads in Azure. The transport is intended to be used in conjunction with the [Azure CosmosDB backend](https://github.com/celery/celery/pull/4720).

This pull request was created together with @ankurokok, @dkisselev, @evandropaula, @martinpeck and @michaelperel.

### DIFF
--- a/kombu/transport/azureservicebus.py
+++ b/kombu/transport/azureservicebus.py
@@ -19,7 +19,6 @@ https://azure.microsoft.com/en-us/services/service-bus/
 """
 from __future__ import absolute_import, unicode_literals
 
-from os import getenv
 import string
 
 from kombu.five import Empty, text_t
@@ -118,12 +117,9 @@ class Channel(virtual.Channel):
     def sbq(self):
         if self._sbq is None:
             self._sbq = ServiceBusService(
-                service_namespace=getenv('AZURE_SERVICEBUS_ACCOUNT',
-                                         self.conninfo.hostname),
-                shared_access_key_name=getenv('AZURE_SERVICEBUS_KEY_NAME',
-                                              self.conninfo.userid),
-                shared_access_key_value=getenv('AZURE_SERVICEBUS_KEY_VALUE',
-                                               self.conninfo.password))
+                service_namespace=self.conninfo.hostname,
+                shared_access_key_name=self.conninfo.userid,
+                shared_access_key_value=self.conninfo.password)
 
         return self._sbq
 
--- a/kombu/transport/azureservicebus.py
+++ b/kombu/transport/azureservicebus.py
@@ -44,7 +44,7 @@ class Channel(virtual.Channel):
 
     default_visibility_timeout = 1800  # 30 minutes.
     domain_format = 'kombu%(vhost)s'
-    _sbq = None
+    _queue_service = None
     _queue_cache = {}
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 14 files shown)


auvipy:
@c-w is there anything remaining from your side?

c-w:
Hi @auvipy. Thanks for your review! If you're happy to merge this, I'm all for it. (I don't have write access to the repository so one of your maintainers will have to hit the merge button.)

kombu/transport/azureservicebus.py
```diff
@@ -0,0 +1,154 @@
...
+            else:
+                n += 1
+
+        return n
+
+    @property
+    def sbq(self):
+        if self._sbq is None:
+            self._sbq = ServiceBusService(
+                service_namespace=getenv('AZURE_SERVICEBUS_ACCOUNT',
```
georgepsarakis: I am not sure environment variables should work at this level and I don't see any other transport reading from the environment. The configuration should have been done at a higher level (in this case Celery).

kombu/transport/azureservicebus.py
```diff
@@ -0,0 +1,154 @@
...
+    _queue_cache = {}
+
+    def __init__(self, *args, **kwargs):
+        if ServiceBusService is None:
+            raise ImportError('Azure Service Bus transport requires the '
+                              'azure-servicebus library')
+
+        super(Channel, self).__init__(*args, **kwargs)
+
+        for queue in self.sbq.list_queues():
```
georgepsarakis: There is an [open issue](https://github.com/celery/kombu/pull/834#issue-168161180) with AWS SQS transport, not requiring `ListQueues` permissions. In the proposed fix, a predefined list of queue names is given by configuration. Would that apply here as well?

kombu/transport/azureservicebus.py
```diff
@@ -0,0 +1,154 @@
...
+
+            if not message.body:
+                break
+            else:
+                n += 1
+
+        return n
+
+    @property
+    def sbq(self):
```
georgepsarakis: Not sure, but perhaps the abbreviation will be difficult to read for readers unfamiliar with the service?

kombu/transport/azureservicebus.py
c-w: Done in f260341.

kombu/transport/azureservicebus.py
c-w: Done in a7df447.

kombu/transport/azureservicebus.py
c-w: Azure ServiceBus has three kinds of permissions: `listen`, `send`, and `manage` ([docs](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-sas#shared-access-authorization-policies)). Given that we're creating new queues from the connector, we'll anyways need the `manage` permissions so might as well use the list queues call here. Thoughts?

c-w:
@georgepsarakis Thanks for the comments. Addressed them. Let me know if there's anything else.

auvipy:
please rebase

c-w:
Done, @auvipy.

PR #895, opened 2018-07-09
Making the pyro transport available, fix it for recent Pyro4 versions, add broker daemon

irmen:
I'm trying to get the Pyro transport working for Kombu.

A few minor changes were needed in the transport itself to be compatible with recent Pyro versions again, but I struggle with understanding the following:

Where is the pyro broker component that should  provide the actual queueing methods that are remotely accessed from the Pyro transport? I see nothing in the kombu repo doing this, and can't find anything via google.  Is there such a thing or should I write it myself?

### DIFF
--- a/kombu/transport/__init__.py
+++ b/kombu/transport/__init__.py
@@ -39,7 +39,8 @@ TRANSPORT_ALIASES = {
     'consul': 'kombu.transport.consul:Transport',
     'etcd': 'kombu.transport.etcd:Transport',
     'azurestoragequeues': 'kombu.transport.azurestoragequeues:Transport',
-    'azureservicebus': 'kombu.transport.azureservicebus:Transport'
+    'azureservicebus': 'kombu.transport.azureservicebus:Transport',
+    'pyro': 'kombu.transport.pyro:Transport'
 }
 
 _transport_cache = {}
--- a/kombu/transport/pyro.py
+++ b/kombu/transport/pyro.py
@@ -1,30 +1,50 @@
-"""Pyro transport.
+"""Pyro transport, and Kombu Broker daemon.
 
 Requires the :mod:`Pyro4` library to be installed.
+
+To use the Pyro transport with Kombu, use an url of the form:
+``pyro://localhost/kombu.broker``
+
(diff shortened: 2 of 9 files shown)


irmen:
I've updated the PR to include documentation changes , and a working minimal Pyro Broker daemon to actually be able to use the pyro transport.

thedrow:
Hi sorry for the delay in response.
This deserves a full review once I'll find the time to do so.

irmen:
I squashed & rebased my additions.
The travis CI fails with a unit test error but it seems related to Redis.

kombu/transport/__init__.py
```diff
@@ -39,7 +39,8 @@ def supports_librabbitmq():
     'consul': 'kombu.transport.consul:Transport',
     'etcd': 'kombu.transport.etcd:Transport',
     'azurestoragequeues': 'kombu.transport.azurestoragequeues:Transport',
-    'azureservicebus': 'kombu.transport.azureservicebus:Transport'
+    'azureservicebus': 'kombu.transport.azureservicebus:Transport',
```
thedrow: What does azure service bus has to do with this PR?

kombu/transport/pyro.py
```diff
@@ -1,42 +1,64 @@
...
 """
+
 from __future__ import absolute_import, unicode_literals
 
 import sys
 
-from kombu.five import reraise
-from kombu.utils.objects import cached_property
-
+from ..five import reraise, Queue, Empty
```
thedrow: Why the relative imports?

irmen: it's what I'm used to when importing stuff from other submodules from the same library, but I now see that this is not used in kombu's code base. It's trivially changed ofcourse.

auvipy: please revert as this change isn;t necessary

irmen: Actually, relative imports _are_ used in other places as well... for instance in librabbitmq.py, pyamqp.py etc.  Isn't this a bit arbitrary?

irmen:
Thank you for merging this.  
My question related to the Pyro transport and the broker still stands though #897

PR #900, opened 2018-07-25
Allow SimpleQueue to pass 'queue_arguments' to Queue object.

azuer88:
Update simple.py so SimpleQueue can pass queue_arguments to the created Queue object.

This will allow SimpleQueue to connect to RabbitMQ's  lazy queues (afaik), i.e., 'x-queue-mode' is set to 'lazy'.


thedrow:
You can add another test function in this class https://github.com/celery/kombu/blob/master/t/unit/test_simple.py#L101

thedrow:
Thanks!

azuer88:
Thanks for making it easy to contribute.

Issue #903, opened 2018-08-02
Queue constructor doesn't accept str type as Exchange parameter


chibby0ne:
The [Queue documentation](http://docs.celeryproject.org/projects/kombu/en/latest/reference/kombu.html#queue), states that you can pass an `Exchange` or `str` object to the exchange parameter of the constructor:

![kombu_queue_construct_highlight](https://user-images.githubusercontent.com/4380812/43590572-93454a04-9671-11e8-88a9-06ad7d2a35fb.png)

But in reality passing a str results in an exception of the following type:
```
consumer_1  | Traceback (most recent call last):
consumer_1  |   File "consumer.py", line 51, in <module>
consumer_1  |     main()
consumer_1  |   File "consumer.py", line 47, in main
consumer_1  |     Consumer(conn, queues).run()
consumer_1  |   File "/usr/local/lib/python3.7/site-packages/kombu/mixins.py", line 170, in run
consumer_1  |     for _ in self.consume(limit=None, **kwargs):
consumer_1  |   File "/usr/local/lib/python3.7/site-packages/kombu/mixins.py", line 186, in consume
consumer_1  |     with self.consumer_context(**kwargs) as (conn, channel, consumers):
consumer_1  |   File "/usr/local/lib/python3.7/contextlib.py", line 112, in __enter__
consumer_1  |     return next(self.gen)
consumer_1  |   File "/usr/local/lib/python3.7/site-packages/kombu/mixins.py", line 179, in consumer_context
consumer_1  |     with self.Consumer() as (connection, channel, consumers):
consumer_1  |   File "/usr/local/lib/python3.7/contextlib.py", line 112, in __enter__
consumer_1  |     return next(self.gen)
consumer_1  |   File "/usr/local/lib/python3.7/site-packages/kombu/mixins.py", line 228, in Consumer
consumer_1  |     with self._consume_from(*self.get_consumers(cls, channel)) as c:
consumer_1  |   File "consumer.py", line 28, in get_consumers
consumer_1  |     callbacks=[self.on_message])]
consumer_1  |   File "/usr/local/lib/python3.7/site-packages/kombu/messaging.py", line 386, in __init__
consumer_1  |     self.revive(self.channel)
consumer_1  |   File "/usr/local/lib/python3.7/site-packages/kombu/messaging.py", line 404, in revive
consumer_1  |     queue = self._queues[queue.name] = queue(self.channel)
[...]

PR #904, opened 2018-08-10
Queue: add handling of Exchange as str to constructor

chibby0ne:
To fit with the documentation, Queue can now handle a str argument for
the exchange parameter in the constructor, as well as an actual Exchange
object.
Added relevant unit tests to avoid regressions.

Fixes: https://github.com/celery/kombu/issues/903

Signed-off-by: Antonio Gutierrez <chibby0ne@gmail.com>

### DIFF
--- a/kombu/entity.py
+++ b/kombu/entity.py
@@ -568,7 +568,10 @@ class Queue(MaybeChannelBound):
                  **kwargs):
         super(Queue, self).__init__(**kwargs)
         self.name = name or self.name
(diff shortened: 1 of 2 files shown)


chibby0ne:
Why does landscape say there's an error when the score didn't decrease?

chibby0ne:
Might it be because the landscape link is down? https://landscape.io/github/celery/kombu/pulls/904

PR #906, opened 2018-08-15
fix/Broadcast-fail-by-give-queue-kwargs

htkuan:
Because give Broadcast(name='xxx', queue='xxx'),
it will create one specific name for queue,
and if launch more worker for consuming this queue,
this specific name queue already exist then it will not create new queue to
binding fanout exchange.
So all worker will consuming same queue, and exchange only binding this
queue then lose the effect of broadcasting.

### DIFF
--- a/kombu/common.py
+++ b/kombu/common.py
@@ -87,7 +87,7 @@ class Broadcast(Queue):
 
     def __init__(self, name=None, queue=None, auto_delete=True,
                  exchange=None, alias=None, **kwargs):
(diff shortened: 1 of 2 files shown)


htkuan:
test fail on Broadcast('test_Broadcast', 'explicit_queue_name'),
it want 'explicit_queue_name', but it is reason for broadcast fail...

htkuan:
Sorry, I don't understand the "AppVeyor build failed" error.
It seems ci service error?

auvipy:
yes.

Issue #862, opened 2018-05-07
Support rediss:// protocol to connect to redis over TLS


remeika:
Kombu should allow the user to pass a connection URL with [rediss://](https://www.iana.org/assignments/uri-schemes/prov/rediss) protocol  that will create a connection to redis over TLS.

gonzalodiaz:
+1

rnegron:
+1

snstanton:
Yes, this is making it difficult to configure django-redis and celery using the same url scheme.  As it stands, I have to parse the rediss:// url to break it up into bite sized pieces for kombu.

thedrow:
This is already done in master.

PR #910, opened 2018-08-26
Add support for 'rediss' scheme for secure redis connections.

blairboy362:
Putting a 'rediss'-specific check into a generic connection module feels
like a code smell but there doesn't appear to be another way to intercept
the connection initiation code. It defaults to the least secure form, as
there is no suitable default location for `ca_certs`. The recommendation
would still be to follow the documentation and specify `broker_use_ssl` if
coming from celery.

https://github.com/celery/kombu/issues/862

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -10,6 +10,14 @@ from contextlib import contextmanager
 from itertools import count, cycle
 from operator import itemgetter
 
+
+try:
+    from ssl import CERT_NONE
+    ssl_available = True
+except ImportError:  # pragma: no cover
+    CERT_NONE = None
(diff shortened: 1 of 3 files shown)


auvipy:
are you using this in a real environment?

blairboy362:
Not this exact change, but a very similar, more hacky one: https://github.com/celery/kombu/compare/master...alphagov:4.2.1-rediss (currently in a production deployment).
That combined with a correctly configured `broker_use_ssl` in celery is enough to get `rediss` connections to work. All I've done here in addition to that is add in a default for that config so an `SSLConnection` will be established, albeit an insecure one.

Issue #853, opened 2018-04-29
redis: add configuration parameter to prefix all keys used with a custom string


wetneb:
I use kombu (via celery) with a shared redis instance. The administrators of that instance require me to prefix all the keys I use in the redis server with a custom string. While Celery provides ways to prefix queue names, it seems to me that there is currently no way in Kombu (and hence Celery) to prefix all the redis keys used.

My understanding is that the keys used for channels are defined by `keyprefix_queue = '_kombu.binding.%s'`, which not easily modifiable from Celery (via `BROKER_TRANSPORT_OPTIONS`). Would it make sense to add a configuration parameter for that? If that sounds like the correct approach, I can try to provide a PR for that.

georgepsarakis:
@wetneb have you read the Celery documentation section regarding [routing](http://docs.celeryproject.org/en/latest/userguide/routing.html#changing-the-name-of-the-default-queue). I think that if you set the queue name it will also be reflected in the key pattern.

Please let me know if you need additional feedback.

wetneb:
@georgepsarakis indeed it is - but technically it isn't adding a prefix to the keys (as they still start with `_kombu.binding.`), right?

Prefixing redis keys seems to be a fairly widespread practice so I was just thinking that it could make sense for Celery to support that too. But maybe it is just redundant with queue names and you don't want to create confusion with that?

georgepsarakis:
@wetneb I am not sure I understand your point. Isn't `_kombu.binding.<QUEUE_NAME>` the Redis key prefix?

wetneb:
@georgepsarakis yes exactly - what I would like to do is add another custom prefix in front of that. For instance, if both Mary and John use the same redis database, they could want to make sure that all the keys they use are prefixed with different strings. So they would want to use `john._kombu.binding.<QUEUE_NAME>` and `mary._kombu.binding.<QUEUE_NAME>` respectively.

In other words, it would be useful if the `_kombu.binding.%s` pattern could be altered via the `broker_transport_options` configuration setting.

This quite common practice: for instance, when using Redis with Django for caching, you can specify such a prefix with `KEY_PREFIX`:

```
CACHES = {
    "default": {
        "BACKEND": "django_redis.cache.RedisCache",
        "LOCATION": "redis://127.0.0.1:6379/1",
        "OPTIONS": {
            "CLIENT_CLASS": "django_redis.client.DefaultClient"
        },
        "KEY_PREFIX": "example"
    }
}
```

[...]

PR #912, opened 2018-08-30
Introduce global key prefix for redis transport

wetneb:
@georgepsarakis here is a proposal for #853. Do you think this is the right approach?

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -426,6 +426,11 @@ class Channel(virtual.Channel):
     #: Disable for backwards compatibility with Kombu 3.x.
     fanout_patterns = True
 
+    #: The global key prefix will be prepended to all keys used
+    #: by Kombu, which can be useful when a redis database is shared
+    #: by different users. By default, no prefix is prepended.
+    global_key_prefix = ''
+
     #: Order in which we consume from queues.
     #:
     #: Can be either string alias, or a cycle strategy class
@@ -466,6 +471,7 @@ class Channel(virtual.Channel):
          'unacked_restore_limit',
          'fanout_prefix',
          'fanout_patterns',
+         'global_key_prefix',
          'socket_timeout',
          'socket_connect_timeout',
          'socket_keepalive',
@@ -499,6 +505,10 @@ class Channel(virtual.Channel):
             # previous versions did not set a fanout, so cannot enable
             # by default.
             self.keyprefix_fanout = ''
+
+        # Prepend the global key prefix to the two key prefixes in use
+        self.keyprefix_fanout = self.global_key_prefix + self.keyprefix_fanout
+        self.keyprefix_queue = self.global_key_prefix + self.keyprefix_queue
 
         # Evaluate connection.
         try:
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -767,7 +767,8 @@ class Channel(virtual.Channel):
 
     def _q_for_pri(self, queue, pri):
         pri = self.priority(pri)
-        return self.global_key_prefix + '%s%s%s' % ((queue, self.sep, pri) if pri else (queue, '', ''))
+        key = '%s%s%s' % ((queue, self.sep, pri) if pri else (queue, '', ''))
+        return self.global_key_prefix + key
 
     def priority(self, n):
         steps = self.priority_steps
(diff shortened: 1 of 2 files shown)


georgepsarakis:
@wetneb seems correct. This will require a small unit test here, just to verify functionality. We can then possibly add a simple integration test in Celery test suite.

Also, perhaps the same needs to be done for the Celery Redis Backend?

wetneb:
@georgepsarakis thanks for the quick review. I have added a unit test, I think the rest belongs to the Celery repository.

t/unit/transport/test_redis.py
```diff
@@ -828,6 +828,17 @@ def test_rediss_connection(self):
...
 
+    def test_global_key_prefix(self):
+        conn = Connection(transport=Transport, transport_options={
+            'global_key_prefix': 'foo',
+        })
+        chan = conn.channel()
+        c = chan._create_client = Mock()
+
+        body = {'hello': 'world'}
+        chan._put_fanout('exchange', body, '')
```
georgepsarakis: I think you should also check [Channel._put](https://github.com/celery/kombu/pull/912/files#diff-4b690d9e140dd19f7d015f4595540bcaR570), not only the fanout case.

thedrow:
@wetneb Can you please address the review from @georgepsarakis so we can merge this?

wetneb:
@thedrow it is on my todo list, but it might have to wait a bit more. If this is urgently required, feel free to take over the PR.

wetneb:
I have no clue why one test fails for Python 3.5 - can I leave it to the experts @thedrow @georgepsarakis ?

auvipy:
I have restarted the failing build

kombu/transport/redis.py
```diff
@@ -757,7 +767,8 @@ def _size(self, queue):
 
     def _q_for_pri(self, queue, pri):
         pri = self.priority(pri)
-        return '%s%s%s' % ((queue, self.sep, pri) if pri else (queue, '', ''))
+        key = '%s%s%s' % ((queue, self.sep, pri) if pri else (queue, '', ''))
+        return self.global_key_prefix + key
```
georgepsarakis: I think this is redundant, doesn't the queue names (`queue` variable here) already contain `self.keyprefix_queue`, which in its turn is prefixed with the global namespace?

t/unit/transport/test_redis.py
georgepsarakis: @wetneb did you perhaps consider the above?

kombu/transport/redis.py
wetneb: @georgepsarakis I was puzzled by this too!

I assume that because this method comes from the base interface, it is just called with the bare queue name without prefix (because prefixes are introduced by the redis implementation).
https://github.com/celery/kombu/blob/0af7519f98453bae62e0f4ad70e613e04c8fa379/kombu/transport/virtual/base.py#L361

But I might be wrong! If this method is indeed called with the prefixed queue name, then I guess the second test case becomes trivial…

t/unit/transport/test_redis.py
wetneb: @georgepsarakis I did add a test case for that, yes!

t/unit/transport/test_redis.py
```diff
@@ -828,6 +828,27 @@ def test_rediss_connection(self):
...
+
+    def test_global_key_prefix_put(self):
+        conn = Connection(transport=Transport, transport_options={
+            'global_key_prefix': 'foo',
+        })
+        chan = conn.channel()
+        c = chan._create_client = Mock()
+        message = {'hello': 'world'}
+        chan._put('queue', message)
+        c().lpush.assert_called_with('fooqueue', dumps(message))
```
wetneb: @georgepsarakis the test case for `Channel._put` is here.

wetneb:
@georgepsarakis okay I have tried to understand what is happening here and I realized that I just don't have the necessary background on kombu and redis to debug this correctly.

This change seemed fairly simple intuitively but I just cannot afford to spend hours learning all the bits involved, given that it is no longer critical for me. I have a lot of other things going on and I need to prioritize, so I will stop working on this.

I hope someone knowledgeable will come and finish this PR so that the effort is not lost.

georgepsarakis:
@wetneb that is perfectly understandable, thank you very much for your efforts so far. Sorry for not giving you better guidelines sooner, unfortunately I am not adequately familiar with certain internals either and this change seems to have much broader scope than I anticipated.

@thedrow @auvipy any idea why the integration tests did not catch the above issue? Perhaps we should open an issue to review these test cases?

I suggest we close this for the time being and leave #853 open.

thedrow:
Yes I agree.
Let's close this PR for now.

PR #923, opened 2018-09-26
Fix infinity loop in create_loop

### LINKED ISSUE #685: Prefetch with SQS enters infinite loop
I'm using kombu (4.0.2 or git master) with the SQS transport, starting Celery with `-c 2` and pretty quickly (within seconds of starting Celery, after processing a handful of items), I see the master process at 100% CPU. GDB attach, py-bt always shows 

```
Traceback (most recent call first):
  File "/var/virtualenvs/dbg/lib/python3.5/site-packages/kombu-4.0.2-py3.5.egg/kombu/async/hub.py", line 193, in call_soon
    if not isinstance(callback, Thenable):
  File "/var/virtualenvs/dbg/lib/python3.5/site-packages/kombu-4.0.2-py3.5.egg/kombu/transport/SQS.py", line 284, in _loop1
    self.hub.call_soon(self._schedule_queue, queue)
  File "/var/virtualenvs/dbg/lib/python3.5/site-packages/kombu-4.0.2-py3.5.egg/kombu/transport/SQS.py", line 293, in _schedule_queue
    self._loop1(queue)
  File "/var/virtualenvs/dbg/lib/python3.5/site-packages/vine/promises.py", line 136, in __call__
    retval = fun(*final_args, **final_kwargs)
  File "/var/virtualenvs/dbg/lib/python3.5/site-packages/kombu-4.0.2-py3.5.egg/kombu/async/hub.py", line 282, in create_loop
    item()
....
```
In the SQS channel, I see that such a loop can occur if `qos.can_consume()` always returns `False`, but I don't understand enough of the virtual channel implementation (yet) to see why it would get stuck in that state, when the SQS queue has messages available and celery workers are idling (i.e. waiting on `_recv` somewhere in `billiard`).

Any ideas or debugging tips? Are there some parameters to be tuned to avoid this?

FWIW This doesn't occur with Celery 3.1.25 & kombu 3.0.37, but we'd like to upgrade. 

_edit_ the Celery config is really simple, just pointing at a valid SQS queue and a few routes & queues defined.

_edit 2_ To wit, starting Celery with `-c 1 -P solo`, items are processed normally. I've also noticed that exactly 8 tasks are received each time before the loop, which suggests (to me) that the channel is working, but the workers aren't consuming the messages?

floatdrop:
We have problem with hanged workers (and SQS queue).

Unfortunately - #760 is not quite enough to fix infinite loop. Consider this scenario:

1. Take first item from todo
2. Item schedules itself back to self._ready and gets to the front
3. Since todo is reference to self._ready, `pop` takes it again => goto 1.

Fixes #685

### DIFF
--- a/kombu/asynchronous/hub.py
+++ b/kombu/asynchronous/hub.py
@@ -3,7 +3,6 @@
 from __future__ import absolute_import, unicode_literals
 
 import errno
-import itertools
 from contextlib import contextmanager
 from time import sleep
 from types import GeneratorType as generator  # noqa
@@ -279,24 +278,15 @@ class Hub(object):
         consolidate_callback = self.consolidate_callback
         on_tick = self.on_tick
         propagate = self.propagate_errors
-        todo = self._ready
 
         while 1:
+            todo = self._ready
+            self._ready = set()
+
             for tick_callback in on_tick:
                 tick_callback()
 
-            # To avoid infinite loop where one of the callables adds items
-            # to self._ready (via call_soon or otherwise), we take pop only
-            # N items from the ready set.
-            # N represents the current number of items on the set.
-            # That way if a todo adds another one to the ready set,
-            # we will break early and allow execution of readers and writers.
-            current_todos = len(todo)
-            for _ in itertools.repeat(None, current_todos):
-                if not todo:
-                    break
-
-                item = todo.pop()
+            for item in todo:
                 if item:
                     item()
 
--- a/t/unit/asynchronous/test_hub.py
+++ b/t/unit/asynchronous/test_hub.py
@@ -508,30 +508,27 @@ class test_Hub:
         assert list(hub.scheduler), [1, 2 == 3]
 
     def test_loop__tick_callbacks(self):
-        self.hub._ready = Mock(name='_ready')
-        self.hub._ready.__len__ = Mock(name="_ready.__len__")
-        self.hub._ready.__len__.side_effect = RuntimeError()
         ticks = [Mock(name='cb1'), Mock(name='cb2')]
         self.hub.on_tick = list(ticks)
 
-        with pytest.raises(RuntimeError):
-            next(self.hub.loop)
+        next(self.hub.loop)
 
         ticks[0].assert_called_once_with()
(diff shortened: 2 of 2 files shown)


floatdrop:
@auvipy we will try to get working test, but this bug quite dependant on where taks will be placed in `set` to reproduce hanging. Any hints will be very useful.

thedrow:
I'm not sure I understand the changes yet. I'll review more closely later on.

floatdrop:
@thedrow here's a deal:

At some point `self._ready` contains 5 tasks: one for fetching from SQS and 4 with actual work.

Fetching task is not getting anything beyond 4 already fetched tasks and exits early, but schedules itself **back to the self.ready at the front of set** (because ordering is not guaranteed in set).

Each next loop iteration now getting SQS tasks and does this 5 times. Hence infinity loop for fetching tasks from SQS and not getting to actual tasks in `self._ready`. Deadlocked.

What we propose is kind same thing, that was intentionally proposed in #760 – copy `self._ready` to `todo` variable and clear `self._ready`. But instead of copy-clear we just move set to new variable and assign new set to `self._ready` which avoids heavy copying.

Also this grabbing of `self._ready` moved into while loop to ensure, that new tasks in todo gets processed on next loop iteration.

I hope this clears purpose of this PR a little. Happy to answer any questions.

floatdrop:
We have been monitoring our workers for a 24h now and did not notice hangin ones anymore. Also this fixes problem with reappearing tasks, because ‘delete_message’ is scheduled by ‘todo’ and it was not working properly.

@thedrow do you have eta, when you can review this PR?

floatdrop:
@thedrow yeah, we thought about this too, but this will introduce quite a big change.

In fact, I would advocate to revert this commit - https://github.com/celery/kombu/commit/19e83733c26d907522616e3ab3c7e3b323eaf50e or just hear a reason, why `deque` was replaced by a `set`.

I think new structure can be added in separate issue/pr in which proper structure will be considered.

auvipy:
I would rather let this in as is, and allow another pr for refactoring things suggested by @thedrow

PR #926, opened 2018-10-07
Fixed Worker shutdown creates duplicate messages in SQS broker

mintusah25:


kombu/transport/SQS.py
```diff
@@ -377,8 +384,8 @@ def basic_ack(self, delivery_tag, multiple=False):
         except KeyError:
             pass
         else:
-            self.asynsqs.delete_message(message['sqs_queue'],
-                                        sqs_message['ReceiptHandle'])
+            self.sqs.delete_message(QueueUrl=message['sqs_queue'],
```
thedrow: So deleting messages isn't async anymore?
What's the cause of this change?

mintusah25: So deleting messages isn't async anymore?
> Yes. 
What's the cause of this change?
> Due async delete, Worker was not able to delete any message completed after warm shutdown. 
Eg: We have a message:M1 and worker1 is process that message. When worker Warm Shutdown, MainProcess is waiting till M1 processing done but due to async delete system exit before deleting that message from SQS.

kombu/asynchronous/hub.py
```diff
@@ -254,6 +254,22 @@ def close(self, *args):
...
             callback(self)
 
+        # Complete remaining todo befor Hub close
+        # Eg: Acknowledge message
+        # To avoid infinite loop where one of the callables adds items
+        # to self._ready (via call_soon or otherwise), we take pop only
+        # N items from the ready set.
+        # N represents the current number of items on the set.
+        # That way if a todo adds another one to the ready set
+        todo = self._ready
```
georgepsarakis: Nice thought to avoid a possible infinite loop. So here we give a last change for items to be processed?

kombu/asynchronous/hub.py
```diff
@@ -254,6 +254,22 @@ def close(self, *args):
...
+        current_todos = len(todo)
+        for _ in range(0, current_todos):
+            if not todo:
+                break
+            item = todo.pop()
```
georgepsarakis: `.pop` raises an error on an empty set, is that why you added the `break` statement above?

kombu/asynchronous/hub.py
```diff
@@ -254,6 +254,22 @@ def close(self, *args):
...
 
+        # Complete remaining todo befor Hub close
+        # Eg: Acknowledge message
+        # To avoid infinite loop where one of the callables adds items
+        # to self._ready (via call_soon or otherwise), we take pop only
+        # N items from the ready set.
+        # N represents the current number of items on the set.
+        # That way if a todo adds another one to the ready set
+        todo = self._ready
+        current_todos = len(todo)
```
georgepsarakis: I think you can achieve the same with the possibly simpler loop:

```python
for item in list(self._ready):
    item()
```

auvipy:
could you increase the coverage a bit?

PR #936, opened 2018-10-23
Update default login method for librabbitmq and pyamqp

daviskirk:
Using PLAIN as the default seems to have become the obvious choice:  See https://www.rabbitmq.com/authentication.html

AMQPLAIN is only retained for backwards compatability and has become non-standard.  If it's an active decision to keep AMQPLAIN I am fine with that, I've just run across many installations where people are taking AMQPLAIN out of the server config for new setups and people using kombu are wondering why their connections won't work.

### DIFF
--- a/kombu/transport/librabbitmq.py
+++ b/kombu/transport/librabbitmq.py
@@ -182,5 +182,5 @@ class Transport(base.Transport):
             'port': (self.default_ssl_port if self.client.ssl
                      else self.default_port),
             'hostname': 'localhost',
-            'login_method': 'AMQPLAIN',
+            'login_method': 'PLAIN',
         }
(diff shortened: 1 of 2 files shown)


clokep:
Any idea of what version of RabbitMQ this became the default?

This would definitely need to be documented in the release notes.

daviskirk:
@clokep Actually, apparently the default is to have both PLAIN and AMQPLAIN (for backwards compatibility) enabled: https://www.rabbitmq.com/authentication.html 
Rabbitmq 2.3 https://www.rabbitmq.com/release-notes/README-2.3.0.txt added pluggable sasl auth mechanisms and it seems that since then AMQPLAIN has been "for backwards compatability".

thedrow:
Is this a breaking change? If so, I'd rather save it for the next major version of Kombu.

auvipy (commented):
I think we can consider this for kombu 5.1

thedrow (approved):

thedrow:
@auvipy Thanks for letting me know.

PR #938, opened 2018-10-28
Removed Jython bz2 import fallback

atombrella:
This appears to have been resolved. The touched lines were from 2011, and the satus of the buf changed in 2013. I didn't check on Jython, my suggestion is only based on that report.

http://bugs.jython.org/issue1445

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,7 +10,7 @@ all_files = 1
 [flake8]
 # classes can be lowercase, arguments and variables can be uppercase
 # whenever it makes the code more readable.
-ignore = N806, N802, N801, N803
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -10,7 +10,7 @@ all_files = 1
 [flake8]
 # classes can be lowercase, arguments and variables can be uppercase
 # whenever it makes the code more readable.
-ignore = N806, N802, N801, N803
+ignore = W504, N806, N802, N801, N803
```
atombrella: `flake8` raised a lot of warnings (not locally, only on Travis) without this change.

thedrow:
Thanks!

PR #945, opened 2018-11-11
Used a tuple instead of a list to define __all__.

atombrella:
`pydocstyle` gives warnings if `__all__` is not defined as a tuple.


thedrow:
Thank you for your contribution.

PR #946, opened 2018-11-15
Pin Redis-py version

ashb:
Redis-py 3.0.0 was released today and introduced a backwards incompatible changes.

https://github.com/andymccurdy/redis-py/blob/9b03af26dc829beea232a3248768de933f4c3b67/CHANGES#L27-L29

This isn't the "best" fix but is the quickest.


thedrow:
Thanks!
Keep me posted if https://github.com/andymccurdy/redis-py/issues/1068 is resolved.

ashb:
I think the interface to `zadd` in redis changed. This line from redis-py's changelog seems to cover it:

> ZADD now requires all element names/scores be specified in a single
> dictionary argument named mapping

I.e. I think

```pipe.zadd(self.unacked_index_key, time(), delivery_tag)```

needs to become

```pipe.zadd(self.unacked_index_key, {time(): delivery_tag})```

ashb:
In the mean time what is the chance of getting a quick Kombu release out with this version pin included?
(I'm not sure what the Kombu release process looks like)

thedrow:
I'll try to get to it.

PR #948, opened 2018-11-16
Support redis-py v2 and v3

ashb:
Further to #946 this fixes the underlying issue in a easy-to-upgrade way for end users, many of whom will have redis installed via other means. By having this check here and supporting both versions concurrently it makes it easier for end users, and to use celery/kombu in projects that use redis elsewhere.

With this change it is possibly worth reverting #946

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -146,8 +146,15 @@ class QoS(virtual.QoS):
     def append(self, message, delivery_tag):
         delivery = message.delivery_info
         EX, RK = delivery['exchange'], delivery['routing_key']
+        # TODO: Remove this once we soley on Redis-py 3.0.0+
+        if redis.VERSION[0] >= 3:
+            # Redis-py changed the format of zadd args in v3.0.0
+            zadd_args = [{time(): delivery_tag}]
+        else:
+            zadd_args = [time(), delivery_tag]
+
         with self.pipe_or_acquire() as pipe:
-            pipe.zadd(self.unacked_index_key, time(), delivery_tag) \
+            pipe.zadd(self.unacked_index_key, *zadd_args) \
                 .hset(self.unacked_key, delivery_tag,
                       dumps([message._raw, EX, RK])) \
                 .execute()
--- a/t/unit/transport/test_redis.py
+++ b/t/unit/transport/test_redis.py
@@ -75,8 +75,15 @@ class Client(object):
     def sadd(self, key, member, *args):
         self.sets[key].add(member)
 
-    def zadd(self, key, score1, member1, *args):
-        self.sets[key].add(member1)
+    def zadd(self, key, *args):
+        if redis.redis.VERSION[0] >= 3:
+            (mapping,) = args
+            for item in mapping:
+                self.sets[key].add(item)
+        else:
+            # TODO: remove me when we drop support for Redis-py v2
+            (score1, member1) = args
+            self.sets[key].add(member1)
 
     def smembers(self, key):
         return self.sets.get(key, set())
@@ -840,7 +847,21 @@ class test_Redis:
     def teardown(self):
         self.connection.close()
 
-    def test_publish__get(self):
+    @mock.replace_module_value(redis.redis, 'VERSION', [3, 0, 0])
+    def test_publish__get_redispyv3(self):
+        channel = self.connection.channel()
+        producer = Producer(channel, self.exchange, routing_key='test_Redis')
+        self.queue(channel).declare()
+
+        producer.publish({'hello': 'world'})
+
+        assert self.queue(channel).get().payload == {'hello': 'world'}
+        assert self.queue(channel).get() is None
+        assert self.queue(channel).get() is None
+        assert self.queue(channel).get() is None
+
+    @mock.replace_module_value(redis.redis, 'VERSION', [2, 5, 10])
+    def test_publish__get_redispyv2(self):
         channel = self.connection.channel()
         producer = Producer(channel, self.exchange, routing_key='test_Redis')
         self.queue(channel).declare()


ashb:
@thedrow  PTAL - this might be a better fix than my previous one.

ashb:
It seemed to be the only one that we used in this file. Unfortunately I’ve just moved house so may not have much time for a week or so to make change/fix these tests

kombu/transport/redis.py
```diff
@@ -146,8 +146,14 @@ def __init__(self, *args, **kwargs):
     def append(self, message, delivery_tag):
         delivery = message.delivery_info
         EX, RK = delivery['exchange'], delivery['routing_key']
+        if redis.VERSION[0] >= 3:
+            # Redis-py changed the format of zadd args in v3.0.0
+            zadd_args = [{time(): delivery_tag})
```
ashb: Oh , this syntax might only work on py3?

thedrow: You missed a `]` and replaced it with `)`.

kombu/transport/redis.py
```diff
@@ -146,8 +146,14 @@ def __init__(self, *args, **kwargs):
     def append(self, message, delivery_tag):
         delivery = message.delivery_info
         EX, RK = delivery['exchange'], delivery['routing_key']
+        if redis.VERSION[0] >= 3:
```
thedrow: Please add a TODO to remove this once we drop support for redis-py 2.x.

kombu/transport/redis.py
andymccurdy: I suspect that you may have the arguments reversed here. redis-py 3.0 expects the dict to be a mapping where element names are keys and scores are their values. I believe this should be:

```python
zadd_args = [{delivery_tag: time()}]
```

ashb: Yeah, `zadd_args = [{delivery_tag: time()}]` was what I meant to do. Sorry, I did this via the github editor :( Fixing it now.

kombu/transport/redis.py
ashb: Done.

kombu/transport/redis.py
ashb: Fixed.

ashb:
@thedrow Fixed up the stupid mistakes, and added the TODO-comment. PTAL?

thedrow:
This looks good but the build is still failing!

andymccurdy:
> This looks good but the build is still failing!
> https://travis-ci.org/celery/kombu/jobs/456711798#L1465

Looks like there's some kind of mock interface in the tests here: https://github.com/celery/kombu/blob/master/t/unit/transport/test_redis.py#L78

Something like this would probably workl

```python
    def zadd(self, key, mapping, **kwargs):
        for key in mapping
            self.sets[key].add(key)
```

thedrow:
When this tests were written [fakeredis](https://github.com/jamesls/fakeredis) wasn't available or too premature :)

ashb:
Ran that test locally and it's passing now with the change Andy suggested

ashb:
Added tests (via case.mock.replace_module_value to switch the version back and forth. I think that should bring the coverage back up.

ashb:
@thedrow ptal. Coverage for both versions added

kombu/transport/redis.py
```diff
@@ -146,8 +146,15 @@ def __init__(self, *args, **kwargs):
     def append(self, message, delivery_tag):
         delivery = message.delivery_info
         EX, RK = delivery['exchange'], delivery['routing_key']
+        # TODO: Remove this once we soley on Redis-py 3.0.0+
+        if redis.VERSION[0] >= 3:
+            # Redis-py changed the format of zadd args in v3.0.0
+            zadd_args = [{time(): delivery_tag}]
```
Tenzer: Is this not the wrong way round? [The documentation](https://github.com/andymccurdy/redis-py#mset-msetnx-and-zadd) says:
> For ZADD, the dict is a mapping of element-names -> score.

So the key in the dict should be the element-name (in this case `delivery_tag` and the value should be the score, which here is `time()`?

I did a quick test locally to check:

```py
>>> from redis import Redis
>>> r = Redis()
>>> from time import time
>>> zadd_args = [{time(): 'test_tag'}]
>>> r.zadd('kombu', *zadd_args)
  File "/Users/jeppe/.virtualenvs/tmp-82c2c5e4c227252/lib/python3.6/site-packages/redis/client.py", line 2266, in zadd
    return self.execute_command('ZADD', name, *pieces, **options)
  File "/Users/jeppe/.virtualenvs/tmp-82c2c5e4c227252/lib/python3.6/site-packages/redis/client.py", line 755, in execute_command
    return self.parse_response(connection, command_name, **options)
  File "/Users/jeppe/.virtualenvs/tmp-82c2c5e4c227252/lib/python3.6/site-packages/redis/client.py", line 768, in parse_response
    response = connection.read_response()
  File "/Users/jeppe/.virtualenvs/tmp-82c2c5e4c227252/lib/python3.6/site-packages/redis/connection.py", line 638, in read_response
    raise response
redis.exceptions.ResponseError: value is not a valid float
>>> zadd_args2 = [{'test_tag': time()}]
>>> r.zadd('kombu', *zadd_args2)
1
```

ashb: I don't know - even on redis 2.10.6 I get this:

```python
>>> import redis
>>> redis.VERSION
(2, 10, 6)
>>> from time import time
>>> r = redis.Redis()
>>> r.zadd('kombu', time(), 'test_tag')
... 
redis.exceptions.ResponseError: value is not a valid float
```

and this is definately the same order as the code was previously.

Tenzer: Oh, that's due to the difference between the `Redis` client and `StrictRedis`. In redis-py 3 `StrictRedis` has been renamed to `Redis` and `StrictRedis` has become an alias to `Redis`.

In redis-py 2 there's a difference in the input order, it's mentioned slightly in the README for the 2.10.6 release: https://github.com/andymccurdy/redis-py/tree/2.10.6#api-reference, but it might be easier to compare the code used. This is the `StrictRedis.zadd` method: https://github.com/andymccurdy/redis-py/blob/2.10.6/redis/client.py#L1677-L1697 and here's the `Redis.zadd` method: https://github.com/andymccurdy/redis-py/blob/2.10.6/redis/client.py#L2292-L2321.

```py
>>> import redis
>>> redis.VERSION
(2, 10, 6)
>>> r = redis.StrictRedis()
>>> from time import time
>>> r.zadd('kombu', time(), 'test_tag')
1
```

Tenzer: The order was flipped around in https://github.com/celery/kombu/pull/953.

Issue #950, opened 2018-11-17
SQS channel should default to boto3 region


villasv:
Currently the region property comes from
```
    @cached_property
    def region(self):
        return self.transport_options.get('region') or self.default_region
```
Where `default_region = 'us-east-1'`

I believe the `default_region` should be set to the region that comes from boto3 (which already searches for region configurations in files and environments).

IMHO it should be set as `default_region = boto3.session.Session().region_name or 'us-east-1'` to make it more boto3-friendly, but wanted to discuss before opening a PR.

georgepsarakis:
@villasv I think you are correct. Default region can be set:
- Configuration & Profile files
- `AWS_DEFAULT_REGION` environment variable

If the `Session` instance automatically resolves them (and I think it does!) then this is the correct approach.

PR #951, opened 2018-11-19
Add boto3 default region before hardcoded default

villasv:
Implements a fix for #950, slightly different from the original proposal.

I did that because the boto3 library could be missing. I'm guessing, from the guard condition at the constructor:

```
def __init__(self, *args, **kwargs):
        if boto3 is None:
            raise ImportError('boto3 is not installed')
        super(Channel, self).__init__(*args, **kwargs)
```

So I couldn't just use boto3 in the class definition to override the default region.
Instead I think it's better to just put it as a default with higher precedence than the current default region.

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -467,7 +467,9 @@ class Channel(virtual.Channel):
 
     @cached_property
     def region(self):
-        return self.transport_options.get('region') or self.default_region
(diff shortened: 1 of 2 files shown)


villasv:
Let me just link back my project's issue that depends on this so I can keep track of things :-)
https://github.com/villasv/aws-airflow-stack/issues/51

PR #953, opened 2018-11-20
Correct order of Redis `zadd` dict elements

Tenzer:
The changed `zadd()` behaviour in redis-py 3 states:
https://github.com/andymccurdy/redis-py#mset-msetnx-and-zadd

> For ZADD, the dict is a mapping of element-names -> score.

So the dictionary generated for `zadd()` should be flipped around to have the
`delivery_tag` as the key and the score as value. This commit fixes that.


ashb:
Thanks! Sorry I was making the change slightly blind to both how `ZADD` behaves (it's been a _long_ time since I've done any redis directly) and without understanding how Kombu uses redis.

peterbe:
Any chance to have a Pypi release of this?

PR #957, opened 2018-11-28
Allow setting boto3.sqs.create_queue Attributes via transport_options

### LINKED ISSUE #958: Kombu cannot create SQS queues with server-side encryption turned on
We need to configure queues to use server-side encryption when the queues are automatically created by Kombu. This can be done by setting the `KmsMasterKeyId ` and `KmsDataKeyReusePeriodSeconds` keys of the Attributes `boto3 sqs.create_queue ` parameter.

Having a way to set arbitrary Attributes values as new features are added to SQS would be nice.

hfern:
This is useful for setting encryption options on AWS SQS queue creation. 

This implements a general solution, when new features are added to queues, clients using Kombu can just specify additional transport_options.

Fixes #958

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -9,16 +9,14 @@ task queue situations where tasks are small, idempotent and run very fast.
 
 SQS Features supported by this transport:
   Long Polling:
-    https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/
-      sqs-long-polling.html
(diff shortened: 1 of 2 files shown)


hfern:
Tagging @auvipy since they appear to be on several PR reviews.

auvipy:
which issue is this addressing?

thedrow:
This looks great but requires documentation.

hfern:
@auvipy I've just written an issue for this -- #958.

hfern:
@thedrow I've added some docs --

image

And I fixed a url rendering wrong --

image

PR #960, opened 2018-12-04
Patch conflict with new Redis 3.x API

danielquinn:
It turns out that in the switch from 2.10.x to 3.x, Redis changed its API for `redis.client.Redis.zadd` from `name, *args, **kwargs` to `name, mapping, *args, **kwargs`.  As Kombu is passing `time()` as the second argument to `zadd()`, this breaks all sorts of things.

The right fix would be to patch the code, but in the short-term this is probably a good idea.


danielquinn:
As per discussion in [the Celery pull request](https://github.com/celery/celery/pull/5216), I'm closing this.

clokep:
For reference (I couldn't find it before and just found it now!) Support for Redis >= 3 was added in #953.

PR #964, opened 2018-12-10
Enable force resizing of resource pools to prevent defaulting to 10 o…

hp685:
…n celery startup

### DIFF
--- a/kombu/pools.py
+++ b/kombu/pools.py
@@ -13,7 +13,8 @@ from .utils.compat import register_after_fork
 from .utils.functional import lazy
 
 __all__ = ('ProducerPool', 'PoolGroup', 'register_group',
-           'connections', 'producers', 'get_limit', 'set_limit', 'reset')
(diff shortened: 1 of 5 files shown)


hp685:
Validated these changes in production. Pool resizing for limit values < 10 still isn't exact. It may open more connections than what's specified. 
 @AlexHill's previous change of treating 0 as a special case is simpler and gets the job done.

thedrow:
Wait I'm confused. Why is this closed?

hp685:
@thedrow for my production use case, we disable broker pooling. Broker pool limit of 0 works correctly. However, pool resizing from a higher value (say default value 10) to a lower value (say 5) will on most occasions fail with a RunTimeError. Since the values above are small anyway, defaulting to pool of 10 doesn't make a difference. If there is a use case to resize from bigger values (say 100,200) to lower values at run time, I can reopen and resubmit this PR.

Issue #931, opened 2018-10-16
Celery Worker fails to properly handle loss of SQS connection; fails to process any further messages.


rthille:
Filing this in Kombu, because I think the fix should be here, not in Celery, but I started creating the issue in Celery, so I've attached the info requested for issues there.

## Checklist

- [X] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
- [ ] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce
Have long-running Celery App with SQS connection with Celery 4.2.1 & Kombu 4.2.1
Have AWS SQS have an issue an prematurely close the connection at the wrong time.
Similar to these issues: https://github.com/celery/kombu/issues/796 and https://github.com/celery/celery/issues/3990 (which I commented on), however the process did not appear to attempt to re-establish the connection.

## Expected behavior
Celery app handles exception and reconnects to SQS or exits (and is restarted by external systems).

## Actual behavior

Exception is logged, but no further SQS messages are processed by the celery app, and no further logs were produced until another developer kill -HUP'd the main worker process.  Here's the traceback logged:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/kombu/asynchronous/hub.py", line 136, in fire_timers
    entry()
  File "/usr/local/lib/python3.7/dist-packages/kombu/asynchronous/timer.py", line 68, in __call__
    return self.fun(*self.args, **self.kwargs)
  File "/usr/local/lib/python3.7/dist-packages/kombu/asynchronous/timer.py", line 127, in _reschedules
    return fun(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/kombu/asynchronous/http/curl.py", line 108, in _timeout_check
    self._process_pending_requests()
  File "/usr/local/lib/python3.7/dist-packages/kombu/asynchronous/http/curl.py", line 132, in _process_pending_requests
    self._process(curl, errno, reason)
  File "/usr/local/lib/python3.7/dist-packages/kombu/asynchronous/http/curl.py", line 178, in _process
    buffer=buffer, effective_url=effective_url, error=error,
  File "/usr/local/lib/python3.7/dist-packages/vine/promises.py", line 150, in __call__
    svpending(*ca, **ck)
  File "/usr/local/lib/python3.7/dist-packages/vine/promises.py", line 143, in __call__
    return self.throw()
[...]

PR #976, opened 2019-01-03
Return an empty list in AWS's _on_list_ready on a timeout response

physicalattraction:
This would circumvent the problems described here: https://github.com/celery/kombu/issues/931

I am unaware of the testing procedures in this library. Can someone check if this is a good approach?


kombu/asynchronous/aws/connection.py
```diff
@@ -170,6 +170,11 @@ def _mexe(self, request, sender=None, callback=None):
 class AsyncAWSQueryConnection(AsyncConnection):
     """Async AWS Query Connection."""
 
+    STATUS_CODE_OK = 200
+    STATUS_CODE_REQUEST_TIMEOUT = 408
+    STATUS_CODE_NETWORK_CONNECT_TIMEOUT_ERROR = 599
+    STATUS_CODES_TIMEOUT = (STATUS_CODE_REQUEST_TIMEOUT, STATUS_CODE_NETWORK_CONNECT_TIMEOUT_ERROR)
```
physicalattraction: What's the benefit of a tuple instead of a list?

physicalattraction:
Can you explain how you run the tests? When I run them either in PyCharm or in the terminal, there no tests to execute.

```
$ python -m unittest t.unit.asynchronous.aws.sqs.test_connection

----------------------------------------------------------------------
Ran 0 tests in 0.000s

OK
```

kombu/asynchronous/aws/connection.py
thedrow: Using a tuple results in a smaller memory footprint. They are also faster to allocate on CPython as far as I recall.
Granted, it's only two items but there's no need to use more memory when you can avoid it.

thedrow:
We use pytest instead of unittest.
Check https://github.com/celery/kombu/blob/master/tox.ini#L28 for the exact command run in CI.
You could also use tox yourself.
Also, please rebase. The tests in master should pass now.

physicalattraction:
> Please add test coverage for the new code branch.

@thedrow: I have no clue how to test this. I don't see any existing test case in which this method is called. Can you give me a clue?

> could you please check the flake8 error?

What is flake8?

thedrow:
You can run the tests using [tox](https://tox.readthedocs.io/en/latest/).

physicalattraction:
Running the tests is not a problem, I did this with the command `pytest -rxs -xv`. The problem I am facing is that I am unfamiliar with the test structure in this repository. I added a fix in one particular method, but I don't see any existing unit tests for this method.

I have ran flake8 and in the latest commit made sure that there are no additional warnings compared to the master branch.

thedrow:
You are correct. There are no tests for `get_list()`.
All you need to see is to mock `transform()` and ensure it uses the correct values in your case.
See https://github.com/celery/kombu/blob/c00be52bce3aa1a32a9344dc75e26347b99f96d4/kombu/asynchronous/aws/connection.py#L211

thedrow:
I'm going to merge this and open an issue about writing tests for this later.
I think this is a proper fix for the problem.

PR #981, opened 2019-01-09
Simplify kombu.asynchronous.http.get_client

ikonst:
This is code is equivalent, with the difference that if `Client.__init__` fails, that won't have the AttributeError chained to it (on Python 3). I believe this is slightly less confusing.

From a performance standpoint, this is roughly equivalent, as can be seen in:
```py
from timeit import timeit

secs = timeit("""
class Foo(object):
    pass

foo = Foo()
try:
    foo.bar
except AttributeError:
    foo.bar = 'baz'
""", number=100_000)

print(secs)

secs = timeit("""
class Foo(object):
    pass

foo = Foo()
bar = getattr(foo, 'bar', None)
if not bar:
    foo.bar = 'baz'
""", number=100_000)

print(secs)
```
on my machine resulting in:
```
1.029316808
0.9725863030000002
```

### DIFF
--- a/kombu/asynchronous/http/__init__.py
+++ b/kombu/asynchronous/http/__init__.py
@@ -18,6 +18,6 @@ def get_client(hub=None, **kwargs):
     hub = hub or get_event_loop()
     assert hub is not None
     client = getattr(hub, '_current_http_client', None)
-    if not client:
+    if client is None:
         client = hub._current_http_client = Client(hub, **kwargs)
     return client
--- a/kombu/asynchronous/http/__init__.py
+++ b/kombu/asynchronous/http/__init__.py
@@ -16,8 +16,8 @@ def Client(hub=None, **kwargs):
 def get_client(hub=None, **kwargs):
     """Get or create HTTP client bound to the current event loop."""
     hub = hub or get_event_loop()
-    try:
-        return hub._current_http_client
-    except AttributeError:
+    assert hub is not None
+    client = getattr(hub, '_current_http_client', None)
+    if not client:
         client = hub._current_http_client = Client(hub, **kwargs)
-        return client
+    return client


ikonst:
cc @georgepsarakis

ikonst:
@auvipy there is already a test for this function, and if the test didn't start failing, then it shows that we didn't break the functionality/contract

ikonst:
@auvipy how do I merge this? I don't have write access.

kombu/asynchronous/http/__init__.py
```diff
@@ -16,8 +16,8 @@ def Client(hub=None, **kwargs):
 def get_client(hub=None, **kwargs):
     """Get or create HTTP client bound to the current event loop."""
     hub = hub or get_event_loop()
-    try:
-        return hub._current_http_client
-    except AttributeError:
+    assert hub is not None
```
georgepsarakis: I am not sure this is equivalent:
- if `hub` is `None` an `AssertionError` will now be raised, instead of `AttributeError` here.
- if assertions are disabled (`python -O` flag), this code will raise an `AttributeError` on line 22 now.

As far as I can see this adds failure cases, instead of simplifying operations.

ikonst: @georgepsarakis I can rewrite it as:
```py
if hub is None:
   raise RuntimeError('hub is not initialized')
```
(I used `assert` since I noticed there's many `assert`s in the code whenever an assumption was made.)

The larger issue is that this code absolutely expects hub to be initialized. The `except AttributeError` was intended to handle `_current_http_client` being unset, but it ended up unintentionally handling `None._current_http_client`. Of course the handling code was not designed to create a hub (I'd assume it's not the task of `get_client` but I'm not very familiar with the code).

ikonst:
Let's do this instead: https://github.com/celery/kombu/pull/987

PR #987, opened 2019-01-10
Add must_get_event_loop

ikonst:
We'll use `must_get_event_loop` where we assume a non-`None` hub in the following code.


ikonst:
(I'll add tests assuming it's a good approach)

PR #990, opened 2019-01-15
[WIP] Drop Python 2: Removal of Python 2 specific codes

auvipy:
Will merge after 4.3 release. work in progress!


atombrella:
There are some remnants in `kombu/__init__.py`, see `2.5 does not define __package__`. A trivial one is also in `introduction.txt` (`print body` should be `print(body)`.

auvipy:
thank you. will complete before the release of v5

auvipy:
@thedrow you can focus on kombu 5 branch for this. for kombu 4.7 onwards 4.8, 4.9, etc the existing users won't get python2 support anymore but the other functionalities will remain to keep BC compat as much as possible with the gradual adaptation of new features which are less destructive for existing large codebases.

thedrow:
Where exactly can I pull this from?

PR #1002, opened 2019-02-06
Support SSL URL parameters correctly for rediss protocol

pizzapanther:
If you use the `rediss://` protocol as recommended in Celery:
```
result_backend = 'rediss://:password@host:port/db?\
    ssl_cert_reqs=CERT_REQUIRED\
    &ssl_ca_certs=%2Fvar%2Fssl%2Fmyca.pem\                  # /var/ssl/myca.pem
    &ssl_certfile=%2Fvar%2Fssl%2Fredis-server-cert.pem\     # /var/ssl/redis-server-cert.pem
    &ssl_keyfile=%2Fvar%2Fssl%2Fprivate%2Fworker-key.pem'   # /var/ssl/private/worker-key.pem
```
http://docs.celeryproject.org/en/latest/userguide/configuration.html#configuring-the-backend-url

The URL does not get initialized with the correct SSL parameters.

I added some extra logic in the URL parser to extract SSL parameters and pass them through correctly.

### DIFF
--- a/kombu/utils/url.py
+++ b/kombu/utils/url.py
@@ -45,21 +45,20 @@ def parse_url(url):
     """Parse URL into mapping of components."""
     scheme, host, port, user, password, path, query = _parse_url(url)
     if query:
-        keys = list(query.keys())
+        keys = [key for key in query.keys() if key.startswith('ssl_')]
         for key in keys:
-            if key.startswith('ssl_'):
-                if key == 'ssl_cert_reqs':
-                    if ssl_available:
-                        query[key] = getattr(ssl, query[key])
-                    else:
-                        query[key] = None
-                        logger.warn('Defaulting to insecure SSL behaviour.')
+            if key == 'ssl_cert_reqs':
+                if ssl_available:
+                    query[key] = getattr(ssl, query[key])
+                else:
+                    query[key] = None
+                    logger.warn('Defaulting to insecure SSL behaviour.')
 
-                if 'ssl' not in query:
-                    query['ssl'] = {}
+            if 'ssl' not in query:
+                query['ssl'] = {}
 
-                query['ssl'][key] = query[key]
-                del query[key]
+            query['ssl'][key] = query[key]
+            del query[key]
 
     return dict(transport=scheme, hostname=host,
                 port=port, userid=user,
--- a/kombu/utils/url.py
+++ b/kombu/utils/url.py
@@ -15,10 +15,16 @@ try:
 except ImportError:
     from urllib import quote, unquote                  # noqa
     from urlparse import urlparse, parse_qsl    # noqa
+try:
+    import ssl
+    ssl_available = True
+except ImportError:  # pragma: no cover
+    ssl_available = False
 
 from kombu.five import bytes_if_py2, string_t
 
 from .compat import NamedTuple
+from ..log import get_logger
 
 safequote = partial(quote, safe=bytes_if_py2(''))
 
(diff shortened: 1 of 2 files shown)


kombu/utils/url.py
```diff
@@ -38,6 +44,24 @@ def parse_url(url):
     # type: (str) -> Dict
     """Parse URL into mapping of components."""
     scheme, host, port, user, password, path, query = _parse_url(url)
+    if query:
+        keys = list(query.keys())
```
thedrow: Why do we need to make a list out the keys?

kombu/utils/url.py
```diff
@@ -38,6 +44,24 @@ def parse_url(url):
...
+        for key in keys:
+            if key.startswith('ssl_'):
+                if key == 'ssl_cert_reqs':
+                    if ssl_available:
+                        query[key] = getattr(ssl, query[key])
+
```
thedrow: Please remove additional whitespace.

kombu/utils/url.py
```diff
@@ -38,6 +44,24 @@ def parse_url(url):
     # type: (str) -> Dict
     """Parse URL into mapping of components."""
     scheme, host, port, user, password, path, query = _parse_url(url)
+    if query:
+        keys = list(query.keys())
+        for key in keys:
+            if key.startswith('ssl_'):
```
thedrow: We could turn this into a generator comprehension.
The code will be clearer that way.

kombu/utils/url.py
```diff
@@ -38,6 +44,24 @@ def parse_url(url):
...
+                if key == 'ssl_cert_reqs':
```
thedrow: Why do we need a loop if we're only searching for this key?
We can simply check for it and pop before the loop.

kombu/utils/url.py
```diff
@@ -38,6 +44,24 @@ def parse_url(url):
...
+                    if ssl_available:
+                        query[key] = getattr(ssl, query[key])
+
+                    else:
+                        query[key] = None
+                        logger.warn('Defaulting to insecure SSL behaviour.')
+
+                if 'ssl' not in query:
```
thedrow: Why is this part of the loop?

kombu/utils/url.py
pizzapanther: In python 3 `.keys()` is a generator so if I don't make it a list then I'm iterating over the dictionary and modifying it at the same time (line 63) which would cause a warning and is dangerous activity.

kombu/utils/url.py
pizzapanther: interesting thought, i'll try it out and see how it looks

kombu/utils/url.py
pizzapanther: it is not one key but also `ssl_ca_certs`, `ssl_certfile`, and `ssl_keyfile`. This is here because `ssl_cert_reqs` is the only key that needs to be modified from a string to a `ssl.` constant.

kombu/utils/url.py
pizzapanther: on the first iteration of the loop `query['ssl']` may not exist. If it does not this initializes it.

thedrow: We could extract it from the loop and initialize the key if it doesn't exist.

pizzapanther: I didn't want to do that because then there will always be an `ssl` key in the return which could potentially overwrite kwargs passed in here: https://github.com/celery/kombu/blob/master/kombu/connection.py#L157-L171

Return from `parse_url` overrides those here: https://github.com/celery/kombu/blob/master/kombu/connection.py#L189-L193

I only want to override `ssl` if there is an ssl parameter in the URL. I forget where I saw it in Celery code but it stated query string parameters can override functions kwargs. Query strings are given the final say but if they are not there and we should initialize that in the return.

kombu/utils/url.py
pizzapanther: ok I added a list comprehension instead because that cuts down on an extra line, makes the code less nested, and still works around the pitfall of `.keys` being a generator.

pizzapanther:
OK I added testing for the no SSL branch of code and coverage looks better. I think everything else is addressed sufficiently?

pizzapanther:
@thedrow ping for re-review

Issue #1004, opened 2019-02-06
New 4.3.0 release breaks Redis Sentinel connection


jobec:
The new 4.3.0 breaks the Redis Sentinel connection, triggering an error like when queueing a task:
```
redis.sentinel.MasterNotFoundError: No master found for 'celery-master'
```
While starting a celery worker gives this error
```
 ERROR - celery.worker.consumer.consumer - consumer: Cannot connect to sentinel://localhost:26379/0: No master found for 'celery-master'.
```
`kombu==4.2.2-post1` did not have this issue.

Redis library = 3.0.1

auvipy:
can you help to bisect the commit causing this?

jobec:
Looks like it happened in this commit: https://github.com/celery/kombu/commit/a59b44e9fb8d00758dd6a3c560e2f158e672580d

When running though a debugger, `self.connection.client.alt` is empty on this code line https://github.com/celery/kombu/blob/master/kombu/transport/redis.py#L1104

This causes the next part to connect to an empty list of sentinels:
https://github.com/celery/kombu/blob/master/kombu/transport/redis.py#L1107-L1111

jobec:
Celery (Django) setting:

```python
CELERY_BROKER_URL = "sentinel://localhost:26379/0"
CELERY_BROKER_TRANSPORT_OPTIONS = {
    'master_name': 'celery-master',
}
```

auvipy:
@badcure could you check?

lithammer:
I'm thinking that maybe something like this ought to fix it. But it's just guesses.

```diff
-        connection_list = []
+        connection_list = [(connparams['host'], connparams['port'])]
         for url in self.connection.client.alt:
             url = _parse_url(url)
             if url.scheme == 'sentinel':
                 connection_list.append([url.hostname, str(url.port)])

```

Or:

```diff
         connection_list = []
         for url in self.connection.client.alt:
             url = _parse_url(url)
             if url.scheme == 'sentinel':
                 connection_list.append([url.hostname, str(url.port)])
+        if not connection_list:
+            connection_list = [(connparams['host'], connparams['port'])]
```

lithammer:
Tried to fix this in https://github.com/celery/kombu/pull/1010.

hajarada:
Hello All, I know this is issue is for 4.3, but it seems that 4.2.1 has that issue as well. and the same workaround applies.
[...]

PR #1009, opened 2019-02-20
Use "_parse_url()" util to parse Sentinel master URL

### LINKED ISSUE #1007: Kombu 4.3.0 Sentinel password error
Hi, I have problem with password in sentinel url after update to kombu 4.3.0.
In release was change method, witch set host and port to redis sentinel (https://github.com/celery/kombu/commit/a59b44e9fb8d00758dd6a3c560e2f158e672580d#diff-1fdc4d4a0fe8a800a42d1341599d8679R1106). If I use password in sentinel url (e.g. 'sentinel://:password@redis-sentinel:26379/0'), method `sentinel.Sentinel` is call with `[('', 'password@redis-sentinel', '26379')]` instead `[('redis-sentinel', '26379')]`

Call finish with traceback:
  File "/lib/python3.4/site-packages/kombu/transport/redis.py", line 1113, in _sentinel_managed_pool
    **additional_params)
  File "/lib/python3.4/site-packages/redis/sentinel.py", line 182, in __init__
    for hostname, port in sentinels]
  File "/lib/python3.4/site-packages/redis/sentinel.py", line 182, in <listcomp>
    for hostname, port in sentinels]
ValueError: too many values to unpack (expected 2)

Its some other way to set sentinel password for broker and result_backend?

lithammer:
Fixes #1007

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -1102,8 +1102,9 @@ class SentinelChannel(Channel):
         additional_params.pop('port', None)
         connection_list = []
         for url in self.connection.client.alt:
-            if url and 'sentinel://' in url:
-                connection_list.append(url.split('/')[2].split(':'))
+            url = _parse_url(url)
+            if url.scheme == 'sentinel':
+                connection_list.append([url.hostname, str(url.port)])
         sentinel_inst = sentinel.Sentinel(
             connection_list,
             min_other_sentinels=getattr(self, 'min_other_sentinels', 0),
--- a/t/unit/transport/test_redis.py
+++ b/t/unit/transport/test_redis.py
@@ -1344,7 +1344,10 @@ class test_RedisSentinel:
     def test_getting_master_from_sentinel(self):
         with patch('redis.sentinel.Sentinel') as patched:
             connection = Connection(
-                'sentinel://localhost:65534/;sentinel://localhost:65535/;sentinel://localhost:65536/;',  # noqa: E501
+                'sentinel://localhost:65532/;'
+                'sentinel://user@localhost:65533/;'
+                'sentinel://:password@localhost:65534/;'
+                'sentinel://user:password@localhost:65535/;',
                 transport_options={
                     'master_name': 'not_important',
                 },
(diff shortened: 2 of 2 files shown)


kombu/transport/redis.py
```diff
@@ -1102,8 +1102,9 @@ def _sentinel_managed_pool(self, asynchronous=False):
         additional_params.pop('port', None)
         connection_list = []
         for url in self.connection.client.alt:
-            if url and 'sentinel://' in url:
```
lithammer: I must admit I'm a bit unsure why the `if url` check was in there.

t/unit/transport/test_redis.py
```diff
@@ -1353,9 +1356,10 @@ def test_getting_master_from_sentinel(self):
             connection.channel()
             patched.assert_called_once_with(
                 [
+                    [u'localhost', u'65532'],
+                    [u'localhost', u'65533'],
                     [u'localhost', u'65534'],
                     [u'localhost', u'65535'],
-                    [u'localhost', u'65536']
```
huangsam: Good catch as `2^16-1` is `65,535`

kombu/transport/redis.py
huangsam: Perhaps there were `None` results in the list beforehand? Not sure why either.

t/unit/transport/test_redis.py
lithammer: Yeah `_parse_url()` (or `urllib.parse` to be exact) was actually throwing an exception because of it.

```python
>>> _parse_url('sentinel://:password@localhost:65536/')
  File "/Users/peter/src/github.com/celery/kombu/kombu/utils/url.py", line 58, in url_to_parts
    parts.port,
  File "/usr/local/Cellar/python/3.7.2_2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/parse.py", line 171, in port
    raise ValueError("Port out of range 0-65535")
ValueError: Port out of range 0-65535
```

kombu/transport/redis.py
lithammer: Ping @badcure as the author of the check.

huangsam: This `for` loop is mentioned in https://github.com/celery/kombu/issues/1004 as well.

lithammer: So I figured this out myself. Apparently `url` can be an empty string. This happens if you provade an URL string like this: `"sentinel://localhost:65532/;"` (i.e. a trailing semi-colon).

But it's not a problem becuase `url.scheme` will also be an empty string:

```python
>>> from kombu.utils.url import _parse_url
>>> _parse_url('')
urlparts(scheme='', hostname=None, port=None, username=None, password=None, path=None, query={})
```

PR #1020, opened 2019-03-06
make redis transport sep param configurable

### LINKED ISSUE #422: Is there a reason for \x06\x16 to be used as priority separator?
It makes it really difficult to query redis from command line. 

I tried `redis-cli type "queue_name\x06\x163"` and many other variants with some sort of escaping, but couldn't make it work. 
Also as far as I understand the naming convention in redis is to use ':' as a separator. [redis.io](http://redis.io/topics/data-types-intro)

Could we maybe make it configurable via an environment variable or otherwise?

jobec:
Fixes #422

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -464,7 +464,8 @@ class Channel(virtual.Channel):
 
     from_transport_options = (
         virtual.Channel.from_transport_options +
-        ('ack_emulation',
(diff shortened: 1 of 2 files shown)


jobec:
No clue why the builds fail. They complete just fine here: 
https://travis-ci.com/jobec/kombu/builds/103320048

Issue #918, opened 2018-09-08
When broker is suddenly stops calling a task is stuck forever (Connection.ensure())


wwarne:
It's simular to #902 and https://github.com/celery/celery/issues/4296 but those issues about default_channel property and my problem is not exactly in default_channel.
 
```
software -> celery:4.2.1 (windowlicker)
 kombu:4.2.1 py:3.6.5
billiard:3.5.0.4
py-amqp:2.3.2
# tried on windows and on linux
```

default_channel works fine, it uses  `broker_transport_options` to determine max_retries and call `self.ensure_connection` with it.

Problem appears only if you send a task to celery, then turn off broker and then try to send  task again. Probably connection is still cached so it passed ensure_connection() check.
But then `Queue.on_task_call` tries to call `maybe_declare(self.binding(producer.channel), retry=True)` (it doen't use broker_transport_options or retry_policy in this call).
`maybe_declare` call `_imaybe_declare` which calls `entity.channel.connection.client.ensure`

Every function call from `maybe_declare` are passing `**retry_policy` argument but since we were called from `Queue.on_task_call` without any retry_policy this dict is empty.
As a result the `ensure()` (Connection.ensure) is used its defaults for max_retries, interval_start, interval_step etc... `max_retries=None` which means infinite retries.

Steps to reproduce
 tasks.py
```python
from celery import Celery

app = Celery('tasks', broker='amqp://worker:password@192.168.99.100:5672/myvhost')
app.conf.result_backend = 'rpc://'
app.conf.broker_transport_options = {
    'max_retries': 3,
    'interval_start': 0,
    'interval_step': 0.2,
    'interval_max': 0.5,
}


@app.task
def add(x, y):
    return x + y
```
run rabbitmq and python console
```python
>>> from tasks import add
>>> add.delay(2, 2)
<AsyncResult: 32c15fee-0d34-4baf-9c53-058996844939>
```
Don't close the python console, stop rabbitmq, and try to send new task in python console
```python
>>>  add.delay(2, 2)
Here goes infifnite loop
```

Here is a monkeypatch I end with to avoid this problem in my programm. It ensures that celery uses settings from broker_transport_options then tries to ensure a connection.
I call it once before sending tasks to celery.
```python
def monkeypath_kombu_ensure():
    import kombu
    assert kombu.__version__ == '4.2.1', 'Check if patching ensure() is still needed'
    from kombu import Connection
    original_ensure = Connection.ensure

[...]

PR #1026, opened 2019-03-24
Override missing transport options with the retry policy provided by the publisher

### LINKED ISSUE #902: When RabbitMQ is down calling a task (sending a message) is stuck forever (retry settings are ignored)
#735 was closed but the problem I mentioned in https://github.com/celery/kombu/issues/735#issuecomment-327157158 was not resolved.

Recap:
If trying to call a task when RabbitMQ is down, the call will be stuck forever, ignoring my retry_policy or retry=False.

Here's a minimal Python 2.7 example to reproduce the issue:
```python
# tasks.py
import time

from celery import Celery

app = Celery('tasks', broker='amqp://guest@localhost//')

@app.task
def some_long_computation():
    print('In some_long_computation, about to sleep for 2 seconds')
    time.sleep(2)
    print('Exiting some_long_computation')

if __name__ == '__main__':
    print('>>> About to call some_long_computation task')
    # Disable retry:
    some_long_computation.apply_async(retry=False)
    # Or with a retry policy:
    # some_long_computation.apply_async(retry=True, retry_policy={
    #     'max_retries': 3, 'interval_start': 0, 'interval_step': 0.2, 'interval_max': 0.2})
    print('>>> After calling some_long_computation task')
```

Run the worker:
```
celery -A tasks worker --loglevel=info
```

Execute the task in another shell/session:
```
python tasks.py
```

The task will complete and exit successfully.

Now, stop RabbitMQ using:
```
sudo service rabbitmq-server stop
```

Execute the task again, and you will see it's **stuck**, even though the call passed `retry=False` and it should have thrown an exception.

-------------------------------------

I re-tested it with celery and kombu 4.2.1 and 4.2.0 and in both versions it remained stuck and didn't respect `retry=False`.

Supplying a retry policy also didn't work (see the commented-out example in the code above).

I tested this with python 2.7.15 on Ubuntu 18.04.1. My test environment pip freeze:
```
amqp==2.3.2
billiard==3.5.0.4
celery==4.2.1
kombu==4.2.1
pytz==2018.5
vine==1.1.4
```

### LINKED ISSUE #1024: Producer.publish blocks forever on a closed connection
`Kombu version = 4.4.0`

When I call publish on a closed connection, it blocks forever; No exception, No error messages whatsoever. I tried to set a retry policy but The behaviour is the same.

### Reproduce
```python
from kombu import Connection

with Connection('pyamqp://localhost:8000') as conn:
    producer = conn.Producer(serializer='json')
    producer.publish(
        'Hello', errback=lambda: print('An error occured!'),
        retry=True, retry_policy={
            'interval_start': 0,
            'interval_step': 1,
            'interval_max': 5,
            'max_retries': 3,
        }
    )

    print('Published!')  # Never gets called
```

There's no Rabbitmq server on `localhost:8000` So I expect `publish` to raise a `socket.timeout` error when it's unable to connect. Instead, it just hangs forever because it's trying to establish a connection.

thedrow:
If transport options are not provided for the connection override them with the retry policy provided for the publisher.
This is done in order to workaround the fact that the default channel ensures that the connection exists to allow broker failover.
This operation uses the retry policy of the connection instead of the publisher.
See https://github.com/celery/kombu/commit/816e3dc.

This fixes #902 and fixes #1024.
celery/celery#4627 is not fixed as our celery configuration is still a mess in that matter but at least the situation is better.
This PR addresses a similar problem to #918 but does not resolve it.

### DIFF
--- a/kombu/messaging.py
+++ b/kombu/messaging.py
@@ -173,13 +173,42 @@ class Producer(object):
                 # XXX declare should be a Set.
                 declare.append(self.exchange)
 
+        old_transport_opts = None
         if retry:
             _publish = self.connection.ensure(self, _publish, **retry_policy)
-        return _publish(
+
+            # If transport options are not provided for the connection
+            # override them with the retry policy provided for the publisher.
+            # This is done in order to workaround the fact that
+            # the default channel ensures that the connection exists
+            # to allow broker failover.
+            # This operation uses the retry policy of the connection instead
+            # of the publisher.
+            # See https://github.com/celery/kombu/commit/816e3dc.
+            old_transport_opts = self.connection.transport_options or {}
+            # Since we don't want these defaults to last we copy the dictionary
+            temp_transport_opts = self.connection.transport_options.copy()
+            temp_transport_opts.setdefault('max_retries',
+                                           retry_policy.get('max_retries'))
+            temp_transport_opts.setdefault('interval_start',
+                                           retry_policy.get('interval_start'))
+            temp_transport_opts.setdefault('interval_step',
+                                           retry_policy.get('interval_step'))
+            temp_transport_opts.setdefault('interval_max',
+                                           retry_policy.get('interval_max'))
+            self.connection.transport_options = temp_transport_opts
+
+        result = _publish(
             body, priority, content_type, content_encoding,
             headers, properties, routing_key, mandatory, immediate,
             exchange_name, declare,
         )
+
+        if old_transport_opts is not None:
+            # Restore the old transport options
+            self.connection.transport_options = old_transport_opts
+
+        return result
 
     def _publish(self, body, priority, content_type, content_encoding,
                  headers, properties, routing_key, mandatory,


kombu/messaging.py
```diff
@@ -173,14 +173,43 @@ def publish(self, body, routing_key=None, delivery_mode=None,
                 # XXX declare should be a Set.
                 declare.append(self.exchange)
 
+        old_transport_opts = None
         if retry:
             _publish = self.connection.ensure(self, _publish, **retry_policy)
-        return _publish(
+
+            # If transport options are not provided for the connection
```
georgepsarakis: Perhaps the reasoning here, could be transferred to the commit message instead.

kombu/messaging.py
```diff
@@ -173,14 +173,43 @@ def publish(self, body, routing_key=None, delivery_mode=None,
...
+            temp_transport_opts = self.connection.transport_options.copy()
+            temp_transport_opts.setdefault('max_retries',
+                                           retry_policy.get('max_retries'))
+            temp_transport_opts.setdefault('interval_start',
+                                           retry_policy.get('interval_start'))
+            temp_transport_opts.setdefault('interval_step',
+                                           retry_policy.get('interval_step'))
+            temp_transport_opts.setdefault('interval_max',
+                                           retry_policy.get('interval_max'))
+            self.connection.transport_options = temp_transport_opts
```
georgepsarakis: How about transferring this operation to `Connection` and using a [context manager](https://docs.python.org/3.6/library/contextlib.html#contextlib.contextmanager)? It will be reusable too.

```python
with self.connection.set_temporary_options(interval_max=...):
    return _publish(...)
```

kombu/messaging.py
thedrow: They could but the code isn't clear without them.

kombu/messaging.py
thedrow: I agree that's better.

thedrow:
@georgepsarakis Yes that is correct.
That's because people expect that the retry_policy they provide in `publish` to apply to the connection as well.

kombu/messaging.py
auvipy: is this addressed?

thedrow: No I haven't had the time to deal with it.

auvipy: If you want me to push on your branch let me know

matusvalo (changes requested):
I am not sure whether this PR is still valid because:
1. Connection is ensured before publishing: https://github.com/celery/kombu/blob/5bed2a8f983a3bf61c12443e7704ffd89991ef9a/kombu/messaging.py#L176 (connection.ensure internally calls `_ensure_connection()` before calling `default_channel` property:
https://github.com/celery/kombu/blob/5bed2a8f983a3bf61c12443e7704ffd89991ef9a/kombu/connection.py#L541-L547
2.  In commit for comment in PR: https://github.com/celery/kombu/commit/816e3dc it is pointed that the problem is calling `ensure_connection()` method. But this method also calls `_ensure_connection()` internally

Hence in publishing before calling `default_channel` property `_ensure_connection()` is already called. It means that this PR is not needed anymore (if I am not wrong)

auvipy:
closing this for now.

PR #1029, opened 2019-03-26
qpid: correctly declare support for async event loop

rohanpm:
qpid transport supports an async interface, but this wasn't declared
properly. It was missing an appropriate 'implements' class property.

This seems to have been missed because:

- kombu 4.x was branched from 3.0.21
- qpid transport was originally added after that in 3.0.24
- Transport.implements was added in commit 6a1abb7e9 on the 4.x branch
  only, before the qpid transport was in that branch
- qpid transport code was then copied between 3.x and 4.x branches but
  addition of Transport.implements was missed.

### DIFF
--- a/kombu/transport/qpid.py
+++ b/kombu/transport/qpid.py
@@ -119,7 +119,7 @@ except ImportError:  # pragma: no cover
 from kombu.five import Empty, items, monotonic, PY3
 from kombu.log import get_logger
 from kombu.transport.virtual import Base64, Message
-from kombu.transport import base
+from kombu.transport import base, virtual
 
 
 logger = get_logger(__name__)
@@ -1406,7 +1406,9 @@ class Transport(base.Transport):
     polling_interval = None
 
     # This Transport does support the Celery asynchronous event model.
-    supports_ev = True
+    implements = virtual.Transport.implements.extend(
+        asynchronous=True,
+    )
 
     # The driver type and name for identification purposes.
     driver_type = 'qpid'
--- a/t/unit/transport/test_qpid.py
+++ b/t/unit/transport/test_qpid.py
@@ -1663,9 +1663,6 @@ class test_Transport_class_attributes(object):
     def test_verify_polling_disabled(self):
         assert Transport.polling_interval is None
 
-    def test_transport_verify_supports_asynchronous_events(self):
-        assert Transport.supports_ev
-
     def test_verify_driver_type_and_name(self):
         assert Transport.driver_type == 'qpid'
         assert Transport.driver_name == 'qpid'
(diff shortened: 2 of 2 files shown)


rohanpm:
@bmbouter looks like you're the original author of the qpid transport, could you help look at this please?  Does the explanation in the commit message make sense to you?

The background here is that we've upgraded our ancient Pulp server from a version using kombu 3.x to latest 2-master using some 4.x, and I noticed (what I consider to be) a major regression in task handling: cancelling a running task (usually) won't terminate it any more, instead the revoke request is effectively ignored until after the task completes.

Eventually I managed to figure out that on kombu/celery 3.x, celery was using its async event loop for the Pulp workers, while on 4.x it was using the fallback synchronous event loop.  That difference seems to be why task revoke wasn't terminating tasks reliably any more and this commit fixes it to re-enable usage of the async loop.

dkliban:
@rohanpm I tested this patch with 

python2-celery.noarch                  4.0.2-7.fc28                    @updates 
python2-amqp.noarch                    2.2.2-3.fc28                    @fedora  
python2-billiard.x86_64                1:3.5.0.3-3.fc28                @fedora  

And my worker continued to execute the task after I cancelled it. I confirmed that the worker is running the new code. Do you have ideas for why cancellation is still not working on my machine?

rohanpm:
@dkliban I was testing on an *older* celery, something between 3.x and 4.x, because I did "git bisect" to find the earliest commit where the problem was introduced and prepared the fix on top of that.

Retesting and winding it forward now, I found that even though this commit fixed the problem initially, another problem with the same symptoms was introduced later in celery, specifically commit https://github.com/celery/celery/commit/2f58c35340f648.

So, there have been two separate issues with the same symptom of not being able to terminate tasks when using qpid transport:

- qpid transport didn't declare support for async event loop, so fallback synchronous loop was used, which left it unable to process a revoke message in many cases as it'd already be blocking on something else when that message arrived.  Fixed by this PR.
- qpid transport didn't declare support for 'fanout' exchange type; then per https://github.com/celery/celery/commit/2f58c35340f648, pidbox / remote control component would be disabled, so it wouldn't process requests to terminate.  I have pushed a new PR for this: https://github.com/celery/kombu/pull/1034

In summary, it turns out this wasn't a complete fix and another (very similar) change is also needed.

Also, one gotcha if you are looking at testing or backporting this on kombu 4.0.  I believe the attribute back then was called 'async' and not 'asynchronous'.  I guess a patch adding asynchronous may still apply without conflicts, but may silently do nothing and the attribute would have to be renamed to 'async' for that old kombu version.

PR #1031, opened 2019-03-27
Revert "fix/Broadcast-fail-by-give-queue-kwargs"

### LINKED ISSUE #1014: kombu.common.Broadcast does not accept custom queue name
the following happens at the moment

```python
>>> import kombu.common
>>> queue_name = 'my_unique_queue'
>>> bcast = kombu.common.Broadcast(name='foo', queue=queue_name)
>>> bcast.queue == queue_name
False
```

This behavior is inconsistent with the docs for the class which claims that you can pass along a queue name to override the default behavior of generating a unique one.

This, in turn, makes using `kombu.common.Broadcast` with `celery.Celerty.conf.task_queues` and `celery.Celerty.conf.task_routes` impossible since the queue is always different.

changing the following line
https://github.com/celery/kombu/blob/d78a8fc7a7b19d5c510e194a06a2280c99a88af4/kombu/common.py#L90

to
```python3
    if queue is None:
        queue = '{0}.{1}'.format(queue or 'bcast', uuid())
```
resolves the issue, however.

I can make a PR if there's interest.

thedrow:
Reverts celery/kombu#906.

Fixes #1014

### DIFF
--- a/kombu/common.py
+++ b/kombu/common.py
@@ -87,7 +87,7 @@ class Broadcast(Queue):
 
     def __init__(self, name=None, queue=None, auto_delete=True,
                  exchange=None, alias=None, **kwargs):
-        queue = '{0}.{1}'.format(queue or 'bcast', uuid())
+        queue = queue or 'bcast.{0}'.format(uuid())
         return super(Broadcast, self).__init__(
             alias=alias or name,
             queue=queue,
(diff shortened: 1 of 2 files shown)


clokep:
I actually find the behavior added in #906 rather nice. When looking at queues in RabbitMQ you can easily see what broadcast exchange an individual queue is hooked up to.

thedrow:
I agree but it bothers some people.
So we will introduce a `unique` keyword argument instead.

PR #1032, opened 2019-03-28
Allow users to switch URLs while omitting the resource identifier

ngie-eign:
Prior to this change, one needed to specify a URL using a URI
identifier, e.g., `pyamqp://foo.bar`. This change makes it so calling
`.switch(..)` again results in switching the host, not switching the
resource identifier.

This simplifies setting up connections with just hostnames specifying
the resource identifier once, separately.

Signed-off-by: Enji Cooper <yaneurabeya@gmail.com>

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -220,16 +220,22 @@ class Connection(object):
 
         self.declared_entities = set()
 
-    def switch(self, url):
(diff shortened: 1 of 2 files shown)


thedrow:
Please check the lint errors.

auvipy:
will take care of flke8

PR #1034, opened 2019-04-02
qpid: ensure supported exchange types are declared

rohanpm:
This is a follow-up to ddba8aeaf0f68. Prior to that commit,
"implements" was missing entirely for qpid. The earlier commit
added it, but only declared that the transport was async-capable.

Turns out it's important also to declare the supported exchange
types. In particular, if the transport doesn't declare support
for the 'fanout' exchange type, pidbox wouldn't be used[1] and so
it would still not be possible to revoke a task with terminate=True
when using the qpid transport.

[1] https://github.com/celery/celery/commit/2f58c35340f648

### DIFF
--- a/kombu/transport/qpid.py
+++ b/kombu/transport/qpid.py
@@ -1408,6 +1408,7 @@ class Transport(base.Transport):
     # This Transport does support the Celery asynchronous event model.
     implements = virtual.Transport.implements.extend(
         asynchronous=True,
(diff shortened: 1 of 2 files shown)


dkliban:
@rohanpm I tested this patch along with the previous one. This fixed the cancellation problem. Thank you!

rohanpm:
Travis CI failure against fa422ff8f8b6eb seems to be an unrelated flaky test.  I pushed an update with commit message changes only to trigger a retest.

Issue #994, opened 2019-01-24
`DirectExchange` type always lookup for queue in `delivery` method


stevanmilic:
I have profiled the `apply_async` method, and I've figured out that a lot of time is being spend on `_lookup` method in [exchange](https://github.com/celery/kombu/blob/master/kombu/transport/virtual/exchange.py#L70) module if redis is the backend for celery(thus calling `SMEMBERS` function). I didn't find a clean way to add another exchange type, so in the end I've monkey patched the `DirectExchange` type to **only** execute `_put` method. I would appreciate if someone could point out how can I define custom exchange type.

Check out the snakeviz(profile) graph:
![publish_task](https://user-images.githubusercontent.com/6879030/51686157-21220300-1ff0-11e9-8b0b-fdfcf37cedb0.png)
You can see^ that 22.7s are being spent on the `_lookup` method, even more time than for the actual `_put` method.

So is this `_lookup` really required, and could it be skipped without monkey patching?

thedrow:
To be honest, I haven't looked into the implementation of the Redis broker at all.

The only place we call `SMEMBERS` is in `get_table()`:
https://github.com/celery/kombu/blob/3a0454c21d5eee7b641af31ac6e046296aa38d57/kombu/transport/redis.py#L824-L830

`get_table()` is called in `_lookup()` as you said:
https://github.com/celery/kombu/blob/3a0454c21d5eee7b641af31ac6e046296aa38d57/kombu/transport/virtual/base.py#L700-L726

Since we're talking about the direct exchange we can use `SISMEMBER` instead to check if a queue with the same name exists in the set.

We could introduce a `_lookup_direct` method to the virtual channel which will do so when needed.

Care to provide the PR?

stevanmilic:
@thedrow I've made a PR.
Although the PR does bring some optimization, it didn't solve my problem. Exchanges have only one queue set, so the lookup with `SMEMBERS` comes down to O(1) anyway.

It's up to you to decide if we need this feature.

thedrow:
Do you happen to have an analysis of how much this improved the performance of the Redis broker?
Do you know what else could be affecting performance in this case?

stevanmilic:
> Do you happen to have an analysis of how much this improved the performance of the Redis broker?

Last time I measured it, it took same amount of time.

> Do you know what else could be affecting performance in this case?

[...]

PR #1041, opened 2019-05-07
Use SIMEMBERS instead of SMEMBERS to check for queue (redis broker)

stevanmilic:


stevanmilic:
Fixes #994

kombu/transport/redis.py
```diff
@@ -829,6 +829,24 @@ def get_table(self, exchange):
...
+            return [routing_key]
+
+        key = self.keyprefix_queue % exchange
+        pattern = ''
+        queue = routing_key
+        queue_bind = self.sep.join([
+            routing_key or '',
+            pattern,
+            queue or '',
+        ])
```
auvipy: would be helpful to have some comment about that this bock achieve

kombu/transport/virtual/base.py
```diff
@@ -725,6 +725,22 @@ def _lookup(self, exchange, routing_key, default=None):
...
 
+    def _lookup_direct(self, exchange, routing_key):
+        """Find queue matching `routing_key` for given direct `exchange`.
+
+        Returns:
+            str: queue name
+        """
+        if not exchange:
+            return [routing_key]
+
```
auvipy: and this

Issue #1050, opened 2019-05-30
In memory and on_error raises 'Channel' object has no attribute 'events


Gsantomaggio:
Hi team,
The following code works without problems:

```python
def on_return(exception, exchange, routing_key, message):
    pass
conn = kombu.Connection('amqp://guest:guest@localhost//')
producer = conn.Producer(on_return=on_return)
producer.publish("Hello there!")
``` 

We have unit tests and we use memory transport, and the following code:
```python
def on_return(exception, exchange, routing_key, message):
    pass
conn = kombu.Connection('memory:////')
producer = conn.Producer(on_return=on_return)
producer.publish("Hello there!")
```

raises:

```python
/home/gabriele/git/OpenStack/.venv3.mandatory/bin/python3 /home/gabriele/git/OpenStack/oslo.messaging.mand/kombu_mem.py
Traceback (most recent call last):
  File "..kombu_mem.py", line 15, in <module>
    producer.publish("Hello there!")
  File "...venv3.mandatory/lib/python3.6/site-packages/kombu/messaging.py", line 181, in publish
    exchange_name, declare,
  File "...venv3.mandatory/lib/python3.6/site-packages/kombu/messaging.py", line 187, in _publish
    channel = self.channel
  File "/...venv3.mandatory/lib/python3.6/site-packages/kombu/messaging.py", line 212, in _get_channel
    channel.events['basic_return'].add(self.on_return)
AttributeError: 'Channel' object has no attribute 'events'
```

The `events` field is undefined.
Unless I missed something, I think this is a bug.
Thank you

auvipy:
version?

Gsantomaggio:
```
pip freeze | grep kombu
kombu==4.5.0
```

Gsantomaggio:
I think that we can close this issue since the in-memory transport does not follow the AMQP standard and it is not possible to test the `mandatory` flag.

Here is an example:
```python
import time

from kombu import Connection, Exchange, Queue, Consumer

media_exchange = Exchange('media', 'direct')


def handle_message(body, message):
    print("%s RECEIVED MESSAGE: %r" % (time.time(), body))
    message.ack()



connection = Connection("memory:///")
producer = connection.Producer(serializer='json')
producer.publish({"foo": "bar"}, mandatory=True, routing_key="does_not_exist")
producer.publish({"foo1": "bar2"}, mandatory=True, routing_key="does_not_exist")
[...]

PR #1053, opened 2019-06-03
Add events field on memory channel

Gsantomaggio:
Part of https://github.com/celery/kombu/issues/1050, with this fix
the function does not raise the exception.
The init `defaultdict(set)` is the same init from amqp lib.

### DIFF
--- a/kombu/transport/memory.py
+++ b/kombu/transport/memory.py
@@ -5,11 +5,13 @@ from kombu.five import Queue, values
 
 from . import base
 from . import virtual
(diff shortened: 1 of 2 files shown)


Gsantomaggio:
Added, please let me know if it is enough.
Thanks

auvipy:
could you check the flake8 error?

Gsantomaggio:
@auvipy ok, everything should be fine!

thedrow:
@auvipy Which milestone does this belong to?

PR #1054, opened 2019-06-03
Removed bz2 fallbacks.

atombrella:
https://docs.python.org/3/library/bz2.html bz2 is in the standard library.


clokep:
It seems there are certain configs where this isn't true, check out #938 / #1003.

atombrella:
@clokep OK. Closing.

PR #1055, opened 2019-06-05
Use a sentinel value to fix Celery's build

browniebroke:
The build from Celery is broken (see https://github.com/celery/celery/issues/5555) because one test tries to enable all serializers by calling disable_insecure_serializers(allowed=None), but because `None` is the default value, it fallbacks to the default behaviour.

Using a [sentinel value](https://treyhunner.com/2019/03/unique-and-sentinel-values-in-python/) as default instead of `None` should fix this.

Fixes https://github.com/celery/celery/issues/5555

### DIFF
--- a/kombu/serialization.py
+++ b/kombu/serialization.py
@@ -410,8 +410,10 @@ _setupfuns = {
     'application/x-msgpack': register_msgpack,
 }
 
+NOTSET = object()
 
-def enable_insecure_serializers(choices=None):
+
+def enable_insecure_serializers(choices=NOTSET):
     """Enable serializers that are considered to be unsafe.
 
     Note:
@@ -419,7 +421,7 @@ def enable_insecure_serializers(choices=None):
         can also specify a list of serializers (by name or content type)
         to enable.
     """
-    choices = ['pickle', 'yaml', 'msgpack'] if not choices else choices
+    choices = ['pickle', 'yaml', 'msgpack'] if choices is NOTSET else choices
     for choice in choices:
         try:
             registry.enable(choice)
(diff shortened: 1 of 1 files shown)


kombu/serialization.py
```diff
@@ -410,24 +410,26 @@ def not_available(*args, **kwargs):
     'application/x-msgpack': register_msgpack,
 }
 
+_noset = object()
```
georgepsarakis: Would you mind reconsidering the name? Perhaps it should uppercase as a constant?

browniebroke: Sure happy to change to whatever. `NOTSET` then?

georgepsarakis: Sounds good!

auvipy:
should we push a minor release?

vallee11:
It's not necessary to use a sentinel value unless 'None' is a valid input。The sentinel value will make the project difficult to maintain if  other files has the same problem。The or expression is simpler and easier to understand than the if expression.

auvipy:
If you have better fix in mind, please come up with a PR :)

vallee11:
https://github.com/encode/django-rest-framework/blob/master/rest_framework/request.py

line:81

https://github.com/encode/django-rest-framework/blob/master/rest_framework/fields.py

line:40
     
https://github.com/django/django/blob/master/django/db/models/fields/__init__.py

line: 46

vallee11:
I don't think it is a bug, if you insist, it needs to be revised as follows：
```
NOTSET = object()


def enable_insecure_serializers(choices=NOTSET):
    """Enable serializers that are considered to be unsafe.
    Note:
        Will enable ``pickle``, ``yaml`` and ``msgpack`` by default, but you
        can also specify a list of serializers (by name or content type)
        to enable.
    """
    choices = ['pickle', 'yaml', 'msgpack'] if choices is NOTSET else choices
    if choices is not None:
        for choice in choices:
            try:
                registry.enable(choice)
            except KeyError:
                pass

```

auvipy:
thanks, I will try to implement as you suggested.

auvipy:
https://github.com/celery/kombu/pull/1057

PR #1056, opened 2019-06-06
Use fastuuid on CPython>=3.5

thedrow:
Use FastUUID when possible.
FastUUID provides bindings to Rust's UUID library.

It is much faster and I wrote the bindings so I know they are production ready.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ before_install:
   - sudo apt install libcurl4-openssl-dev libssl-dev gnutls-dev
 install:
     - pip --disable-pip-version-check install -U pip setuptools wheel | cat
-    - pip --disable-pip-version-check install -U tox | cat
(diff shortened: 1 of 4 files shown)


auvipy:
restarted the failing builds

thedrow:
I forgot to mention the library is only available on Linux for now...

thedrow:
Something is wrong with the wheel. I'll try figuring it out on Sunday.

thedrow:
Ok I thought the build environment is the problem but apparently it's because the pip version inside the virtualenv that tox creates is too old.
I'll fix that real quick.

yun-gao:
Hi we are using centOS and AWS Linux, and when we try to install Celery which depends on Kombu, there is no correct fastuuid for us.

auvipy:
it's reverted, try kombu 4.6.3

yun-gao:
It's fixed! Thanks!

PR #1057, opened 2019-06-07
code updates

auvipy:


kombu/serialization.py
```diff
@@ -415,18 +415,18 @@ def not_available(*args, **kwargs):
...
 def enable_insecure_serializers(choices=NOTSET):
     """Enable serializers that are considered to be unsafe.
-
     Note:
         Will enable ``pickle``, ``yaml`` and ``msgpack`` by default, but you
         can also specify a list of serializers (by name or content type)
         to enable.
     """
     choices = ['pickle', 'yaml', 'msgpack'] if choices is NOTSET else choices
-    for choice in choices:
```
georgepsarakis: Why not add an early return here?

```python
return if choices is None
```

auvipy: If you have time, can you push suggested edits on my branch?

thedrow:
@auvipy Milestone?

Issue #1045, opened 2019-05-16
`maybe_declare` takes an unbound Exchange w/out retry, but not with retry


Pike:
The code path to bind the exchange to the given channel is only taken on the `_maybe_declare` code path, but not on the `_imaybe_declare`.

This was regressed by #779, AFAICT.

auvipy:
@tyarimi

thedrow:
This is fixed in #1066.
I'm closing this issue.

PR #779, opened 2017-08-04
Bugfix: infinite loop when entity.channel is replaced by revive() o…

tyarimi:
…n an connection drop

PR #1066, opened 2019-06-28
Issue 129 fix maybe declare

matteius:
I was studying this recent issue https://github.com/celery/kombu/issues/1045 and how it related to the change merged here:  https://github.com/celery/kombu/pull/779/files

I think the spirit of change in that prior kombu PR is good from @tyarimi, but the Entity (Exchange or Queue or ...) must be bound before calling entity.channel.connection.client.ensure as it was before.

My Change here factors out the logic that ensures it is bound so that it can be reused.  Since I was adjusting that code, I updated the two asserts in the file to not be asserts because I learned at Pycon that production code should not use asserts since they can be stripped out in optimize and other things, it is better to make logical checks and raise appropriate exceptions.   Asserts are good for unit tests and other code that does not get run in production.

Anyway, I think this restores the behavior before the regression by ensuring the entity is bound to the channel before calling to the connection to ensure it uses the retry policy in _imaybe_declare

### DIFF
--- a/kombu/common.py
+++ b/kombu/common.py
@@ -122,7 +122,8 @@ def maybe_declare(entity, channel=None, retry=False, **retry_policy):
 
 
 def _ensure_channel_is_bound(entity, channel):
-    """Make sure the channel is bound to the entity
+    """Make sure the channel is bound to the entity.
+
     :param entity: generic kombu nomenclature, generally an exchange or queue
     :param channel: channel to bind to the entity
     :return: the updated entity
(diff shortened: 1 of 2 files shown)


matteius:
Yeah, I still need to add tests.  The PR that caused this regression requested testes that were never added and it got merged, it seemed obvious that would have caused this issue report but do plan to go back and try to add some tests.  There was a comment in that other PR about a testing strategy I need to review more to get to that point.

auvipy:
actually, we took a risk at that time.

PR #1069, opened 2019-07-02
call out specific vine requirement

matteius:
Checking if this fixes appveyor build.  @thedrow mentioned that it was because pip was using easy_install, but I see that in the tox conf a few places that install from the dev.txt which was just pointing at the master tarballs for github vine.

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,2 +1,2 @@
 https://github.com/celery/py-amqp/zipball/master
-https://github.com/celery/vine/zipball/master
+vine==1.3.0


thedrow:
I already fixed that by using tox.
The problem is that curl doesn't provide wheels for windows.
Maybe we should remove it from the Windows build entirely.

thedrow:
Note that master does not contain 5.0 code.

matteius:
Yeah sorry I was not done working on it yet, but having a PR is the only
way for me to kick off appveyor builds and learn more about it.  Kind of
looks like appveyor is always running on Windows bit tries all the tox
builds.  Anyway I am hiking today and don't have time to work on it every
day but still planning to pick it apart more and learn more about tox.

PR #1071, opened 2019-07-05
Switch from pkg_resources importlib-metadata

davidszotten:
See e.g. https://github.com/pypa/setuptools/issues/510#issuecomment-463667124

On a moderatly large project (~200 installed packages) at $work this saves ~150 ms on
`python -c 'from kombu import Exchange'`

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -4,6 +4,11 @@
  Change history
 ================
 
+Unreleased
+==========
+
+- Use importlib-metadata instead of pkg_resources for better performance
+
 .. _version-4.6.3:
 
 4.6.3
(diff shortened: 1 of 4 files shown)


davidszotten:
looks like appveyor failed to install some unrelated dependency

auvipy:
should we wait for this change into a major version like 4.7?

davidszotten:
don't feel strongly either way.  what's the typical release cadence for major and minor (=patch?) versions atm?

i should also note that i don't use any kombu plugins myself. i made a dummy one to check this works (since the tests mock it out) but it would be good with some more tests if you know of projects making use of plugins

auvipy:
report about breakage of kombu with py2.7 here https://github.com/celery/kombu/issues/1096 could you check?

PR #1072, opened 2019-07-11
fixes to librabbitmq to support passing ssl flag

dhananjaysathe:
Added basic ssl support to librabbitmq in https://github.com/celery/librabbitmq/pull/134/files would like kombu to pass this on and not error out


dhananjaysathe:
Yep the test depend on a new release from the librabbitmq-c codebase these test likely use the older one and hence blow up

Issue #1035, opened 2019-04-03
Fix Windows Build


thedrow:
Following the release of vine 5.0.0a1 and py-amqp 5.0.0a1 the AppVeyor build started failing.

This happened since we were using `python setup.py test` in that environment instead of tox.
That command installs dependencies using `easy_install` which apparently does not implement [PEP-440](https://www.python.org/dev/peps/pep-0440/). Therefore, it picks the latest available version although it is an alpha version unlike pip which requires the `--pre` option to install alpha versions.

An attempt was made to migrate that build environment to tox, but tox installs our extra dependencies such as pycurl which does not release wheels for Windows.

We can either:

- Create a tox environment for Windows which avoids installing any extras.
- Work our way to resolve each and every build issue of our dependencies.

matteius:
This is an area of interest to me as I run windows 10 and us VirtualBox to run ubuntu images for 99% of my development efforts these days.   I even see I have some stashed billiard changes to try and debug some issues related to this a while back when I saw the appveyor builds always failing.   There are some areas I think could be improved for running in a windows environment and it is easy for me to test some of that based on my setup, but it is difficult since the automated CI is always broken around that.  

While it may be possible to get to the second bullet of "Work our way to resolve each and every build issue of our dependencies." it may be useful to do the first thing of "Create a tox environment for Windows which avoids installing any extras." so that we can actually have a clean environment to get to that second part.

thedrow:
We probably need to remove curl from Windows CI altogether until pycurl/pycurl#569 is resolved.

mpuhacz:
> We probably need to remove curl from Windows CI altogether until [pycurl/pycurl#569](https://github.com/pycurl/pycurl/issues/569) is resolved.

It's also possible to install libcurl from a prebuilt wheel. Unfortunately, I don't have access to a windows machine.
https://www.lfd.uci.edu/~gohlke/pythonlibs/#pycurl

matteius:
I have spent some hours on this looking at curl and libcurl from different angles.  I got pretty far with hacking through the setup.py and winbuild.py in pycurl but still ran into downstream dependencies with libcurl not being in the paths I searched on appveyor (though I did not check every path) it was a long hack to get a build to go this far:  https://ci.appveyor.com/project/ask/kombu/builds/26551748/job/3tt1881v9aggvvyu

[...]

PR #1076, opened 2019-08-06
Fixing MongoDB backend to work properly with TTL

mpuhacz:
This PR fixes an issue with MonogDB backend which caused ignoring TTL indexes. Interestingly, the previous implementation used a keyword argument for pymongo collection update method that hasn't existed. Besides that, we are also bumping pymongo version to the latest one.

Some of appveyor tests should fail: #1035


mpuhacz:
Some of the tests seem to be failing due to network problems during `before_install` phase. Could you please rerun CI?

thedrow:
I don't know anyone who is using the MongoDB broker. I don't mind merging this though.
In any case, the minimum version should be what the MongoDB [result backend](https://github.com/celery/celery/blob/master/requirements/extras/mongodb.txt) supports.

.travis.yml
```diff
@@ -9,16 +9,19 @@ env:
...
   - test
   - lint
-  
+
 
 matrix:
   include:
     - python: 2.7
       env: TOXENV=2.7
+      before_install: sudo apt-get update && sudo apt-get install libgnutls-dev
```
thedrow: Why do we need to re-add this?

requirements/extras/mongodb.txt
```diff
@@ -1 +1 @@
-pymongo>=2.6.2,<3.0
+pymongo>=3.8.0
```
thedrow: Version should be `>=3.3.0`.

.travis.yml
mpuhacz: Travis build was failing without this package. See: https://travis-ci.org/celery/kombu/jobs/568362312

mpuhacz:
@thedrow Thanks for the review. MongoDB backend is quite useful especially if you don't want to integrate additional components into your backend. As long as the message throughput is kept on a reasonable level Mongo is doing pretty well.

Issue #1079, opened 2019-08-13
utils.functional.retry_over_time() treats max_retries=0 as infinite


EricHigdon:
In ensure()._ensured, max_retries=0 means that the connection is never retried.
In retry_over_time(), however, setting max_retries to 0 is treated the same as None.
This means that the Connection.default_channel must retry at least once.
Changing line 346 of utils.functional.py to explicitly check if max_retries is None should fix this issue.

PR: #1080

PR #1080, opened 2019-08-13
Make sure that max_retries=0 is treated differently than None

EricHigdon:
When max_retries=0, it should not retry at all instead of retrying infinitely. See #1079


thedrow:
Can you please adjust the tests?

EricHigdon:
@thedrow Right, I fixed the test that was failing and added one for max_retries=0.

Looks like some of the AppVeyor test fail, but the travis-ci ones are good. Is that a known issue?

thedrow:
Yes we have an issue with the Windows build.
Thank you!

PR #1083, opened 2019-08-18
Issue 1035 windows build

matteius:
Scrapped the last PR with the tags inadvertently because I recreated my branch after I had an idea to use the prior pycurl version which has all the binary wheels uploaded to pypi.  The ticket they are working in pycurl is related to why the current version of pycurl could not upload those wheels--so lets try a build with the prior version here and see how appveyor does.


matteius:
Ok -- sometimes stepping away from the computer for a bit and living life leads to revised thinking that may have been better than the brains prior ideas.  After spending sometime on the pycurl thread that @thedrow linked in on my other PR, I realized that that something changed in pypi validation that rejected the wheels on their latest build.  However it was Ask that added the pycurl requirement without a pin 3 years ago, and the prior version that was released Jun 2, 2018 has all the wheels to let appveyor happily build as before.  I propose that we merge this in pinned and when we want to upgrade past this version we check on https://github.com/pycurl/pycurl/issues/569

Issue #1082, opened 2019-08-17
Kombu 4.6.4 breaks Openstack with python 3.7


2uasimojo:
I'll start by saying I don't really know what kombu is/does. I'm basically proxying for [1].

When we upgraded to kombu 4.6.4, we started seeing failures like this in python3.7:

2019-08-16 22:56:25.446  |   File "/usr/local/lib/python3.7/dist-packages/eventlet/green/os.py", line 107, in open
2019-08-16 22:56:25.446  |     fd = __original_open__(file, flags, mode, dir_fd=dir_fd)
2019-08-16 22:56:25.446  | TypeError: open: path should be string, bytes or os.PathLike, not _NormalAccessor

Here's some reports from other places [2][3][4].

I'm here because I was asked [5] to link to a kombu bug.

Thanks,
efried

[1] http://lists.openstack.org/pipermail/openstack-discuss/2019-August/008553.html
[2] https://github.com/eventlet/eventlet/issues/534
[3] https://github.com/nameko/nameko/issues/655
[4] https://bugs.launchpad.net/nova/+bug/1840551
[5] http://eavesdrop.openstack.org/irclogs/%23openstack-release/%23openstack-release.2019-08-17.log.html#t2019-08-17T17:45:16

auvipy:
are you using kombu 4.6.4 with celery 4.4.0rc3?

2uasimojo:
In my local repro of the problem, I don't have any celery at all.

matteius:
@2uasimojo Any idea what your file variable is then thats being passed in fits not a string, bytes or os.PathLike.  I'll start by saying I have no idea what a_NormalAccessor object is, but when I searched Google for that error the first thing that come up is this eventlet bug and you have eventlet in your stack trace, so I would start there:   https://github.com/eventlet/eventlet/issues/534

Kombu is a component of celery that "The author of this package has not provided a project description" on pypi so in some ways its entirely open to interpretation what kombu is/does which kind of makes my job hard.  Some believe kombu is an algae or edible kelp from mostly the family Laminariaceae and is widely eaten in East Asia.  However I can say in this context its got underlying classes and utilities for managing connections and pools of workers for Celery.   

I guess though any joking aside, the docs say kombu is a messaging library and I would have to agree, and you can read more about it here:  https://kombu.readthedocs.io/en/stable/introduction.html#about

matteius:
[...]

PR #1086, opened 2019-08-20
Use importlib.metadata from the standard library on Python 3.8+

hroncok:
Note that I've tried to run tox, but was getting some version conflicts in amqp, missing test deps and later this failure:

```
self = <kombu.transport.pyro.Transport object at 0x7fe36537e0a0>

    def driver_version(self):
>       return pyro.__version__
E       AttributeError: 'NoneType' object has no attribute '__version__'

kombu/transport/pyro.py:119: AttributeError
```

I don't think it is related, so I've decided not to investigate further.


tox.ini
```diff
@@ -34,6 +34,7 @@ basepython =
     3.5: python3.5
     3.6,apicheck,pydocstyle: python3.6
     3.7: python3.7
+    3.8: python3.8
```
auvipy: is this released yet?

clokep: Python 3.8 is not yet released. Beta 3 was released recently. Looks like they're [targeting October 21, 2019](https://www.python.org/dev/peps/pep-0569/#schedule).

hroncok: Exactly. We are updating to [Python 3.8 in Fedora 32](https://fedoraproject.org/wiki/Changes/Python3.8) already.

auvipy: remove this. we will need this in Travis as well

kombu/utils/compat.py
```diff
@@ -7,7 +7,11 @@
 from functools import wraps
 
 from contextlib import contextmanager
-import importlib_metadata
+
+try:
+    from importlib import metadata as importlib_metadata
```
auvipy: this import fails in python 3.7

hroncok: Correct.

tox.ini
hroncok: Done.

kombu/utils/compat.py
hroncok: That's why it is in a try block.

auvipy: why not try import_metadata first?

hroncok: To prefer the standard library module if it is there, over the backport library.

auvipy: actually, kombu 4.6 ain't gonna get python 3.8 support. and 3.8 isn't stable yet.

hroncok: OK.

kombu/utils/compat.py
```diff
@@ -7,7 +7,11 @@
...
+except ImportError:
+    import importlib_metadata
```
matteius: This line is still going to cause an issue for eventlet and therefore open stack folks (because they use eventlet).  Celery [does support eventlet](https://docs.celeryproject.org/en/latest/userguide/concurrency/eventlet.html).  Concurrent thread:  https://github.com/celery/kombu/issues/1082

auvipy: good catch Matt

matteius: Actually though -- if we ensure eventlet users try the fix that @davidszotten mentioned "eventlet 0.25.1 has been released with a fix for the pathlib issue" maybe this code would be safer.

hroncok:
I can try tomorrow.

Issue #1087, opened 2019-08-20
Latest working celery/redis cannot inspect: Error: No nodes replied within time constraint.


jslusher:
- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

#### Related Issues

- https://github.com/celery/celery/issues/1456
- https://github.com/celery/celery/issues/4688
- https://github.com/celery/celery/issues/3453

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 4.3.0

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
software -> celery:4.3.0 (rhubarb) kombu:4.6.4 py:2.7.16
            billiard:3.6.1.0 redis:3.2.1
platform -> system:Linux arch:64bit
            kernel version:3.10.0-957.27.2.el7.x86_64 imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:sentinel results:disabled

CELERY_QUEUES:
    (<unbound Queue celery -> <unbound Exchange celery(direct)> -> celery>,
 <unbound Queue fast -> <unbound Exchange fast(direct)> -> fast>,
 <unbound Queue slow -> <unbound Exchange slow(direct)> -> slow>,
[...]

Issue #1081, opened 2019-08-16
Kombu 4.6.4 breaks celery inspect


olemoign:
Hi,

It had been mentioned that kombu 4.6 was breaking celery workers pings: https://github.com/celery/kombu/issues/1052
The guilty code was reverted, then, in 4.6.4, the revert has been reverted. https://github.com/celery/kombu/commit/052f7601d3a1e09263cc0ebda7ef8c16e6b90175

And ... celery workers pings are broken again.

I'm on holidays, I would be happy to help if I can be of assistance.

arikfr:
Same thing happened to us as well. 

It's somewhat worrying that the revert was reverted without actually checking that the issue was re-introduced, considering how core this functionality is.

auvipy:
we need to find a better solution

matteius:
I was just researching the same ticket on the other side and came to the same conclusion -- I don't quite understand the chain of events because I followed the latest commit of what I saw the obvious code change was in the redis module and it lead me through these two revision loops.  So .. what would give me confidence is understanding why the current code in 4.6.4 is calling new method _lookup_direct that did not exist in 4.6.3 -- in 4.6.4 _lookup_direct has two implemtnations, one in base and one in redis but prior it all just called the one _lookup method that still exists in base.

matteius:
I am going to X-Post this to the other thread as well because it is clear that this is why the logging message doesn't occur (there is no log in the redis version).

`# Calls From Base (4.6.3)
def _lookup(self, exchange, routing_key, default=None):
    """Find all queues matching `routing_key` for the given `exchange`.

    Returns:
        str: queue name -- must return the string `default`
            if no queues matched.
    """
    if default is None:
        default = self.deadletter_queue
    if not exchange:  # anon exchange
        return [routing_key or default]

    try:
        R = self.typeof(exchange).lookup(
            self.get_table(exchange),
            exchange, routing_key, default,
        )
    except KeyError:
        R = []

    if not R and default is not None:
        warnings.warn(UndeliverableWarning(UNDELIVERABLE_FMT.format(
            exchange=exchange, routing_key=routing_key)),
        )
        self._new_queue(default)
        R = [default]
    return R

# Calls From Redis file now (4.6.4)
def _lookup_direct(self, exchange, routing_key):
    if not exchange:
        return [routing_key]

    key = self.keyprefix_queue % exchange
    pattern = ''
    queue = routing_key
    queue_bind = self.sep.join([
        routing_key or '',
[...]

PR #1089, opened 2019-08-22
Issue #1087 redis fix

matteius:
This is the start of a PR to address issues introduced with:  https://github.com/celery/kombu/pull/1068/files
Reported as:
https://github.com/celery/kombu/issues/1087
https://github.com/celery/kombu/issues/1081
Possibly others ...

This does not revert the entire change from @auvipy and instead assumes that Redis implementation.  I would be curious of anyone with bug reports tried if that resolves their issue.  I also would be curious to hear from auivpy more about this change.  I see the intent was some kind of optimization O(1) vs O(n) but was this ever actually a performance issue?   Do we know for sure that the _lookup_direct method being called from base now is working well?  I would assume so but also we haven't upgraded up to this version that was affected yet.


matteius:
I'll take a look again tomorrow -- time for rest.

auvipy:
this is the main issue https://github.com/celery/kombu/issues/994 it would e great if you could check this first.

thedrow:
@matteius This was a potential performance issue.
However, it didn't solve #994.

I still think we need to figure out why it breaks our Redis broker implementation.

matteius:
I am going to do more analysis soon, but my hunch is that the method signature changed to not contain a default -- but also its confusing code because we are cross pollinating discussion about Redis data structures but the code changes seemed to change even how we communicate with the sockets beyond just how we use the data structures underneath Redis.  The sun is shining and I've got pressure washing to do, so I'll check back later.

matteius:
Okay -- waiting for the appveyor build to catch up but I think we have something here, I am not entirely sure what test case I would add to demonstrate the changes here, because partly I am trying to fix the Redis bug introduced in 4.6.4.  I did not just want to back out @auvipy changes because I think he made some good progress on self documenting the code with variable names and adding the O(1) vs O(n) enhancement.  Also I did not want to keep the suggestion yet for a private _lookup_direct method series because prior we had just the one base _lookup method, and then ended up 3 private methods in 4.6.4 related to _lookup -->  I think having two right now we can try to test where this one is a redis specific implementation, will be useful.  I kept parity between the try/except error handling and log warnings between the two methods.

kombu/transport/redis.py
```diff
@@ -829,23 +831,45 @@ def get_table(self, exchange):
...
+            redis_key = self.keyprefix_queue % exchange
+            queue_bind = self.sep.join([
+                routing_key or '',
+                pattern,
+                queue or '',
+            ])
+            with self.conn_or_acquire() as client:
+                if client.sismember(redis_key, queue_bind):
+                    result = [queue]
+                if not client.scard(redis_key):
```
thedrow: Why is this below `sismember`?
Won't `sismember` return false anyway if `redis_key`'s cardinality is zero?

kombu/transport/redis.py
```diff
@@ -829,23 +831,45 @@ def get_table(self, exchange):
...
+                    raise InconsistencyError(NO_ROUTE_ERROR.format(
```
thedrow: Are we sure we want to raise an exception instead of returning `[]` as before?
Why change this?

See https://github.com/celery/kombu/blob/v4.3.0/kombu/transport/virtual/base.py#L718 for example.

kombu/transport/redis.py
matteius: Yup good catch -- I mean't for it to go first.

kombu/transport/redis.py
matteius: Yes, it is subtle but that method you referenced calls lookup which has as one of its arguments self.get_table(exchange) -- which will raise this InconsistencyError if there are no members in the set matching that key.   This is why I generally prefer that code not be so deeply nested -- I'd prefer that the call in base define the variable name that is get_table and pass that in, because it becomes self documenting code -- its not clear initially to the reader without much analysis that get_table is going to return all members in the set, nor is it clear that the call to lookup might raise an InconsistencyError but really anything that might raise its own error should probably be on its own line instead of tightly nested in a function call.

matteius: https://github.com/celery/kombu/blob/master/kombu/transport/redis.py#L829

matteius: What I am mostly unclear on is if the patrern is actually always going to be an empty string '' for redis?

matteius:
I will soon -- I need to get some rest, but I started adding a couple redis test cases, I just haven't perfected them yet.  Back soon.

t/unit/transport/test_redis.py
```diff
@@ -33,151 +35,104 @@ def poll(self, timeout):
...
+        self._sock = FakeRedisSocket(server=server)
 
-    def hdel(self, key, k):
-        self.hashes[key].pop(k, None)
+    def disconnect(self):
+        self.disconnected = True
 
-    def sadd(self, key, member, *args):
-        self.sets[key].add(member)
+    # def send_command(self, cmd, *args):
```
thedrow: If its commented out just remove it.

matteius:
Ok this has been a lot of work, but I think it will end up being worth it.  I had check point my progress or I wasn't going to be able to think about anything else ever again.  As a result, I commented out two test Redis test cases that are still failing for me -- test_publish__get and test_publish__consume.  

The problem I ran into here, and why I probably have going on 30 hours on this single effort, is that in refactoring one things and trying to add some tests, I found that the test mocks were insufficient.  I did not like much what we had for a FakeRedisClient and it was mostly duplicated and partial from fakeredis.  Then I went on to learn that fakeredis is partial from what redis actually implements because it doesn't do much at all for pipeline execution although everything else works pretty well from that library.  I had to extend from some of those mocks in kombu as well in order to support the Channel and Transport concepts--as such I've added a FakeRedisKombuChannel and a FakeRedisKombuTransport.  In addition to providing the following fakeredis extensions:  FakeRedisSocket, FakeRedisConnection, FakeRedisConnectionPool, FakeRedisClient, my own FakePipeline concept to support the other fakeredis stuff, and renamed the old mock Pipeline class to FakePipelineLite. 

I need to revisit the two commented out tests as I think they are conceptually important but I need to study them more.  At least getting a fully passing CI build lets me regain confidence in my efforts here both with myself and with the core developers, as well as clean up my linting errors and made myself delete the commented out blocks and the 1000+ pprint statements I added because I haven't remembered in a while why I know how to use an IDE/debugger for next time.

Anyway, this is ready for a more full fledged review and PR feedbacks now.

thedrow:
Can you please check if you resolved #994?

matteius:
@thedrow  I really don't have time to run a code profile on it anytime soon, the next bit of time I do have will be to revisit those two unit tests and the decrease in coverage of kombu/transport/redis.py | 82.18% <84.44%> (-8.27%)

Perhaps @stevanmilic can re-run a trace using this branch and also to check if this code works in their code path?  Might be easier for you since you ran the original trace on this code.  To be honest I don't actually use the kombu redis backend, but I have a bunch of Redis experience and its a pet interest of mine, so I wanted to help out here in that space as well.

The concepts that @auvipy implemented in the change to use sismember is here in my PR, but it holds more true to the original _lookup method and it was that prior change that broke causing the issue report for this PR.  I think this mends both worlds together in a working well, but could use help testing a real integration outside of faked out unit tests.  Thanks!

matteius:
@auvipy Yeah you would not be able to make sense of it if you go commit by commit now -- I can get this branch down to a single commit though.   The changes are not very much so large to review--most of the changes are to the tests and once I correct the code coverage issue this build will be fully passing your CI.  Changes that are included not related to test files are:

- Updated _list to be more like it was before but using the proposed redis commands optimization to be O(1) -- this removes the broken _lookup_direct method and restores a consistent _lookup api.
- Removed any code paths for redis.VERSION[0] < 3 since we assert and require at least 3 already.
- Renamed certain method and variables their usages (easily using pycharm refactor) to have readable easy to understand method names that don't require a pHD in English acronyms to understand, method renamed _q_for_pri --> _queue_for_priority.
- Requires fakeredis now for tests

and thats it, production facing code changes here are very minimal and fixes a bug.

matteius:
@auvipy Alright I recreated my branch with minimal commit history of just the diff of current changes, and then I added some more unit test coverage to get the coverage check to pass.  Right now all 4 checks pass.  My only TODO is really to revisit those two unit test cases I have commented out because they don't work right my current mocks -- but I plan to do something more with those.  Otherwise though I can't resolve your change you requested because its tied to a commit that no longer exists as I force recreated my branch in order to satisfy your requested change.

t/unit/transport/test_redis.py
```diff
@@ -835,88 +953,86 @@ def test_rediss_connection(self):
...
+    #     self.queue(channel).declare()
+    #
+    #     producer.publish({'hello': 'world'})
+    #
+    #     assert self.queue(channel).get().payload == {'hello': 'world'}
+    #     assert self.queue(channel).get() is None
+    #     assert self.queue(channel).get() is None
+    #     assert self.queue(channel).get() is None
+
+    # @skip("This test is currently broken")
```
auvipy: disabling these two tests should be removed? or kept for future improvement? don't they affect any current code?

auvipy: OK

matteius: I've moved these tests down to a new test class that is based on mocks like they originally used.  Now the tests pass -- will see how the latest build goes, maybe a flake8 error to fix, maybe not.

matteius:
Hey guys -- @thedrow and @auvipy -- I've finished all my TODOs and restored those two commented out test cases as their own new TestCase class that utilizes Lite mocks like the ones that were in the test_redis.py file prior.  All 4 CI checks have passed successful--awaiting your final approvals.

t/unit/transport/test_redis.py
```diff
@@ -229,42 +444,97 @@ def _get_response_error(self):
...
         return ((KeyError,), (IndexError,))
 
 
+class FakeRedisKombuTransport(FakeRedisKombuTransportLite):
+    Channel = FakeRedisKombuChannel
+
+
 @skip.unless_module('redis')
-class test_Channel:
+class TestRedisChannel(unittest.TestCase):
```
auvipy: can't we use pytest alternative here?

matteius: I mean probably, although I didn't understand why kombu includes its own version that it already uses everywhere, and I didn't want to introduce too much scope creep for this already overly large set of changes/code diff.

matteius:
@auvipy  Are you able to resolve that old change requested that links off to a commit that no longer exists?  Is that how we get the code coverage bot to post an updated report because the last comment from the bot is definitely way stale at this point.

t/unit/transport/test_redis.py
matteius: @auvipy I am proposing it is a good suggestion that we open as a new ticket, after the next kombu release (or you can open the ticket now).  However I want to suggest something that I think could be awesome for this project and the other Celery dependencies, but lets start with Kombu.  Following the next release of Kombu I would like us to find like 2 determine and schedule two time slots a week for up to 1-hour each that are recurring optional invites and when available we join and groom the Issues report.  By treating that backlog as an Agile scrum backlog of sorts we can clean it up as a group, I mean if myself or you or someone else knows of issues that were fixed and we can relate that back to open issues in a group session and determine which tickets to close and update we can get the list much much smaller than in a shorter amount of time than working entirely disparately.   

Also, for me I became an Agile CSM -- Certified SCRUM Master in October of 2015 and have been serving as a Developer and CSM for my team at work since that time.  They recently changed requirements for renewal of the certification to do additional continuing education units, and I am hoping to petition them that opens source work such as this should be able to count for some of those hours when paying the renewal fees.  Let me know what you think, we can get some kind of WexEx link that we can join and crush this backlog in prep for Celery 5!

PR #1093, opened 2019-08-31
Add test for Retry policy timeout.

danidee10:
Related to #1024

I add a new Pytest plugin (`pytest-timeout`) to stop the tests so It doesn't run forever in Travis/Appveyor

### DIFF
--- a/t/unit/test_connection.py
+++ b/t/unit/test_connection.py
@@ -472,6 +472,19 @@ class test_Connection:
 
         myfun.assert_called()
 
+    @pytest.mark.timeout(5)
+    def test_retry_policy(self):
+        with pytest.raises(OperationalError):
+            conn = Connection('pyamqp://localhost:8000')
+            producer = conn.Producer(serializer='json')
+            producer.publish(
+                'Hello world!', retry=True, retry_policy={
+                    'interval_start': 0,
+                    'interval_step': 1,
+                    'interval_max': 5,
+                    'max_retries': 3,
+                })
+
     def test_SimpleQueue(self):
         conn = self.conn
         q = conn.SimpleQueue('foo')
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -2,5 +2,6 @@ pytz>dev
 case>=1.5.2
 pytest
 pytest-sugar
+pytest-timeout
 Pyro4
 fakeredis==1.0.4


t/unit/test_connection.py
```diff
@@ -472,6 +472,19 @@ def on_call(*args, **kwargs):
 
         myfun.assert_called()
 
+    @pytest.mark.timeout(5)
+    def test_retry_policy(self):
```
matteius: My question abut this test -- my assumption is the Connection is not mocked and the localhost pyamqp isn't going to be there for a unit test.  Looking up the retrying policy in the [retry documentation ](https://docs.celeryproject.org/projects/kombu/en/stable/userguide/producers.html) it would appear this test will First rety immediately, then each retry after will be adding 1 second, so with a max retries of 5 and max wait between intervals of 5--we get the initial request, then we retry immediately, then in 1 second we retry again.   You added the 5 second timeout for the unit test which makes sense relative to the ~1-2 second retry this could take for the policy, however I don't see where any kind of socket connect timeout is provided, does it account for time spent failing to make the network connection before the retry?    The OS is going to spend some amount of time trying to connect, at what point does it decide that it has failed before this retry policy should kick in?   I looked at the failure of the travis/app_very builds relating to this test and it does appear the mark_timeout is the exception that causes it to fail but further up is a socket timeout, it looks like its still trying to connect from the initial attempt.

danidee10: Yeah, your assumption is correct (I didn't consider the `OS`) but I've run the test multiple times with the "Retry policy" patch applied and it takes roughly 3 Seconds for `exceptions.OperationalError` to be raised.

`interval_max: 5` doesn't really apply because the connection to `localhost:8000` fails instantly. So celery doesn't need to wait for 5 seconds to retry the connection. It retries the connection almost instantly and it fails again. Since we have a `max_retries` setting of `3` it takes roughly 3 seconds for the entire process to happen.

I've also doubled the `interval_step` to 2 and it increased the pytest timeout and the time doubles to ~6 seconds. So I believe this proves that the behaviour is consistent (i.e `interval_max` doesn't apply because it's trying to connect to `localhost:8000`)

Finally `!pytest.mark.timeout` causes the test to fail because currently, the retry policy is not being applied and the default behaviour of celery is to retry forever. the timeout serves two purposes:

1. To prevent the test from running forever
2. To prove that the retry policy is not being applied because the exception should be raised before the timeout kills the test.

danidee10:
ping! @matteius :)

PR #1094, opened 2019-09-02
Issue 1093 retry policy testing

matteius:
Just made my own branch based on danidee10 (not sure why its not showing your name in the tag list) to change the pytest timeout value to see how it responds in the CI -- due to a default socket connect timeout of 4 seconds.


matteius:
Something is definitely odd here, I added a logging statement that I expected the test to execute on the first retry and I don't see it in the test output here:  https://travis-ci.org/celery/kombu/jobs/579628956

Also, it took 30 seconds and failed on the first sleep -- based on the retry policy it should have retried at least once, but also is it not respecting the default socket timeout of the connection?  Definitely some weird behavior in the unit test provided by @danidee10

PR #1099, opened 2019-09-10
fixes #1097

### LINKED ISSUE #1097: Incorrect handling in Connection.__init__() if hostname is a list starting with an invalid hostname
# How to reproduce:

* Create a `Connection` object, passing as the `hostname` parameter a list of non-empty strings where the first string does *not* contain '://'
* Call `as_uri()` on the connection object

# Expected result:

* An error is generated - preferably when the `Connection` is created - warning that the provided hostname is invalid because it does not specify a protocol

# Actual result:

The following error is generated:
```
File "/[REDACTED]/tools/python/lib/python3.6/urllib/parse.py", line 800, in quote_from_bytes
    raise TypeError("quote_from_bytes() expected bytes")
```

# Notes:

The issue is that the `hostname` parameter is permitted take several forms - a single hostname; a string containing a semicolon-separated list of hostnames; or an actual `list` of hostnames - but the `Connection.hostname` instance member is assumed to always be a single hostname string. `Connection.__init__()` does have handling to detect when it has been passed a list and extract the first value from it, but if this values *does not* contain a protocol, then the value of `params['hostname']` is never updated; when `self._init_params(**params)` is called, the original list value is thus assigned to `self.hostname`, resulting in a bizarre and inscrutable error message when `as_uri()` subsequently passes `self.hostname` to a function which expects a string or bytes object, not a list.

shekhar-kamble:


auvipy:
are you trying to fix this issue https://github.com/celery/kombu/issues/1097?

shekhar-kamble:
> are you trying to fix this issue #1097?

Yes. Also trying to fix the case when multiple urls are passed as `hostname` in the format of `example.com;_protocol://example.com`

auvipy:
next time you send a PR, always create a new branch against upstream/master by using the command: git checkout -b name-of-new-branch upstream/master

PR #1104, opened 2019-10-06
Added redis transport key_prefix from envvars

attiasr:


auvipy:
which issue you are trying to fix? can you add a test?

attiasr:
sorry for the late reply but I see you've already approved it, closed it and reopen so I'll explain;

This PRs will allow you to namespace different celery installation running against the same redis instance

matusvalo:
Is there any need of this patch? All key names  can be modified via transport options:

https://github.com/celery/kombu/blob/a042d57c575092ab131d2609d1d9523991d42e34/kombu/transport/redis.py#L477-L498

Moreover this PR is missing unittests and documentation. I prefer to revert this PR and rethink it if it is needed or not.

thedrow:
@matusvalo Can you take care of that?

matusvalo:
Yes if there is agreement on it, I will revert it.

matusvalo:
@attiasr we have reverted the change. If you still need prefixing functionality, we can discuss about it. Please create a new bug and we can drive discussion there.

Issue #1101, opened 2019-09-27
Error 'missing message for exchange `reply.celery.pidbox`' on ping with git package


Ashish-Bansal:
Hi there!

In attempt to try and fix https://github.com/celery/kombu/issues/1019, I tried installing kombu from master branch.

After installing git package, I tried ping the worker using `celery -A app inspect ping`, but in the celery logs it prints following message: 

```
[2019-09-27 14:23:13,969: DEBUG/MainProcess] pidbox received method ping() [reply_to:{'exchange': 'reply.celery.pidbox', 'routing_key': '002a0f42-a9a6-33a2-86b6-8587884d26ee'} ti
cket:4d480db3-effd-452d-a72f-b35be19043f4]
[2019-09-27 14:23:13,973: ERROR/MainProcess] Control command error: OperationalError("\nCannot route message for exchange 'reply.celery.pidbox': Table empty or key no longer exis
ts.\nProbably the key ('_kombu.binding.reply.celery.pidbox') has been removed from the Redis database.\n",)
Traceback (most recent call last):
  File "/home/mrphantom/foo/venv/src/kombu/kombu/connection.py", line 439, in _reraise_as_library_errors
    yield
  File "/home/mrphantom/foo/venv/src/kombu/kombu/connection.py", line 518, in _ensured
    return fun(*args, **kwargs)
  File "/home/mrphantom/foo/venv/src/kombu/kombu/messaging.py", line 203, in _publish
    mandatory=mandatory, immediate=immediate,
  File "/home/mrphantom/foo/venv/src/kombu/kombu/transport/virtual/base.py", line 605, in basic_publish
    message, exchange, routing_key, **kwargs
  File "/home/mrphantom/foo/venv/src/kombu/kombu/transport/virtual/exchange.py", line 70, in deliver
    for queue in _lookup(exchange, routing_key):
  File "/home/mrphantom/foo/venv/src/kombu/kombu/transport/redis.py", line 877, in _lookup
    exchange, redis_key))
kombu.exceptions.InconsistencyError:
Cannot route message for exchange 'reply.celery.pidbox': Table empty or key no longer exists.
Probably the key ('_kombu.binding.reply.celery.pidbox') has been removed from the Redis database.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/mrphantom/foo/venv/lib/python3.6/site-packages/celery/worker/pidbox.py", line 46, in on_message
    self.node.handle_message(body, message)
[...]

PR #1106, opened 2019-10-10
Revert "Issue #1087 redis fix"

auvipy:
Reverts celery/kombu#1089

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -3,12 +3,12 @@ from __future__ import absolute_import, unicode_literals
 
 import numbers
 import socket
-import warnings
 
 from bisect import bisect
 from collections import namedtuple
 from contextlib import contextmanager
 from time import time
+
 from vine import promise
 
 from kombu.exceptions import InconsistencyError, VersionMismatch
@@ -25,7 +25,6 @@ from kombu.utils.uuid import uuid
 from kombu.utils.compat import _detect_environment
 
 from . import virtual
-from .virtual.base import UndeliverableWarning, UNDELIVERABLE_FMT
 
 try:
     import redis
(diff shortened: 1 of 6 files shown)


thedrow:
Why are we reverting this?

auvipy:
this might have introduced a regression

thedrow:
How exactly?
This brings support for py-redis 2.x which we don't want.

littlebtc:
See #1101

auvipy:
I will handle pyredis 3.x in a separate PR. this doesn't seem to add pyresis 2.x though

kombu/transport/redis.py
```diff
@@ -147,8 +146,12 @@ def __init__(self, *args, **kwargs):
     def append(self, message, delivery_tag):
         delivery = message.delivery_info
         EX, RK = delivery['exchange'], delivery['routing_key']
-        # Redis-py changed the format of zadd args in v3.0.0 to be like this
-        zadd_args = [{delivery_tag: time()}]
+        # TODO: Remove this once we soley on Redis-py 3.0.0+
+        if redis.VERSION[0] >= 3:
```
thedrow: Yes it does...

auvipy: I will handle this in a pr after merging this

matteius:
@auvipy  This is rough news to see that my work was reverted -- I had spent quite a few long nights and hours on #1089 which you reverted.  All of the work to get the module to use fakeredis for the tests as well as fixing the other reports of #1087 and #1081 ... I don't understand how we let a different bug report cause us to revert back to a point that was known to have issues.   I want to help but when my work is just flat ripped out after it was accepted back in August makes it hard.

matteius:
I honestly think we are making a mistake, we did a bunch of work on Kombu to make it better with respect to some bug reports, there is a new report, we should have continued improving the code from that point, but then a number of revisions have been made to turn it it back closer into a much older release of Kombu -- without much discussion. 
image

I am going to diff where trunk is now to prior release or two, and also compare to where it was when I was working on it pre-October with respect to Redis and what the report of #1101 was, because it would have been nice to fix #1101 under the new framework that was also using fakeredis client in the tests.

PR #1110, opened 2019-10-18
Add wait timeout settings to receive queue message

nicolasmota:
This PR adds `wait_timeout_seconds` to be used to receive queue message when timeout isn't informed  and can be set using `broker_transport_options`

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -25,7 +25,7 @@ omit =
     *kombu/transport/zmq.py
     *kombu/transport/django.py
     *kombu/transport/pyro.py
-    *kombu/transport/azure*
+    *kombu/transport/azurestoragequeues.py
     *kombu/transport/qpid*
 exclude_lines =
     pragma: no cover
--- a/kombu/transport/azureservicebus.py
+++ b/kombu/transport/azureservicebus.py
@@ -49,7 +49,7 @@ class Channel(virtual.Channel):
     """Azure Service Bus channel."""
 
     default_visibility_timeout = 1800  # 30 minutes.
-    default_wait_time_seconds = 10  # 10 seconds
+    default_wait_time_seconds = 5  # in seconds
     domain_format = 'kombu%(vhost)s'
     _queue_service = None
     _queue_cache = {}
--- a/kombu/transport/azureservicebus.py
+++ b/kombu/transport/azureservicebus.py
@@ -49,6 +49,7 @@ class Channel(virtual.Channel):
     """Azure Service Bus channel."""
 
     default_visibility_timeout = 1800  # 30 minutes.
+    default_wait_time_seconds = 10  # 10 seconds
     domain_format = 'kombu%(vhost)s'
     _queue_service = None
     _queue_cache = {}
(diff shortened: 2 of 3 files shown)


kombu/transport/azureservicebus.py
```diff
@@ -92,7 +93,10 @@ def _put(self, queue, message, **kwargs):
     def _get(self, queue, timeout=None):
         """Try to retrieve a single message off ``queue``."""
         message = self.queue_service.receive_queue_message(
-            self.entity_name(queue), timeout=timeout, peek_lock=False)
+            self.entity_name(queue),
+            timeout=timeout or self.wait_time_seconds,
```
dukejeffrie: instead of the `or` clause, you could inline the usage and not cache anything:

```py
if timeout is None:
  timeout = self.transport_options.get('wait_time_seconds',
                                          self.default_wait_time_seconds)

**Comment by nicolasmota** [comment]
adding unittests

**Comment by auvipy** [comment]
check the other review comments as well

File: kombu/transport/azureservicebus.py (hunk 1 above)
Comment (nicolasmota) [reply kombu/transport/azureservicebus.py]: this attribute will be accessed many times, once per message, the cache gets some gain from it and other implementations follow this pattern

**Comment by nicolasmota** [comment]
@thedrow added test coverage

File: kombu/transport/azureservicebus.py [hunk 2]
```diff
@@ -49,6 +49,7 @@ class Channel(virtual.Channel):
     """Azure Service Bus channel."""
 
     default_visibility_timeout = 1800  # 30 minutes.
+    default_wait_time_seconds = 10  # 10 seconds
```
italux: Why not use a lower default timeout? Like 5 seconds?

.coveragerc
```diff
@@ -25,7 +25,7 @@ omit =
     *kombu/transport/zmq.py
     *kombu/transport/django.py
     *kombu/transport/pyro.py
-    *kombu/transport/azure*
+    *kombu/transport/azurestoragequeues.py
```
thedrow: Why did you exclude servicebus from test coverage?

nicolasmota: I removed it from the `omit` because it was not being generated his coverage report

t/unit/transport/test_azureservicebus.py
```diff
@@ -0,0 +1,232 @@
...
+        return self.queues
+
+    def send_queue_message(self, queue_name=None, message=None):
+        queue = self.get_queue(queue_name)
+        queue.messages.append(message)
+
+    def receive_queue_message(self, queue_name, peek_lock=True, timeout=60):
+        queue = self.get_queue(queue_name)
+        if queue:
+            try:
```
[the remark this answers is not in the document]
dukejeffrie: Also, shouldn't you remove it from `self._queue_cache`?

auvipy:
@dukejeffrie It's OK

t/unit/transport/test_azureservicebus.py
```diff
@@ -0,0 +1,232 @@
...
+                return queue.messages.pop(0)
+            except IndexError:
+                return Message()
+
+    def read_delete_queue_message(self, queue_name, timeout='60'):
```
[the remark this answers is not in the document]
nicolasmota: on sdk they implemented using string 😢 https://github.com/Azure/azure-sdk-for-python/blob/27bed2b6296f95e219cef89f46f1238ce3c6cf48/sdk/servicebus/azure-servicebus/azure/servicebus/control_client/servicebusservice.py#L999

PR #1128, opened 2019-11-21
Add missing parameter queue_args in kombu.connection.SimpleBuffer

mrodm:
When using SimpleBuffer from kombu.connection, exchange_opts are not filled with the parameters desired.

`queue_args` parameter was introduced in `kombu.simple.SimpleQueue` and at the same time in
`kombu.connection.SimpleQueue` method. However, `kombu.connection.SimpleBuffer`
was not updated in the same way as the other method.

As `kombu.connection.SimpleBuffer` is using positional arguments to create a
`kombu.simple.SimpleBuffer` object, the values of that instance will not have the expected values:
- `queue_args` would get the value of `exchange_opts` from kombu.connection.SimpleBuffer, because that variable is set the third one.
- `exchange_opts` would get the default value None since none was provided by the caller.

Here is the call from SimpleBuffer method and the SimpleQueue class definition:
kombu.connection.SimpleBuffer
```
    def SimpleBuffer(self, name, no_ack=None, queue_opts=None,
                     exchange_opts=None, channel=None, **kwargs):
        ...
        from .simple import SimpleBuffer
        return SimpleBuffer(channel or self, name, no_ack, queue_opts,
                            exchange_opts, **kwargs)
```

kombu.simple.SimpleBuffer
```
class SimpleQueue(SimpleBase):
    """Simple API for persistent queues."""

    ...
    def __init__(self, channel, name, no_ack=None, queue_opts=None,
                 queue_args=None, exchange_opts=None, serializer=None,
                 compression=None, **kwargs):

class SimpleBuffer(SimpleQueue):
    ...
```

In this Pull Request, it's added the same `queue_args` parameter followiing the same approach as in the method `kombu.connection.SimpleQueue`.

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -771,6 +771,7 @@ class Connection(object):
                            exchange_opts, **kwargs)
 
     def SimpleBuffer(self, name, no_ack=None, queue_opts=None,
+                     queue_args=None,
                      exchange_opts=None, channel=None, **kwargs):
         """Simple ephemeral queue API.
 
@@ -785,6 +786,7 @@ class Connection(object):
         """
         from .simple import SimpleBuffer
         return SimpleBuffer(channel or self, name, no_ack, queue_opts,
+                            queue_args,
                             exchange_opts, **kwargs)
 
     def _establish_connection(self):
--- a/t/unit/test_connection.py
+++ b/t/unit/test_connection.py
@@ -495,6 +495,37 @@ class test_Connection:
         q2 = conn.SimpleBuffer('foo', channel=chan)
         assert q2.channel is chan
 
+    def test_SimpleQueue_with_parameters(self):
+        conn = self.conn
+        q = conn.SimpleQueue(
+            'foo', True, {'durable': True}, {'x-queue-mode': 'lazy'},
+            {'durable': True, 'type': 'fanout', 'delivery_mode': 'persistent'})
+
+        assert q.queue.exchange.type == 'fanout'
+        assert q.queue.exchange.durable
+        assert not q.queue.exchange.auto_delete
+        delivery_mode_code = q.queue.exchange.PERSISTENT_DELIVERY_MODE
+        assert q.queue.exchange.delivery_mode == delivery_mode_code
+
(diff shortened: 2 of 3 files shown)


mrodm:
Sure! I'm working to have some tests about these two methods (SimpleBuffer and SimpleQueue) running successfully.

Just one question, what would you prefer for testing the methods in kombu.connection.py ? @auvipy 
- With one mock, and just ensure that kombu.simple.SimpleBuffer (or SimpleQueue) is called with the right parameters at the right positions:
```
    def test_SimpleBuffer_with_parameters(self):
        with patch('kombu.simple.SimpleBuffer') as simple_buffer:
            conn = self.conn
            conn.SimpleBuffer(
                'foo', True, {'durable': True}, {'x-queue-mode': 'lazy'},
                {'durable': True, 'type': 'fanout', 'no_declare': True})
            simple_buffer.assert_called_with(
                self.conn, 'foo', True, {'durable': True, },
                {'x-queue-mode': 'lazy'},
                {'durable': True, 'type': 'fanout', 'no_declare': True}
            )
```
- Or call directly the method, and assert all the parameters of the `q` variable:
```
    def test_SimpleBuffer_with_parameters(self):
        conn = self.conn
        q = conn.SimpleBuffer(
            'foo', True, {'durable': True}, {'x-queue-mode': 'lazy'},
            {'durable': True, 'type': 'fanout', 'delivery_mode': 'persistent'})
        assert q.queue.exchange.type == 'fanout'
        assert q.queue.exchange.durable
        assert q.queue.exchange.auto_delete
        delivery_mode_code = q.queue.exchange.PERSISTENT_DELIVERY_MODE
        assert q.queue.exchange.delivery_mode == delivery_mode_code
        assert q.queue.queue_arguments['x-queue-mode'] == 'lazy'
        assert q.queue.durable
        assert q.queue.auto_delete
```

mrodm:
I've added the tests following the second approach I've mentioned above. 
In any case, I can change it to the other approach (using Mock instead) if you consider it better.

Added tests for:
- SimpleBuffer and simpleQueue classes (kombu.simple module)
- SimpleBuffer and SimpleQueue methods in Connection class (kombu.connection module)

PR #1132, opened 2019-12-18
Port of redis code improvements from prior revision

matteius:
This brings back the naming convention refactor and the test cases to make use of the fakeredis testing module which allowed me to introduce some new tests and cleanup how imports were handled in the test file.   This was lost when my prior "redis fix" was reverted:  https://github.com/celery/kombu/pull/1106/files   

The changes here are more minimal and change not how the redis module functions, but changes it to be more readable and to update the test class to be more complete.

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -816,9 +816,8 @@ class Channel(virtual.Channel):
                     queue_name = self._queue_for_priority(queue, pri)
                     pipe = pipe.llen(queue_name)
                 sizes = pipe.execute()
-                size = sum(s for s in sizes
+                return sum(s for s in sizes
                            if isinstance(s, numbers.Integral))
-                return size
 
     def _queue_for_priority(self, queue, pri):
         queue_priority = self.priority(pri)
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -766,7 +766,8 @@ class Channel(virtual.Channel):
         queues = self._queue_cycle.consume(len(self.active_queues))
         if not queues:
             return
-        keys = [self._q_for_pri(queue, pri) for pri in self.priority_steps
+        _q_for_pri = self._queue_for_priority
+        keys = [_q_for_pri(queue, pri) for pri in self.priority_steps
                 for queue in queues] + [timeout or 0]
         self._in_poll = self.client.connection
         self.client.connection.send_command('BRPOP', *keys)
@@ -802,7 +803,8 @@ class Channel(virtual.Channel):
     def _get(self, queue):
         with self.conn_or_acquire() as client:
             for pri in self.priority_steps:
-                item = client.rpop(self._q_for_pri(queue, pri))
+                queue_name = self._queue_for_priority(queue, pri)
+                item = client.rpop(queue_name)
                 if item:
                     return loads(bytes_to_str(item))
             raise Empty()
@@ -811,12 +813,14 @@ class Channel(virtual.Channel):
         with self.conn_or_acquire() as client:
             with client.pipeline() as pipe:
                 for pri in self.priority_steps:
-                    pipe = pipe.llen(self._q_for_pri(queue, pri))
+                    queue_name = self._queue_for_priority(queue, pri)
+                    pipe = pipe.llen(queue_name)
                 sizes = pipe.execute()
-                return sum(size for size in sizes
+                size = sum(size for size in sizes
                            if isinstance(size, numbers.Integral))
+                return size
 
-    def _q_for_pri(self, queue, pri):
+    def _queue_for_priority(self, queue, pri):
         pri = self.priority(pri)
         if pri:
             return f"{queue}{self.sep}{pri}"
@@ -831,7 +835,7 @@ class Channel(virtual.Channel):
         pri = self._get_message_priority(message, reverse=False)
 
         with self.conn_or_acquire() as client:
-            client.lpush(self._q_for_pri(queue, pri), dumps(message))
+            client.lpush(self._queue_for_priority(queue, pri), dumps(message))
 
     def _put_fanout(self, exchange, message, routing_key, **kwargs):
         """Deliver fanout message."""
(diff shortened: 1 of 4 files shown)


matteius:
@auvipy  It should not break master -- there are no merge conflicts and the travis-ci passes.  It is not functional changes but rather refactoring.

auvipy:
I am talking about celery master, not kombu

matteius:
@auvipy Does not break celery master, that code base does not reference `_q_for_pri` which is a private kombu method.  Otherwise the changes were isolated to code kombu code readability and redis unit testing enhancements

kombu/transport/redis.py
```diff
@@ -768,14 +770,21 @@ def _size(self, queue):
...
                            if isinstance(size, numbers.Integral))
+                return size
 
-    def _q_for_pri(self, queue, pri):
+    def _queue_for_priority(self, queue, pri):
         pri = self.priority(pri)
-        return '%s%s%s' % ((queue, self.sep, pri) if pri else (queue, '', ''))
+        if pri:
+            queue_args = (queue, self.sep, pri)
+        else:
```
drsantos20: I would suggest changing the var name `pri` from line 781 to other name, that's because you're defining a new var with the same name from your parameter

kombu/transport/redis.py
```diff
@@ -768,14 +770,21 @@ def _size(self, queue):
         with self.conn_or_acquire() as client:
             with client.pipeline() as pipe:
                 for pri in self.priority_steps:
-                    pipe = pipe.llen(self._q_for_pri(queue, pri))
+                    queue_name = self._queue_for_priority(queue, pri)
+                    pipe = pipe.llen(queue_name)
                 sizes = pipe.execute()
-                return sum(size for size in sizes
+                size = sum(size for size in sizes
```
drsantos20: Could you clear these names?  `size` `sizes`

kombu/transport/redis.py
matteius: Good idea, I'll update it to be `priority`.

matteius:
@auvipy This one has been updated for review again.

matteius:
I spent days on this before this summer and when the original PR was reverted I was heart broken so I thought to make the best of it I would capture the testing bits and other variable naming improvements that is not functional code changes, and so to see it sit here and linger and be not considered important is again heart breaking.  Especially when on the other half of my PRs that have no testing the common ask is for testing, but we need to enhance the Redis tests and that is exactly what this PR does!

thedrow:
LGTM.

t/unit/transport/test_redis.py
```diff
@@ -226,42 +444,97 @@ def _get_response_error(self):
...
+        super(FakeRedisKombuTransportLite, self).__init__(*args, **kwargs)
 
     def _get_errors(self):
         return ((KeyError,), (IndexError,))
 
 
+class FakeRedisKombuTransport(FakeRedisKombuTransportLite):
+    Channel = FakeRedisKombuChannel
+
+
```
shaib: So, without getting into too much detail -- the above is nearly 500 lines of adaptations and setting up clients on top of FakeRedis.
1) If you're already refactoring, I'd take these out to their own file. Maybe more than one.
2) Why is all this necessary? there seem to be no comments explaining it, and coming in knowing nothing, I'd basically expect a FakeRedis-based test-suite to just set up FakeRedis as the Redis server, and then run everything, as is, against that -- 10-20 lines of setup before the actual tests begin, not 500.

I note that about half of these lines are pre-existing, but IIUC that was before FakeRedis -- so the test basically included "a small FakeRedis" inside it. Now that an actual FakeRedis is used, I'd expect these to go away. Or, otherwise, explain the need for them.

t/unit/transport/test_redis.py
```diff
@@ -226,42 +444,97 @@ def _get_response_error(self):
...
-    def setup(self):
+    def setUp(self):
         self.connection = self.create_connection()
         self.channel = self.connection.default_channel
 
+    def tearDown(self):
+        self.connection = None
+        self.channel = None
+        global _fake_redis_client
+        _fake_redis_client = None
```
shaib: Why is `tearDown()` tearing down something `setUp()` did not set up?

t/unit/transport/test_redis.py
```diff
@@ -755,48 +1037,38 @@ def test_transport_on_readable(self):
...
             assert errors
-            assert ID in errors[1]
-        finally:
-            if DataError is not None:
-                exceptions.DataError = DataError
-            if InvalidData is not None:
-                exceptions.InvalidData = InvalidData
+            assert redis.exceptions.DataError in errors[1]
+        except Exception:
+            raise
```
shaib: Why have a try-block, then?

t/unit/transport/test_redis.py
```diff
@@ -845,88 +1117,49 @@ def test_rediss_connection(self):
...
                 )
 
     def test_sep_transport_option(self):
-        with Connection(transport=Transport, transport_options={
-            'sep': ':',
-        }) as conn:
+        conn_kwargs = dict(
+            transport=FakeRedisKombuTransport,
+            transport_options={'sep': ':'})
+        with Connection(**conn_kwargs) as conn:
```
shaib: Suggestion:
```python
        with Connection(
            transport=FakeRedisKombuTransport,
            transport_options={'sep': ':'},
        ) as conn:
```

t/unit/transport/test_redis.py
```diff
@@ -1,22 +1,35 @@
...
 
 
+_fake_redis_client = None
+
+
+def _get_fake_redis_client():
+    global _fake_redis_client
+    if _fake_redis_client is None:
+        _fake_redis_client = FakeRedisClient()
+    return _fake_redis_client
```
shaib: While this implies you'd want to use the fake redis client as a singleton -- set up once and then used forever -- in actual use in the test suite, you take care to destroy the global in `tearDown()` functions, to force it to be set up anew later. This seems wrong -- it looks like you should be using normal objects, perhaps setting the client as an attribute of the TestCase instance, rather than try to make it a singleton fetched through a function and then "reach behind the function's back" to manipulate the global.

matusvalo:
Thank you @matteius. I am sorry for long delay :-).

auvipy:
> Thank you @matteius. I am sorry for long delay :-).

thanks both of you for bringing & drawing the finishing line.

matusvalo:
I had to revert this PR due failing integration tests. I prefer reopen this PR (not sure how can be done) and fix the problems with failing integration tests.

thedrow:
Revert the revert and edit the title.

PR #1133, opened 2019-12-18
Issue 1093 retry policy testing

matteius:
Simple refactor to the loop variable name to be more clear it is which retry attempt it is (attempt 1 is the initial attempt)  retries before was less clear.  The new test is supposed to exercise this code path for retries.

### DIFF
--- a/kombu/utils/functional.py
+++ b/kombu/utils/functional.py
@@ -29,6 +29,7 @@ __all__ = (
     'is_list', 'maybe_list', 'dictfilter',
 )
 
+
 KEYWORD_MARK = object()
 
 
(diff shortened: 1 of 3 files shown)


kombu/utils/functional.py
```diff
@@ -339,17 +340,17 @@ def retry_over_time(fun, catch, args=None, kwargs=None, errback=None,
                              interval_max + interval_start,
                              interval_step, repeatlast=True)
     end = time() + timeout if timeout else None
-    for retries in count():
+    for retry_attempt in count():
```
auvipy: is this renaming must be needed?

matteius: I felt retries is not a clear name -- it requires more code inference to see what is happening when reading through things quickly.

auvipy:
OK, but builds are not passing.

kombu/utils/functional.py
matteius: Well I think both names are not ideal because the first attempt is not a retry -- its the initial request.   Maybe it should be `for request_attempt` for code clarity.

matteius:
@auvipy Ah right, this one is related to https://github.com/celery/kombu/issues/1024 and its a unit test that @danidee10 originally was working with TDD that shows the retry policy issue -- the test fails demonstrating the issue 1024

matteius:
I spent some time debugging this test locally, it is getting stuck on the first retry here in messaging _publish on `message = channel.prepare_message` https://github.com/celery/kombu/blob/master/kombu/messaging.py#L188 

It never tests the retry policy because it hangs the full 30 seconds on this code path.

PR #1139, opened 2019-12-21
Support standard values for ssl_cert_reqs query parameter.

jdavisp3:
@auvipy 

Follow up to https://github.com/celery/celery/pull/5703 in which we teach `kombu` to accept the standard `redis-py` values for the `ssl_cert_reqs` query parameter for secure `rediss://` URLs.

Tested by adding unit tests and using this version of `kombu` and `celery` 4.4.0 to connect to an AWS ElastiCache instance of Redis configured to use SSL and then launching Celery tasks successfully.

### DIFF
--- a/kombu/utils/url.py
+++ b/kombu/utils/url.py
@@ -49,10 +49,8 @@ def parse_url(url):
         keys = [key for key in query.keys() if key.startswith('ssl_')]
         for key in keys:
             if key == 'ssl_cert_reqs':
-                if ssl_available:
-                    query[key] = getattr(ssl, query[key])
-                else:
-                    query[key] = None
+                query[key] = parse_ssl_cert_reqs(query[key])
+                if query[key] is None:
                     logger.warning('Defaulting to insecure SSL behaviour.')
 
             if 'ssl' not in query:
@@ -120,3 +118,20 @@ def maybe_sanitize_url(url, mask='**'):
     if isinstance(url, string_t) and '://' in url:
         return sanitize_url(url, mask)
     return url
+
+
+def parse_ssl_cert_reqs(query_value):
+    # type: (str) -> Any
+    """Given the query parameter for ssl_cert_reqs, return the SSL constant or None."""
+    if ssl_available:
+        query_value_to_constant = {
+            'CERT_REQUIRED': ssl.CERT_REQUIRED,
+            'CERT_OPTIONAL': ssl.CERT_OPTIONAL,
+            'CERT_NONE': ssl.CERT_NONE,
+            'required': ssl.CERT_REQUIRED,
+            'optional': ssl.CERT_OPTIONAL,
+            'none': ssl.CERT_NONE,
+        }
+        return query_value_to_constant[query_value]
+    else:
+        return None
--- a/t/unit/utils/test_url.py
+++ b/t/unit/utils/test_url.py
@@ -11,7 +11,7 @@ import ssl
 import pytest
 
 import kombu.utils.url
-from kombu.utils.url import as_url, parse_url, maybe_sanitize_url
+from kombu.utils.url import as_url, parse_url, maybe_sanitize_url, parse_ssl_cert_reqs
 
 
 def test_parse_url():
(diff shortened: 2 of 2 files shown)


jdavisp3:
Build failure seems spurious.

kombu/utils/url.py
```diff
@@ -120,3 +118,20 @@ def maybe_sanitize_url(url, mask='**'):
...
+    # type: (str) -> Any
+    """Given the query parameter for ssl_cert_reqs, return the SSL constant or None."""
+    if ssl_available:
+        query_value_to_constant = {
+            'CERT_REQUIRED': ssl.CERT_REQUIRED,
+            'CERT_OPTIONAL': ssl.CERT_OPTIONAL,
+            'CERT_NONE': ssl.CERT_NONE,
+            'required': ssl.CERT_REQUIRED,
+            'optional': ssl.CERT_OPTIONAL,
+            'none': ssl.CERT_NONE,
```
jtratner: doesn't this break earlier behavior if someone had specified ssl.CERT_REQUIRED?

Seems like you should have also add:

```
            'CERT_REQUIRED': ssl.CERT_REQUIRED,
            ssl.CERT_REQUIRED: ssl.CERT_REQUIRED,
            'CERT_OPTIONAL': ssl.CERT_OPTIONAL,
           ssl.CERT_OPTIONAL: ssl.CERT_OPTIONAL,...
            'CERT_NONE': ssl.CERT_NONE,
           ssl.CERT_NONE: ssl.CERT_NONE,
            'required': ssl.CERT_REQUIRED,
            'optional': ssl.CERT_OPTIONAL,
            'none': ssl.CERT_NONE,
```

jtratner: never mind - misread above

auvipy:
restarted

matteius:
@auvipy Please restart this build.  It stalled getting an apt-get dependency for the last Travis CI run.

While the change initially looked good and improves code readability for that feature by documenting which ssl attributes are supported.  The fact it builds up this dictionary and throws it away for each iteration of the loop is not an efficient way to do this and requires more memory.   

It seems to be the intent is to validate the ssl attribute being requested is one of the allowed attributes, so if this is a validation step we should just maintain a list of valid attributes and check that the attribute requested is allowed and continue using getattr for performance reasons.

kombu/utils/url.py
```diff
@@ -120,3 +118,20 @@ def maybe_sanitize_url(url, mask='**'):
     if isinstance(url, string_t) and '://' in url:
         return sanitize_url(url, mask)
     return url
+
+
+def parse_ssl_cert_reqs(query_value):
+    # type: (str) -> Any
+    """Given the query parameter for ssl_cert_reqs, return the SSL constant or None."""
+    if ssl_available:
+        query_value_to_constant = {
```
georgepsarakis: You can specify this mapping as a constant, e.g. `SSL_CERT_REQS`. Although there is minimal performance benefit, as this code runs only initialization, a new dictionary will not have to be created each time.

matteius: Well this is being called within a loop so if the method were reached for multiple keys then it would generate multiple times on initalization.  However upon closer inspection above, it only gets into this section if key is key == 'ssl_cert_reqs' and so then then that is the only key that would ever operate on `query[key] = parse_ssl_cert_reqs(query[key])` ==>  `query['ssl_cert_reqs'] = parse_ssl_cert_reqs(query['ssl_cert_reqs'])`.

So then query of ssl_cert_reqs will only ever be one of the dictionary values per iteration-- this is why I know generating a dictionary each time is not ideal because we are doing this in some kind if iterative calling pattern.   Plus any part of the eventloop that restarts or reinitialize would happen again.  

Anyway, Happy Holidays!!!

jdavisp3: Happy to refactor, but I'm skeptical of the benefit here. Using a constant means the dict will stick around in memory after initialization even though it's never going to be needed again. The analogous PR for `celery` (https://github.com/celery/celery/pull/5703) also uses an inline dict.

georgepsarakis: @jdavisp3 you are right! However, in terms of semantics this is a constant mapping, so this is what I meant and I should probably have explained it better, sorry for that. I thought that by converting to a constant, there are multiple gains (minor performance gain and slightly more understandable code for future readers). Either way this is not so important, but since this is a quite large and complex codebase, it is sometimes better to pay closer attention to these details.

PR #1140, opened 2019-12-23
add accept parameter to SimpleQueue class

lsaavedr:
please add 'accept' parameter to SimpleQueue class

### DIFF
--- a/kombu/simple.py
+++ b/kombu/simple.py
@@ -118,7 +118,7 @@ class SimpleQueue(SimpleBase):
 
     def __init__(self, channel, name, no_ack=None, queue_opts=None,
                  queue_args=None, exchange_opts=None, serializer=None,
-                 compression=None, **kwargs):
+                 compression=None, accept=None, **kwargs):
         queue = name
         queue_opts = dict(self.queue_opts, **queue_opts or {})
         queue_args = dict(self.queue_args, **queue_args or {})
@@ -134,7 +134,7 @@ class SimpleQueue(SimpleBase):
         else:
             exchange = queue.exchange
             routing_key = queue.routing_key
-        consumer = messaging.Consumer(channel, queue)
+        consumer = messaging.Consumer(channel, queue, accept=accept)
         producer = messaging.Producer(channel, exchange,
                                       serializer=serializer,
                                       routing_key=routing_key,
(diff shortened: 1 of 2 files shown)


matusvalo:
Hi @drsantos20. You can start adding tests here: https://github.com/celery/kombu/blob/60dc3456368a06ed4125ea9eac9c06e7fcf5cffe/t/unit/test_simple.py

I would put it to Both test_SimpleQueue and test_SimpleBuffer classes.

matusvalo:
I have added unittests and fixed slightly the PR. Please review. @celery/core-developers

kombu/simple.py
```diff
@@ -134,13 +135,13 @@ def __init__(self, channel, name, no_ack=None, queue_opts=None,
...
             routing_key = queue.routing_key
-        consumer = messaging.Consumer(channel, queue)
+        consumer = messaging.Consumer(channel, queue, accept=accept)
         producer = messaging.Producer(channel, exchange,
                                       serializer=serializer,
                                       routing_key=routing_key,
                                       compression=compression)
         super().__init__(channel, producer,
-                         consumer, no_ack, **kwargs)
+                         consumer, no_ack, accept)
```
maybe-sybr, line 144: Why has this lost the `**kwargs` passthrough? We still accept them in the method signature so I assume we still need to pass them along to be cooperative.

kombu/simple.py
```diff
@@ -67,7 +68,7 @@ def get(self, block=True, timeout=None):
                 remaining = timeout - elapsed
 
     def get_nowait(self):
-        m = self.queue.get(no_ack=self.no_ack)
+        m = self.queue.get(no_ack=self.no_ack, accept=self.accept)
```
maybe-sybr, line 71: Could we avoid this by simply using the `self.consumer.accept` attribute? Then we wouldn't need the new kwarg for this class.

maybe-sybr (changes requested):
A few minor changes which I think would make this better. I'd also like to see a test for what happens if we pass in an empty list/set for the `accept` kwarg.

kombu/simple.py
matusvalo: Yes I think it is reasonable.

matusvalo:
> I'd also like to see a test for what happens if we pass in an empty list/set for the accept kwarg.
done.

@maybe-sybr I have implemented your comments.

maybe-sybr (approved):
lgtm, thanks @matusvalo

maybe-sybr:
We haven't done an RC for kombu like we have with celery, so this could arguably land but I'll leave that to @thedrow since he's managing the releases.

Issue #701, opened 2017-02-28
kombu.transport.redis.Mutex is broken in python 3


zaro:
The Mutex class is defined as:
```python
@contextmanager
def Mutex(client, name, expire):
    """The Redis lock implementation (probably shaky)."""
    lock_id = uuid()
    i_won = client.setnx(name, lock_id)
    try:
        if i_won:
            client.expire(name, expire)
            yield
        else:
            if not client.ttl(name):
                client.expire(name, expire)
            raise MutexHeld()
    finally:
        if i_won:
            try:
                with client.pipeline(True) as pipe:
                    pipe.watch(name)
                    if pipe.get(name) == lock_id:
                        pipe.multi()
                        pipe.delete(name)
                        pipe.execute()
                    pipe.unwatch()
            except redis.WatchError:
                pass
```

The following cleanup condition:

``` python
                    if pipe.get(name) == lock_id:
```

will always fail under python 3  because lock_id is string, and the default behaviour for redis is to return bytes in python 3. This leads to the Mutex never being cleaned up. 

Changing it to :

```python
                    if pipe.get(name).decode() == lock_id:
```
fixes the problem.

auvipy:
any failing test and proposed fix as PR?

zaro:
@auvipy 
OK :) Got some time to figure out how to run tests. The tests use a mock Redis client, which will never show this behaviour, as the data doesn't travel over the network to Redis.

Is there any way to run the tests with actual Redis instance ?

I can make a PR with the fix, no problem, but I have no clue how to add a test case :)

georgepsarakis:
@zaro I think you need to use the [bytes_if_py2](https://github.com/celery/vine/blob/master/vine/five.py#L202) compatibility function:

```
lock_id = bytes_if_py2(uuid())
```

shaib:
Hi, this was relatively recently closed without explanation, but the problem is still there in Kombu 4.6.7. We run tests which include using an actual Redis rather than a mock, and when we run them with the Python flag `-bb` (make mixing bytes and string an error), we get an exception (context doesn't matter much, but our actual use case is through Celery):
```
  File ".../lib/python3.7/site-packages/kombu/transport/redis.py", line 127, in Mutex
    if pipe.get(name) == lock_id:
BytesWarning: Comparison between bytes and string
```
I am going to work on a fix.

PR #1141, opened 2019-12-30
Issue 701

shaib:
Hi,

I include here two commits that handle issue #701 which was closed a few months ago but is in fact still very much present.

This PR includes:
- A commit that changes the tests so that the relevant Redis-mocking operations are more true to the behavior of modern py-redis; and adds the Python `-bb` flag to the test run, for stricter checking of str-vs-bytes issues
- A commit which fixes the operational code to resolve the problem
Thanks, Shai.

(This PR courtesy of [Matific](https://www.matific.com/))

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -124,7 +124,7 @@ def Mutex(client, name, expire):
             try:
                 with client.pipeline(True) as pipe:
                     pipe.watch(name)
-                    if pipe.get(name) == lock_id:
+                    if bytes_to_str(pipe.get(name)) == lock_id:
                         pipe.multi()
                         pipe.delete(name)
                         pipe.execute()
--- a/t/unit/transport/test_redis.py
+++ b/t/unit/transport/test_redis.py
@@ -14,6 +14,7 @@ from kombu.exceptions import InconsistencyError, VersionMismatch
 from kombu.five import Empty, Queue as _Queue, bytes_if_py2
 from kombu.transport import virtual
 from kombu.utils import eventio  # patch poll
+from kombu.utils.encoding import str_to_bytes
 from kombu.utils.json import dumps
 
 
@@ -1311,13 +1312,13 @@ class test_Mutex:
             client.setnx.return_value = True
             client.pipeline = ContextMock()
             pipe = client.pipeline.return_value
-            pipe.get.return_value = lock_id
+            pipe.get.return_value = str_to_bytes(lock_id)  # redis gives bytes
             held = False
             with redis.Mutex(client, 'foo1', 100):
                 held = True
             assert held
             client.setnx.assert_called_with('foo1', lock_id)
-            pipe.get.return_value = 'yyy'
+            pipe.get.return_value = b'yyy'
             held = False
             with redis.Mutex(client, 'foo1', 100):
                 held = True
@@ -1325,7 +1326,7 @@ class test_Mutex:
 
             # Did not win
             client.expire.reset_mock()
-            pipe.get.return_value = lock_id
+            pipe.get.return_value = str_to_bytes(lock_id)
             client.setnx.return_value = False
             with pytest.raises(redis.MutexHeld):
                 held = False
--- a/tox.ini
+++ b/tox.ini
@@ -28,7 +28,7 @@ deps=
     flake8,flakeplus,pydocstyle: -r{toxinidir}/requirements/pkgutils.txt
 
 commands = pip install -U -r{toxinidir}/requirements/dev.txt
-           pytest -rxs -xv --cov=kombu --cov-report=xml --no-cov-on-fail {posargs}
+           python -bb -m pytest -rxs -xv --cov=kombu --cov-report=xml --no-cov-on-fail {posargs}
 
 basepython =
     2.7,flakeplus,flake8,linkcheck,cov: python2.7


auvipy:
thanks for the PR Shai!

requirements/extras/lzma.txt
```diff
@@ -1,2 +1,2 @@
---extra-index-url https://pypi.anaconda.org/nehaljwani/simple # Temporary index for Windows wheels
+# --extra-index-url https://pypi.anaconda.org/nehaljwani/simple # Temporary index for Windows wheels
```
matteius: If this is not needed anymore, maybe we should just remove the line?

shaib: I don't know that it isn't needed; I don't have access to a Windows machine. All I know is that it doesn't seem to work and it breaks the build on my Linux machine.

kombu/transport/redis.py
```diff
@@ -124,7 +124,8 @@ def Mutex(client, name, expire):
             try:
                 with client.pipeline(True) as pipe:
                     pipe.watch(name)
-                    if pipe.get(name) == lock_id:
+                    pipename = pipe.get(name)
+                    if pipename and pipename.decode() == lock_id:
```
georgepsarakis: Kombu has already some utility functions for these common use cases. I think here we could use [bytes_to_str](https://github.com/celery/kombu/blob/fe5adb539ce9b351ffe3df0719ca281a736b0014/kombu/utils/encoding.py#L53).

requirements/extras/lzma.txt
georgepsarakis: @shaib if Travis CI passes without this change, then I think it should be the subject of a different PR. What do you think?

t/unit/transport/test_redis.py
```diff
@@ -1311,21 +1311,21 @@ def test_mutex(self, lock_id='xxx'):
             client.setnx.return_value = True
             client.pipeline = ContextMock()
             pipe = client.pipeline.return_value
-            pipe.get.return_value = lock_id
+            pipe.get.return_value = lock_id.encode()  # redis returns bytes
```
georgepsarakis: I guess that this happens on Python3, right? Perhaps we can use again `bytes_to_str`, so that we are always consistent.

tox.ini
```diff
@@ -28,7 +28,7 @@ deps=
     flake8,flakeplus,pydocstyle: -r{toxinidir}/requirements/pkgutils.txt
 
 commands = pip install -U -r{toxinidir}/requirements/dev.txt
-           pytest -rxs -xv --cov=kombu --cov-report=xml --no-cov-on-fail {posargs}
+           python -bb -m pytest -rxs -xv --cov=kombu --cov-report=xml --no-cov-on-fail {posargs}
```
georgepsarakis: So the `-bb` flag produces the bytes-str comparison warnings?

requirements/extras/lzma.txt
shaib: I'm not sure this change is necessary at all. When I tried at the office earlier, I consistently got a 404 for the index page, so I figured the URL was no longer valid. I just tried from home and it seems to be responding with content, so... this change is a separate commit, which should probably be dropped.

kombu/transport/redis.py
shaib: Good idea

t/unit/transport/test_redis.py
shaib: Yes, except that this one is `str_to_bytes`.

tox.ini
shaib: `-b` produces warnings. `-bb` turns these warnings into errors. Without these flags the comparison "passes silently" but under Python 3 never returns `True`.

requirements/extras/lzma.txt
matteius: The index loads for me now -- it appears it is a set of wheels packages for backports.lzma whereas no wheels files exist in pypi for backports.lzma.   We will want to maintain this line -- perhaps your office DNS was not responding to the index or it was temporarily down.   We require backports.lzma for python < 3.3

shaib: Change removed.

shaib:
PR Updated according to review comments. The package index seems to be working now, so I removed the change related to it from the PR.

shaib:
While at it, I noted that the file `t/unit/transport/test_redis.py` has a whole mock Redis client implementation, which probably has similar issues of returning `str`s where it should return `bytes` on Python 3. I haven't looked deeply into it, and the test modified here doesn't use it at all (it sets up its own mocks) -- but there may be more bugs hidden there. It should probably be a separate issue, though.

shaib:
It isn't clear to me why the AppVeyor builder still tries to compile its own wheel rather than get it from the index.

matteius:
@shaib Could you look at my refactor of the redis test file over here (I introduced fakeredis package):  https://github.com/celery/kombu/pull/1132/files

You'll have to load the diff because it was a lot of changes, and I added a couple tests too.

matteius:
I think we will want both sets of changes ultimately.

shaib:
> @shaib Could you look at my refactor of the redis test file over here (I introduced fakeredis package): https://github.com/celery/kombu/pull/1132/files
> 

I am not familiar with `fakeredis` nor, really, with the Kombu redis tests, so it will take me some effort. I'll try to take a look over the weekend.

PR #1142, opened 2020-01-01
Used a context manager for locks and files in filesystem transport.

atombrella:


atombrella:
It seems that the filesystem transport is not tested under Windows platforms. `test_filesystem.py` is decorated with `@skip.if_win32()`. The use of a context manager for the files should not require additional tests; it may be different with the locks.

auvipy:
I think we should remove win32 bit support

atombrella:
`@skip.if_win32()` doesn't have anything to do with either 32 or 64 bit. If you see the output from AppVeyor, the two tests from `test_filesystem` are skipped on all Windows platforms, e.g., also with `PYTHON_ARCH=64`.

matteius:
I am hoping to keep supporting windows build related to issues -- I finally got pycurl to build locally on my Windows ecosystem -- libcurl, and pycurl building solidly now.  I agree that the 32 here on the pytest decorator does not signify 32 bit support, its just old nomenclature.

auvipy (changes requested):
would you mind adding some unit tests & doc/release note entry for the changes? And add your name to author/contributor list if not already?

atombrella:
> would you mind adding some unit tests & doc/release note entry for the changes? And add your name to author/contributor list if not already?

In any case, the doc-string needs to be added, since there's a failing check for this. I'll see what I can do about unit tests; my changes are mostly updates to newer Python syntax/constructs.

auvipy (changes requested):
just check the lint errors please and some little adjustment to unit tests would be alright

atombrella:
I had completely forgotten about this, and don't have the time to finish it.

Issue #1114, opened 2019-10-28
Support health_check_interval option in broker_transport_options


kmctown:
As a follow-up to #1019 and #1113, update the [hard-coded `health_check_interval`](https://github.com/celery/kombu/blob/master/kombu/transport/redis.py#L1105) to be configurable via `broker_transport_options`.

It also seems that we will want the default to be `25` instead of `30`.

cc: @thedrow

jdavisp3:
Would having the current default be 30 cause the problem to sometimes continue? Still seeing the "error 32" bug on 2d92827746f0f2e5c964d685246836370f14786b.

KFoxder:
@jdavisp3 
> Would having the current default be 30 cause the problem to sometimes continue? Still seeing the "error 32" bug

Per @kmctown 's [comment](https://github.com/celery/kombu/pull/1113#issuecomment-546942030) the number of broken pipes (error 32) should be greatly reduced but not completely eliminated. This issue here, once resolved, should allow us to configure it per our setup. 

> Have an update: it seems to have mostly fixed it but I am still getting some occasional broken pipes, so we are not out of the woods yet. It is greatly reduced, though, at least on my workload. I think the 30 timeout might be too high, and we might want 20-25 per @andymccurdy recommendation over on redis-py. Seems like the best solution is to make the configurable `broker_transport_options` update to control `health_check_interval` and reduce the default to `20` or `25`.

_Originally posted by @kmctown in https://github.com/celery/kombu/pull/1113#issuecomment-546942030_

auvipy:
can we consider this as a blcker ? @thedrow

thedrow:
Yes, this issue is a blocker.

auvipy:
https://github.com/celery/kombu/pull/1122 ?

KFoxder:
@auvipy Just to make sure I understand. It seems that https://github.com/celery/kombu/pull/1122 fixed this:
> It also seems that we will want the default to be 25 instead of 30.

But it didn't allow for custom set `health_check_interval` in the `broker_transport_options`. Is that correct?

KFoxder:
@thedrow @kmctown I took a stab at this. Let me know if I am going down the right path or not.

Ashish-Bansal:
> But it didn't allow for custom set health_check_interval in the broker_transport_options. Is that correct?

Yes, that's right.

I saw https://github.com/celery/kombu/pull/1145 got merged. May we can close this issue now.

PR #1145, opened 2020-01-04
Add support for health_check_interval option in broker_transport_options.

KFoxder:
Attempt to address https://github.com/celery/kombu/issues/1114. Added the ability to specify `health_check_interval` in `broker_transport_options`.

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -349,9 +349,9 @@ class MultiChannelPoller(object):
         for channel in self._channels:
             # only if subclient property is cached
             client = channel.__dict__.get('subclient')
-            if client is not None:
-                if callable(getattr(client, "check_health", None)):
-                    client.check_health()
+            if client is not None \
+                    and callable(getattr(client, 'check_health', None)):
+                client.check_health()
 
     def on_readable(self, fileno):
         chan, type = self._fd_to_chan[fileno]
@@ -428,6 +428,7 @@ class Channel(virtual.Channel):
     socket_keepalive = None
     socket_keepalive_options = None
     max_connections = 10
+    health_check_interval = DEFAULT_HEALTH_CHECK_INTERVAL
     #: Transport option to disable fanout keyprefix.
     #: Can also be string, in which case it changes the default
     #: prefix ('/{db}.') into to something else.  The prefix must
@@ -491,14 +492,14 @@ class Channel(virtual.Channel):
          'socket_keepalive_options',
          'queue_order_strategy',
          'max_connections',
+         'health_check_interval',
          'priority_steps')  # <-- do not add comma here!
     )
 
     connection_class = redis.Connection if redis else None
 
     def __init__(self, *args, **kwargs):
-        super_ = super(Channel, self)
-        super_.__init__(*args, **kwargs)
+        super(Channel, self).__init__(*args, **kwargs)
 
         if not self.ack_emulation:  # disable visibility timeout
             self.QoS = virtual.QoS
@@ -905,14 +906,18 @@ class Channel(virtual.Channel):
             'socket_connect_timeout': self.socket_connect_timeout,
             'socket_keepalive': self.socket_keepalive,
             'socket_keepalive_options': self.socket_keepalive_options,
+            'health_check_interval': self.health_check_interval,
         }
 
         conn_class = self.connection_class
+
+        # If the connection class does not support the `health_check_interval`
+        # argument then remove it.
         if (
             hasattr(conn_class, '__init__') and
-            accepts_argument(conn_class.__init__, 'health_check_interval')
+            not accepts_argument(conn_class.__init__, 'health_check_interval')
         ):
-            connparams['health_check_interval'] = DEFAULT_HEALTH_CHECK_INTERVAL
+            connparams.pop('health_check_interval', None)
 
         if conninfo.ssl:
             # Connection(ssl={}) must be a dict containing the keys:
@@ -1064,8 +1069,12 @@ class Transport(virtual.Transport):
             [add_reader(fd, on_readable, fd) for fd in cycle.fds]
         loop.on_tick.add(on_poll_start)
         loop.call_repeatedly(10, cycle.maybe_restore_messages)
+        health_check_interval = connection.client.transport_options.get(
+            'health_check_interval',
+            DEFAULT_HEALTH_CHECK_INTERVAL
+        )
         loop.call_repeatedly(
-            DEFAULT_HEALTH_CHECK_INTERVAL,
+            health_check_interval,
             cycle.maybe_check_subclient_health
         )
 
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -917,7 +917,7 @@ class Channel(virtual.Channel):
             hasattr(conn_class, '__init__') and
             not accepts_argument(conn_class.__init__, 'health_check_interval')
         ):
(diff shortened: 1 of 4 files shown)


KFoxder:
@auvipy Yeah I can add a test case instead of changing the test case that I changed. Is that what you are asking for?

kombu/transport/redis.py
```diff
@@ -905,14 +906,18 @@ def _connparams(self, asynchronous=False):
...
+
+        # If the connection class does not support the `health_check_interval`
+        # argument then remove it.
         if (
             hasattr(conn_class, '__init__') and
-            accepts_argument(conn_class.__init__, 'health_check_interval')
+            not accepts_argument(conn_class.__init__, 'health_check_interval')
         ):
-            connparams['health_check_interval'] = DEFAULT_HEALTH_CHECK_INTERVAL
+            connparams.pop('health_check_interval', None)
```
georgepsarakis: The missing value argument for `pop` is not required, since the dictionary key always exists at this point. Of course it does not create a problem, but it may be misleading or cover a future bug.

t/unit/transport/test_redis.py
```diff
@@ -733,12 +733,15 @@ def test_register_with_event_loop(self):
         transport.cycle = Mock(name='cycle')
         transport.cycle.fds = {12: 'LISTEN', 13: 'BRPOP'}
         conn = Mock(name='conn')
+        conn.client = Mock(name='client', transport_options={
```
georgepsarakis: I think a dedicated test case for this feature (configurable health check interval) would be more extensible and easier to troubleshoot in the future. What do you think?

georgepsarakis: As far as I can tell, three cases need to be examined:
- what happens if the connection class does not accept the `health_check_interval` argument?
- health check interval is configured and passed accordingly to `call_repeatedly` (already added, just needs to be extracted)
- health check interval is not configured (default value should be used)

KFoxder:
@georgepsarakis Thanks for the review! I will update when I have time this week.

kombu/transport/redis.py
```diff
@@ -1064,8 +1069,12 @@ def on_poll_start():
             [add_reader(fd, on_readable, fd) for fd in cycle.fds]
         loop.on_tick.add(on_poll_start)
         loop.call_repeatedly(10, cycle.maybe_restore_messages)
+        health_check_interval = connection.client.transport_options.get(
```
KFoxder: @georgepsarakis I don't feel great about this. My goal here is to get the passed in `health_check_interval` transport option. I think this will do it, but if someone has a better way, I would happily change it.

kombu/transport/redis.py
KFoxder: @georgepsarakis I was trying to follow the current pattern below here:
https://github.com/celery/kombu/pull/1145/files#diff-1fdc4d4a0fe8a800a42d1341599d8679R939-R941

Would you like me to update those places as well so we are consistent? Happy to leave it better than I found it :)

georgepsarakis: I see! Except for `socket_connect_timeout`, [_filter_tcp_connparams](https://github.com/celery/kombu/pull/1145/files#diff-1fdc4d4a0fe8a800a42d1341599d8679R893) seems to discard the other options. I would not recomment to change the `socket_*` handling in this PR though, as it would complicate the diff and the test cases even further, so if you don't mind let's just apply this on current changes only.

kombu/transport/redis.py
georgepsarakis: How about adding it to the `Channel` instead as a method (since a property already exists)?

```python
health_check_interval = self.channels[0].get_health_check_interval()
```

I am not sure if this is the right interface either since `Transport` uses a list of channels, but I think it is better.

t/unit/transport/test_redis.py
KFoxder: @georgepsarakis I added coverage for all the above.

KFoxder:
@georgepsarakis I updated this PR with those changes so it should be good for another pass.

kombu/transport/redis.py
```diff
@@ -349,9 +349,9 @@ def maybe_check_subclient_health(self):
         for channel in self._channels:
             # only if subclient property is cached
             client = channel.__dict__.get('subclient')
-            if client is not None:
-                if callable(getattr(client, "check_health", None)):
-                    client.check_health()
+            if client is not None \
```
matteius: Why is this change being made?  I personally find the prior version more readable since both end up spanning two lines.

kombu/transport/redis.py
```diff
@@ -905,14 +906,18 @@ def _connparams(self, asynchronous=False):
...
         conn_class = self.connection_class
+
+        # If the connection class does not support the `health_check_interval`
+        # argument then remove it.
         if (
             hasattr(conn_class, '__init__') and
-            accepts_argument(conn_class.__init__, 'health_check_interval')
+            not accepts_argument(conn_class.__init__, 'health_check_interval')
         ):
-            connparams['health_check_interval'] = DEFAULT_HEALTH_CHECK_INTERVAL
```
matteius: The old logic would add the DEFAULT_HEALTH_CHECK_INTERVAL if the connection class did support the argument, but now we are not defaulting it here and we are popping it from the connparams -- would popping it  like that blow up if health_check_interval were not already in the connparams?  I question why we are inverting this logic pattern.

KFoxder: @matteius I inverted it because when you read this entire method the pattern seems to be to remove items not add them conditionally. Not that I necessarily think its a good programming pattern but I thought it should at least be consistent with how all other arguments are work here.

kombu/transport/redis.py
KFoxder: @matteius I don't feel too strong here but I generally avoid nested if statements if they aren't necessary since "flat is better than nested". When I was reading this I found it less readable because of the nest (but maybe thats just me). But I am not a maintainer so it is not up to me 😄.

KFoxder:
@georgepsarakis @auvipy where do we stand with this?

kombu/transport/redis.py
KFoxder: @auvipy Ok I will resolve this conversation then.

KFoxder:
@auvipy do you have a timeline of when this will end up in a dot `4.6.8` release?

thedrow:
It's a new feature so it has to be 4.7.0...

KFoxder:
@thedrow Ok that makes sense. Do we have timeline for that? I only ask because we would like to use this in production

auvipy:
expect kombu 4.7 and celery 4.4.x by 30 jan

KFoxder:
@auvipy do you know if kombu 4.7 and celery 4.4.x will be released soon?

KFoxder:
@auvipy or @thedrow Any update on kombu 4.7 and celery 4.4.x?

auvipy:
yes sir by 28/29th of feb

platech:
@auvipy any news on the release?

auvipy:
consider contributing to my work for celery here https://opencollective.com/celery

KFoxder:
@auvipy we at @ycharts just contributed to the celery project on opencollective for all your hard work! Much appreciated. Do you mind letting us know when the 4.7 release can happen with this fix included in it?
image

auvipy:
thanks, man! very very soon <3

auvipy:
it's already on PyPI :)

Issue #1149, opened 2020-01-28
For pyamqp, if connection url contains amqps, then certificates are never validated against provided ca certificates


pyushagarwal:
For pyamqp transport, if connection url contains the string **amqps**(notice the s), then the server certificates are never validated against the ca certs, even if they are provided.
Inside  pyamqp.py. file line 182

`class SSLTransport(Transport):
    """AMQP SSL Transport."""

    def __init__(self, *args, **kwargs):
        super(SSLTransport, self).__init__(*args, **kwargs)

        # ugh, not exactly pure, but hey, it's python.
        self.client.ssl = True`

**The above code demonstrates the part where they are overwritten.** 

Creating the custom connection object

connection = kombu.Connection(
                    "amqps://guest:guest@localhost:5671",
                    login_method="PLAIN",
                    ssl={
                        'ca_certs': '"ca_cert_path",
                        'cert_reqs': ssl.CERT_REQUIRED
                    }
)

But if the url contains the string amqp(without s), then the server certificates are validated.

pyushagarwal:
I have raised a PR for this bug

PR #1151, opened 2020-01-30
#1149 Enabled ssl certificate verification against ca certificates when amqps is used for pyamqp transport

pyushagarwal:


pyushagarwal:
> is it possible for you to add tests for this change?

@auvipy  Added unit tests for the scenario

PR #1154, opened 2020-02-12
Add predefined_queues option to SQS transport

jonathaningram:
Originally started in #834. I have modified the original concept of that PR. That PR added the transport setting `predefined_queue_urls`. Instead, this PR adds the transport setting `predefined_queues` so that callers can provide the SQS queue URL and the AWS credential pair. This is necessary for our current cloud environment whereby SQS queues are provisioned once outside of Kombu/Celery and we have access to the SQS queue URL and AWS creds which we need to inject into Kombu/Celery. This PR allows us to do that.

Because each queue can have its own AWS creds, I had to modify the `SQS` class `sqs` method/property so that it could be given a queue name. Given a queue name and when predefined queues are being used, a queue-specific SQS client could be returned that used the correct AWS creds.

Note: I didn't allow for the predefined queues to contain the AWS region. The region is still global across the client instance.

I have added a few tests for predefined queues. This also required modifying some of the test classes to support multiple SQS mock clients.

I ran `make lint` across the code base and fixed up obvious issues. I notice that other parts of the codebase fail the lint test so I'm not sure if it's up to date. Note: there are lint failures regarding line length in `SQS.py` but I skipped fixing those due to there being other lint failures in the codebase.

I believe that I have addressed all of the feedback in the original PR but if I missed anything please let me know. Thank you @aeijdenberg for the head start.

### DIFF
--- a/kombu/asynchronous/aws/sqs/connection.py
+++ b/kombu/asynchronous/aws/sqs/connection.py
@@ -58,7 +58,7 @@ class AsyncSQSConnection(AsyncAWSQueryConnection):
             queue.id, callback=callback,
         )
 
-    def receive_message(self, queue,
+    def receive_message(self, queue, queue_url,
                         number_messages=1, visibility_timeout=None,
                         attributes=None, wait_time_seconds=None,
                         callback=None):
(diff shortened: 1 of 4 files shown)


jonathaningram:
Closed in favour of https://github.com/celery/kombu/pull/1156 (this was a fork from the wrong user, apologies for the dupe PR notifications).

PR #1163, opened 2020-03-18
Fix support for python 3.7

sayanarijit:
Rename variable `async` to `asynchronous` because `async` is a Python
keyword.


sayanarijit:
nope.

PR #1170, opened 2020-04-14
Prevent failure if AWS creds are not explicitly defined on predefined…

isugimpy:
…_queues

This allows the fallthrough of the creds to the standard boto3 credential chain if the user does not provide explicit credentials, which allows for the use of IAM roles.


isugimpy:
I see that other appveyor builds are failing for PRs, not sure that there's anything for me to do here.

Issue #1174, opened 2020-04-22
Exchange.publish() is broken when string provided as message


bernd-wechner:
In` Exchange.publish()`, if a string is provided as a message and a content_encoding is not, then it crashes when it comes to `self.channel.prepare_message()`  which has a default value for an explictly named kwarg of None.

```
    def prepare_message(self, body, priority=None,
                        content_type=None, content_encoding=None,
                        headers=None, properties=None):
```
 It subsequently crashes when trying to encode the message.

A number of solutions are possible but the easiest is to assume a string encoding in Echange.Message or use the [chardet](https://pypi.org/project/chardet/) library to estimate one.

The Message method has a lead in comment of:

```
# XXX This method is unused by kombu itself AFAICT [ask].
```

Well I'm asking ;-). It **IS** used by the Exchange.publish() method. To wit should be maintained and fixed.

bernd-wechner:
One proposed PR: https://github.com/celery/kombu/pull/1175

bernd-wechner:
Fix #1174 accepted.

PR #1175, opened 2020-04-22
A possible fix for Issue #1174

bernd-wechner:
Could also use the chardet module, and could also change the default on Channel.prepare_message. This struck me as the simplest fix for now. Open to feedback.

ref: https://github.com/celery/kombu/issues/1174

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -100,12 +100,17 @@ class pytest(setuptools.command.test.test):
         sys.exit(pytest.main(self.pytest_args))
 
 
+def readme():
+    with open('README.rst') as f:
+        return f.read()
+
+
 setup(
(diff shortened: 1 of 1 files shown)


auvipy:
also flake8 error

bernd-wechner:
flake8 compliance is easy (though the logs hard to read).

TavisCI was failing a lot but unrelated.

Need to look into how to write a unit test? Technically easy but needs a running RabbitMQ instance to test that `Exchange.publish(string)` works and does not crash with an encoding exception.

Any tips in that space?

bernd-wechner:
Replaced by https://github.com/celery/kombu/pull/1176

PR #1176, opened 2020-04-23
Another possible fix for #1174

bernd-wechner:
Flake8 compliant this time.

Unit test to follow if possible but see no unit testing infrastructure in the repo. Fail behaviour on this is Exchange.publish(text) crashes with an encoding error.  A Pass is that the message is published, which can be confirmed with Queue.get() and comparing the payload with that which was published. Requires a running RabbitMQ instance to test (as that is the context of the crash).

A unit test though, while holding value is of the functionality of Exchange.publish() - currently broken for text messages, is not central to the value of fixing this break.

ref: https://github.com/celery/kombu/issues/1174

Apologies for a new request, inadvertently deleted the prior one (cutting my teeth with gitkraken, github and more).

### DIFF
--- a/kombu/entity.py
+++ b/kombu/entity.py
@@ -245,9 +245,11 @@ class Exchange(MaybeChannelBound):
 
             headers (Dict): Message headers.
         """
-        # XXX This method is unused by kombu itself AFAICT [ask].
         properties = {} if properties is None else properties
         properties['delivery_mode'] = maybe_delivery_mode(self.delivery_mode)
+        if (isinstance(body, string_t) and
+                properties.get('content_encoding', None)) is None:
+            kwargs['content_encoding'] = 'utf-8'
         return self.channel.prepare_message(
             body,
             properties=properties,


bernd-wechner:
Appveyor seems to be failing for unrelated reasons. What's that about?

kombu/entity.py
```diff
@@ -245,9 +245,11 @@ def Message(self, body, delivery_mode=None, properties=None, **kwargs):
 
             headers (Dict): Message headers.
         """
-        # XXX This method is unused by kombu itself AFAICT [ask].
         properties = {} if properties is None else properties
         properties['delivery_mode'] = maybe_delivery_mode(self.delivery_mode)
+        if (isinstance(body, string_t) and
+                properties.get('content_encoding', None)) is None:
```
drsantos20, line 251: Why we are double-checking `properties` None here since we're already doing this on line 248?

drsantos20: The tests for this change are still passing?

bernd-wechner: Not sure I understand.Line 248 simply provides an empty dict if no props are provided. In fact looking at it line 248 strikes me as poor coding. Easier to make {} the default on the function declaration.

This is not about checking properties. it's about the fact that this:

```
x = Exchange(name, channel)
x.publish("my message")
```
crashes downstream with an encoding error in:

```
...python3.8/site-packages/amqp/method_framing.py", line 110, in write_frame
    body = body.encode(encoding)
TypeError: encode() argument 'encoding' must be str, not None
```

One could argue that a fix in amqp/method_framing.py is better?

But I figure that it's the `Exchange.publish()` method that is broken. `Queue.publish()` isn't and hence I figured a fix inside of Kombu is better.

You have in your comment not included line 252 which actually fixes this, namely provides a default encoding for string arguments. 

Again one could argue that the `chardet` module can take a better stab at a default by inspecting the string.

bernd-wechner:
So, wondering if this fits the criteria for a merge into master for the next Kombu release? Just curious.

PR #1185, opened 2020-05-03
SQLAlchemy transport: Use Query.with_for_update() instead of deprecated Query.with_lockmode().

matusvalo:
Based on SQLAlchemy documentation:

* method `sqlalchemy.orm.query.Query.with_lockmode(mode)` with `mode='update'` - translates to `FOR UPDATE` (Deprecated since version 0.9)
* method `sqlalchemy.orm.query.Query.with_for_update()` When called with no arguments, the resulting `SELECT` statement will have a `FOR UPDATE` clause appended.

For details see:
https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.with_for_update
https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.with_lockmode

### DIFF
--- a/kombu/transport/sqlalchemy/__init__.py
+++ b/kombu/transport/sqlalchemy/__init__.py
@@ -99,7 +99,7 @@ class Channel(virtual.Channel):
             self.session.execute('BEGIN IMMEDIATE TRANSACTION')
         try:
             msg = self.session.query(self.message_cls) \
-                .with_lockmode('update') \
+                .with_for_update() \
                 .filter(self.message_cls.queue_id == obj.id) \
                 .filter(self.message_cls.visible != False) \
                 .order_by(self.message_cls.sent_at) \


rhyamada:
Hi, 

There are a problem with method _get on kombu.transport.sqlalchemy.Channel
If a db throws a exception method should emit a session.rollback, instead on finnaly there are a session.commit. Them long running process of celery beat(workers) can't recover from this error:

```
sqlalchemy.exc.StatementError: (sqlalchemy.exc.InvalidRequestError) Can't reconnect until invalid transaction is rolled back
[SQL: SELECT kombu_queue.id AS kombu_queue_id, kombu_queue.name AS kombu_queue_name
FROM kombu_queue
WHERE kombu_queue.name = %(name_1)s
 LIMIT %(param_1)s]
[parameters: [immutabledict({})]]
```

my monkey patched version:

```python
def _get(self, queue):
    obj = self._get_or_create(queue)
    if self.session.bind.name == 'sqlite':
        self.session.execute('BEGIN IMMEDIATE TRANSACTION')
    try:
        msg = self.session.query(self.message_cls) \
            .with_for_update() \
            .filter(self.message_cls.queue_id == obj.id) \
            .filter(self.message_cls.visible != False) \
            .order_by(self.message_cls.sent_at) \
            .order_by(self.message_cls.id) \
            .limit(1) \
            .first()
        if msg:
            msg.visible = False
            ret = json.loads(bytes_to_str(msg.payload))
            self.session.commit()
            return ret
    except:
        pass
    self.session.rollback()
    raise Empty()
```

Issue #1186, opened 2020-05-04
Requeuing undeliverable message for queue '{celery}': No consumers.


pvyParts:
# Issue
Kombu throws this error for every task that is action'd by a worker, it appears to be superficial and just fixes it it's self but the log spam is wild;
```[2020-05-04 13:12:21,035: WARNING/MainProcess] Requeuing undeliverable message for queue '{celery}': No consumers.```

[Link to the issue on our repo](https://gitlab.com/allianceauth/allianceauth/-/issues/1229)
[Link to our celery config](https://gitlab.com/allianceauth/allianceauth/-/blob/master/allianceauth/project_template/project_name/celery.py)

# Findings
I did some digging and found [this](https://github.com/celery/kombu/commit/f2970e81f823fca0fea3f5f436b02f24b875d39b) commit that appears to have [changed the prio info saved into Redis](https://github.com/celery/kombu/blob/f2970e81f823fca0fea3f5f436b02f24b875d39b/kombu/transport/redis.py#L782) but doesn't change the [parsing of the data back out](https://github.com/celery/kombu/blob/f2970e81f823fca0fea3f5f436b02f24b875d39b/kombu/transport/redis.py#L747).

# Enviro
We have a fairly simple use case where we use a Redis with a priority queue and a few celery workers.

## Setup
```python
import os
from celery import Celery
os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'myauth.settings.local')
from django.conf import settings 
app = Celery('myauth')
app.config_from_object('django.conf:settings')
app.conf.broker_transport_options = {
    'priority_steps': list(range(10)), 
    'queue_order_strategy': 'priority', 
}
app.conf.task_default_priority = 5  priority (9)
app.conf.worker_prefetch_multiplier = 1 
app.conf.ONCE = {
    'backend': 'allianceauth.services.tasks.DjangoBackend',
    'settings': {}
}
```
## Testing
I did some local testing in my dev environment and modified[this line](https://github.com/celery/kombu/blob/f2970e81f823fca0fea3f5f436b02f24b875d39b/kombu/transport/redis.py#L782) to confirm, i haven't seen the error the patch was fixing so figure this is the wrong way to go.
```python
    def _q_for_pri(self, queue, pri):
        pri = self.priority(pri)
        if pri:
            return "{}{}{}".format(queue, self.sep, pri)
        return queue
```
Was just using some basic dummy tasks for testing
```python
from celery import shared_task
import time

@shared_task
[...]

PR #1188, opened 2020-05-06
Fix consumer encoding in Redis priority queues

pvyParts:
Details in #1186

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -779,7 +779,7 @@ class Channel(virtual.Channel):
     def _q_for_pri(self, queue, pri):
         pri = self.priority(pri)
         if pri:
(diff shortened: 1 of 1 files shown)


kombu/transport/redis.py
```diff
@@ -779,7 +779,7 @@ def _size(self, queue):
     def _q_for_pri(self, queue, pri):
         pri = self.priority(pri)
         if pri:
-            return "{{{}}}{}{}".format(queue, self.sep, pri)
+            return "{}{}{}".format(queue, self.sep, pri)
```
auvipy, line 782: can you also update tests?

pvyParts: Yep ofcourse, I've just been fighting with tox locally atm,

PR #1189, opened 2020-05-13
fix 100% cpu usage on linux while using sqs

and800:
fix https://github.com/celery/celery/issues/5299

kombu uses cURL [`multi_socket`](https://curl.haxx.se/libcurl/c/libcurl-multi.html) API, which is non-blocking callback-based API. But it doesn't work well in linux, because of [epoll weird file descriptor management](https://idea.popcount.org/2017-03-20-epoll-is-fundamentally-broken-22/). When you use epoll, you must always remove FD from epoll before closing FD, but when curl decides that it doesn't need some particular socket, it will close that socket and _after that_ it will inform you that the socket is closed - so you don't have a chance to remove it from epoll in time. See https://curl.haxx.se/libcurl/c/curl_multi_socket_action.html for more info about this API.

In our case we periodically query an SQS queue, and curl performs the queries via keep-alive https connection. But sometimes (about once per minute or two) SQS sends a response with `Connection: close` header, then curl closes the socket.

I suggest this workaround: before every curl invocation, let's remove from our selector every curl-related file descriptor. And after we get control back from curl, we insert those back (most of the time curl will update socket states, so we should not forget to consider it). So as a result we are going to work with epoll in old-fashioned select/poll way.

### DIFF
--- a/kombu/asynchronous/http/curl.py
+++ b/kombu/asynchronous/http/curl.py
@@ -75,36 +75,47 @@ class CurlClient(BaseClient):
         self._set_timeout(0)
         return request
 
+    # the next two methods are used for linux/epoll workaround:
+    # we temporarily remove all curl fds from hub, so curl cannot
+    # close a fd which is still inside epoll
+    def _pop_from_hub(self):
+        for fd in self._fds:
+            self.hub.remove(fd)
+
+    def _push_to_hub(self):
+        for fd, events in self._fds.items():
+            if events & READ:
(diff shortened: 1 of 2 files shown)


auvipy:
thanks for handling this, would need to update test too

and800:
yup, that's still a draft, not completely sure about it

auvipy:
OK keep working on it

and800:
It seems to work well @auvipy

PR #1193, opened 2020-05-19
Ensure connection when connecting to broker

### LINKED ISSUE #1179: Failover connection is partially broken
Kombu contains failover connection functionality - see [1]. Currently it is partially broken (reproducible on master branch) - e.g.:

How to reproduce issue:

```python
from kombu import Connection, Producer, Consumer, Queue

test_queue = Queue('test', routing_key='test')

with Connection('amqp://broken_borker.example.com;amqp://broker.example.com') as conn:
    with conn.channel() as channel:
        producer = Producer(channel)
        producer.publish(
            {'hello': 'world'},
            retry=True,
            exchange=test_queue.exchange,
            routing_key=test_queue.routing_key,
            declare=[test_queue],
            serializer='pickle'
        )
```
**Expected Result**: Kombu should connect to `broker.example.com`
**Actual Result**: Kombu raises exception:
```
  File "/home/matus/environments/kombu37/lib/python3.7/site-packages/amqp/transport.py", line 137, in _connect
    host, port, family, socket.SOCK_STREAM, SOL_TCP)
  File "/usr/lib/python3.7/socket.py", line 748, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno 8] Name or service not known

  File "<stdin>", line 2, in <module>
  File "/home/matus/dev/kombu/kombu/connection.py", line 288, in channel
    chan = self.transport.create_channel(self.connection)
  File "/home/matus/dev/kombu/kombu/connection.py", line 839, in connection
    self._connection = self._establish_connection()
  File "/home/matus/dev/kombu/kombu/connection.py", line 794, in _establish_connection
    conn = self.transport.establish_connection()
  File "/home/matus/dev/kombu/kombu/transport/pyamqp.py", line 130, in establish_connection
    conn.connect()
  File "/home/matus/environments/kombu37/lib/python3.7/site-packages/amqp/connection.py", line 311, in connect
    self.transport.connect()
  File "/home/matus/environments/kombu37/lib/python3.7/site-packages/amqp/transport.py", line 77, in connect
    self._connect(self.host, self.port, self.connect_timeout)
  File "/home/matus/environments/kombu37/lib/python3.7/site-packages/amqp/transport.py", line 148, in _connect
    "failed to resolve broker hostname"))
  File "/home/matus/environments/kombu37/lib/python3.7/site-packages/amqp/transport.py", line 161, in _connect
    self.sock.connect(sa)
socket.timeout: timed out
```

But as mentioned, for some usage of kombu library, the failover strategy works -e.g. working script:
```python
from kombu import Connection, Producer, Consumer, Queue

test_queue = Queue('test', routing_key='test')

with Connection('amqp://broken_borker.example.com;amqp://broker.exaple.com') as conn:
    with conn.default_channel as channel:
        producer = Producer(channel)
        producer.publish(
            {'hello': 'world'},
            retry=True,
            exchange=test_queue.exchange,
            routing_key=test_queue.routing_key,
            declare=[test_queue],
            serializer='pickle'
        )
```

The difference between working and broken version is that using `conn.default_channel` vs. `conn.channel()`

The root cause of issue is that  method `Connection.ensure_connection()` is called in `default_channel` property and not in `channel()` method.

[1] https://kombu.readthedocs.io/en/latest/reference/kombu.html#connection

matusvalo:
This PR moves connection ensuring functionality into the connection factories to have consistent functionality across all API. As side efect it unifies `Connection.channel()` and `Connection.default_channel`:
1.  Now both channel factories blocks till connection is created + both provide possibility to configure this via `transport_options` parameter
2. Now both channel factories raises consistently the same exception on connection error: `kombu.exceptons.OperationalError`

Hence this change breaks backward compatibility for users which rely on this API.

Fixes #1179

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -280,7 +280,7 @@ class Connection(object):
     def connect(self):
         """Establish connection to server immediately."""
         conn_opts = self._extract_failover_opts()
-        return self.ensure_connection(**conn_opts)
+        return self._ensure_connection(**conn_opts)
 
     def channel(self):
         """Create and return a new channel."""
@@ -380,10 +380,20 @@ class Connection(object):
         self._close()
     close = release
 
-    def ensure_connection(self, errback=None, max_retries=None,
-                          interval_start=2, interval_step=2, interval_max=30,
-                          callback=None, reraise_as_library_errors=True,
-                          timeout=None):
+    def ensure_connection(self, *args, **kwargs):
+        """Public interface of _ensure_connection for retro-compatibility.
+
+        Returns kombu.Connection instance.
+        """
+        self._ensure_connection(*args, **kwargs)
+        return self
+
+    def _ensure_connection(
+        self, errback=None, max_retries=None,
+        interval_start=2, interval_step=2, interval_max=30,
+        callback=None, reraise_as_library_errors=True,
+        timeout=None
+    ):
         """Ensure we have a connection to the server.
 
         If not retry establishing the connection with the settings
@@ -533,7 +543,7 @@ class Connection(object):
                         remaining_retries = None
                         if max_retries is not None:
                             remaining_retries = max(max_retries - retries, 1)
-                        self.ensure_connection(
+                        self._ensure_connection(
                             errback,
                             remaining_retries,
                             interval_start, interval_step, interval_max,
@@ -850,7 +860,7 @@ class Connection(object):
         if not self._closed:
             if not self.connected:
                 conn_opts = self._extract_failover_opts()
-                self._connection = self.ensure_connection(**conn_opts)
+                self._connection = self._ensure_connection(**conn_opts)
             return self._connection
 
     def _connection_factory(self):
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -279,8 +279,8 @@ class Connection(object):
 
     def connect(self):
         """Establish connection to server immediately."""
-        self._closed = False
-        return self.connection
+        conn_opts = self._extract_failover_opts()
+        return self.ensure_connection(**conn_opts)
 
     def channel(self):
         """Create and return a new channel."""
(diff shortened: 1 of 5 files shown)


matusvalo:
Regarding point 1. and 2. It is questionable which functionality should be used to cover factories. I used semantics of `default_channel` which blocks and returns `OperationalError` *but* this can be changed to semantics of `channel()` which does not block and directly return transport exception.

matusvalo:
@thedrow @georgepsarakis could you also peer review this PR?

auvipy:
semantics of channel() which does not block and directly return transport exception, what benefits will this provide?

matusvalo:
@auvipy not sure. I don't see big benefits because blocking behavior can be adjusted in this way using `transport_options` so you can have both. My biggest concerns are about breaking client code which is expecting this behavior for channel() factory method.

thedrow:
My biggest concern is that this PR will break Celery.
Run all the relevant kombu tests and the celery test suite with this patch applied to py-amqp before we review the design.

auvipy:
not sure why py27 n 35 is failing

auvipy:
the travis failures are ambigious

matusvalo:
@thedrow The kombu test suite is passing (but there were changes needed see PR diff). The Celery CI is also passing with no changes - see:

https://travis-ci.org/github/matusvalo/celery/builds/690546756

kombu/connection.py
```diff
@@ -279,8 +279,8 @@ def _debug(self, msg, *args, **kwargs):
 
     def connect(self):
         """Establish connection to server immediately."""
-        self._closed = False
-        return self.connection
+        conn_opts = self._extract_failover_opts()
+        return self._ensure_connection(**conn_opts)
```
ConorStevenson, line 283: @matusvalo this now creates and returns a new connection, but doesn't set `self._connection`
```
conn = Connection(...)
conn.connect()
conn.connected -> False
```

kontinuity: This broke our status check flow

auvipy: can yu check new commits on master?

Issue #1198, opened 2020-05-26
Celery crashes in cases where there aren’t enough available workers to start acting on “in-flight” messages in the SQS queue


frehty:
We are using Celery and SQS in our current project.
We observed that Celery actually crashes in cases where there aren’t enough available workers to start acting on “in-flight” messages in the SQS queue. When Celery gets a message from an SQS queue, it also gets a "receipt handle" for this message (that is now "in-flight" state). If the number of workers is insufficient and Celery could not start processing this message within visibility timeout, the message returns to the beginning of SQS queue. The “receipt handle” for this message becomes invalid, and when Celery tries to acknowledge it, Celery crashes because botocore.exceptions.ClientError occurs.

**Ex.:**
ClientError('An error occurred (InvalidParameterValue) when calling the **DeleteMessage** operation: Value AQEBomFDt0zBfod6H+XQQ77B13TlwZXnJKVtV4cMo5AsvoloQVOaT
xC/vhZj7QxclQ8roWrsVyLzpqJUxcXzm0HvS0AZXNz+DyKCP9NotAhxJvYrNIt5h9Z9UdiwvMZMwkFUpPxg2nSZx12O+6eeKABEnsKwAAS0vEwGklxgB9dXC7Vmnk3qTmhZdvn9e2GnzGKjRRsnJT4J5CWH87ukKrQTkL8Twuni31A2X5KkP1c2eQKUC8UhK8sOnkwVMu7YX13JdTXM
VYdTHIO4CCEtDTcN1ujJ0uGysOOY1ChiupiIcpI= for parameter **ReceiptHandle is invalid. Reason: The receipt handle has expired.**',)

**NOTE:** An “in-flight” message is a message that has been read by the main Celery worker and added to Celery’s internal queue, where it waits for an available worker to acknowledge it and begin processing the given task. It’s at this point that the SQS visibility timeout is applied to each message and begins its countdown. Once the message is acknowledged, Celery removes it from the SQS queue, so it is no longer “in-flight” and the visibility timeout no longer applies.

**Steps to reproduce:**
1. **Amount of workers:** 1
2. **Visibility timeout:** 30 secs
3. **Amount of tasks (messages in SQS queue):** 12
4. **Average task run time:** 1-2 mins

**Actual result:** Due to the visibility timeout value, which is less than the minimum execution time for one task, Celery crashed after the very first message returned to the beginning of SQS queue.

[...]

PR #1199, opened 2020-05-26
Fix for issue #1198: Celery crashes in cases where there aren’t enough available workers to start acting on “in-flight” messages in the SQS queue

frehty:
This fix works fine for our project.

**Steps to reproduce:**
1. Amount of workers: 1
2. Visibility timeout: 30 secs
3. Amount of tasks (messages in SQS queue): 12
4. Average task run time: 1-2 mins

**Actual result:** Due to the visibility timeout value, which is less than the minimum execution time for one task, messages periodically returned to the beginning of SQS queue, but Celery successfully processed them when a worker becomes available. No messages have been lost. No duplicate tasks were created.


frehty:
> can you please update the tests?

Sure, I will do it ASAP. I will mark my PR as WIP.

frehty:
Hello @auvipy!

I've added the following tests in test_SQS.py:
- test_basic_ack_with_mocked_channel_methods
- test_basic_ack_without_sqs_message
- test_basic_ack_invalid_receipt_handle

I also added a line with additional check to the test_basic_ack  test

PR #1211, opened 2020-06-05
Revert incompatible changes in #1193 and additional improvements

### LINKED ISSUE #1210: Connecting to broker now blocks forever by default
After #1193 several methods started to block connection to broker till is not established. This can take forever when  broker won't appear. Moreover, it changed raised exceptions to OperationalError instead of native transport one. It seems that this change affects users  - see: #429 

The semantics of this should be at least postponed an introduced properly after proper discussion.

matusvalo:
Fixes #1210

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -279,8 +279,9 @@ class Connection(object):
 
     def connect(self):
         """Establish connection to server immediately."""
-        conn_opts = self._extract_failover_opts()
-        return self._ensure_connection(**conn_opts)
+        return self._ensure_connection(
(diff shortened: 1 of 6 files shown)


auvipy:
ping me when ready for review

auvipy:
2 integration tests were failing

auvipy:
would be great if you could complete this by this week

matusvalo:
@auvipy I am ready. There is failing just apicheck due new release of botocore.

matusvalo:
I have extended tests. Moreover I have verified them against older version of kombu and they are passing so the API should be compatible.

Issue #1212, opened 2020-06-11
binascii.Error: Incorrect padding


qasimzee:
While consuming a message I get the following error:
```
Traceback (most recent call last):
  File "/home/vagrant/.virtualenvs/service/lib/python3.6/site-packages/kombu/mixins.py", line 175, in run
    for _ in self.consume(limit=None, **kwargs):
  File "/home/vagrant/.virtualenvs/service/lib/python3.6/site-packages/kombu/mixins.py", line 197, in consume
    conn.drain_events(timeout=safety_interval)
  File "/home/vagrant/.virtualenvs/service/lib/python3.6/site-packages/kombu/connection.py", line 323, in drain_events
    return self.transport.drain_events(self.connection, **kwargs)
  File "/home/vagrant/.virtualenvs/service/lib/python3.6/site-packages/kombu/transport/virtual/base.py", line 963, in drain_events
    get(self._deliver, timeout=timeout)
  File "/home/vagrant/.virtualenvs/service/lib/python3.6/site-packages/kombu/utils/scheduling.py", line 56, in get
    return self.fun(resource, callback, **kwargs)
  File "/home/vagrant/.virtualenvs/service/lib/python3.6/site-packages/kombu/transport/virtual/base.py", line 1001, in _drain_channel
    return channel.drain_events(callback=callback, timeout=timeout)
  File "/home/vagrant/.virtualenvs/service/lib/python3.6/site-packages/kombu/transport/SQS.py", line 174, in drain_events
    self._poll(self.cycle, callback, timeout=timeout)
  File "/home/vagrant/.virtualenvs/service/lib/python3.6/site-packages/kombu/transport/virtual/base.py", line 402, in _poll
    return cycle.get(callback)
  File "/home/vagrant/.virtualenvs/service/lib/python3.6/site-packages/kombu/utils/scheduling.py", line 56, in get
    return self.fun(resource, callback, **kwargs)
  File "/home/vagrant/.virtualenvs/service/lib/python3.6/site-packages/kombu/transport/SQS.py", line 366, in _get_bulk
    for msg in self._messages_to_python(resp['Messages'], queue):
  File "/home/vagrant/.virtualenvs/service/lib/python3.6/site-packages/kombu/transport/SQS.py", line 328, in _messages_to_python
    return [self._message_to_python(m, queue, q) for m in messages]
[...]

PR #1213, opened 2020-06-11
Base 64 encode before decoding [WIP]

qasimzee:
Issue: https://github.com/celery/kombu/issues/1212

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -284,6 +284,8 @@ class Channel(virtual.Channel):
             body = base64.b64decode(message['Body'].encode())
         except TypeError:
             body = message['Body'].encode()
+        except base64.binascii.Error:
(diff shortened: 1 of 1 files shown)


kombu/transport/SQS.py
```diff
@@ -284,6 +284,8 @@ def _message_to_python(self, message, queue_name, queue):
             body = base64.b64decode(message['Body'].encode())
         except TypeError:
             body = message['Body'].encode()
+        except base64.binascii.Error:
+            body = base64.b64decode(base64.b64encode(message['Body'].encode()))
```
matusvalo, line 288: Do we need to encode and decode message?

auvipy:
should we close this as per the issue?

thedrow:
I don't understand why your fix solves the problem.

matusvalo:
This PR seems to be inactive and the purpose seems to be unclear. I would close this issue.

PR #1218, opened 2020-06-24
buffer() was removed in Python 3 so use bytes

cclauss:
https://docs.python.org/2.7/library/stdtypes.html#memoryview-type

https://portingguide.readthedocs.io/en/latest/etc.html#raw-buffer-protocol-buffer-and-memoryview

### DIFF
--- a/kombu/transport/qpid.py
+++ b/kombu/transport/qpid.py
@@ -116,7 +116,7 @@ except ImportError:  # pragma: no cover
     qpid = None
 
 
-from kombu.five import buffer, Empty, items, monotonic, PY3
+from kombu.five import buffer_t, Empty, items, monotonic, PY3
 from kombu.log import get_logger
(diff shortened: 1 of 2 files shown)


kombu/transport/qpid.py
```diff
@@ -124,6 +124,11 @@
 
 logger = get_logger(__name__)
 
+try:
+    buffer
```
thedrow, line 128: You should import buffer from `kombu.five` instead.

cclauss: ImportError: cannot import name 'buffer' from 'vine.five'

thedrow:
The tests are failing.
Please ensure that your refactor works.

cclauss:
My refactor worked fine until https://github.com/celery/kombu/pull/1218#discussion_r453283999

Reverting to something that works.

PR #1219, opened 2020-06-25
Add an SQS transport option for custom botocore config

kjagiello:
We have experienced some problems with the client not being able to connect to SQS and reaching the wsgi worker time limit (60s) and thus terminating. The default connect timeout for botocore is 60s and we want to be able to adjust it, so that we can handle this issue more gracefully. This PR implements a way of doing just that.


kjagiello:
Let me know if I can help you push this change forward @auvipy

PR #1226, opened 2020-07-24
Stop automatically converting the amqp:// prefix to librabbitmq:// + refactored test

Alireza2n:
Based on issue [#4693](https://github.com/celery/celery/issues/4693#issuecomment-390771991) on celery repo, librabbitmq is no longer the recommended package, but the code in kombu does not follow this recommendation.
It's also been discussed in [PR #5873](https://github.com/celery/celery/pull/5873).
This PR removes the code where `kombu.Connection()`'s `__init__()` method automatically converts the amqp:// prefix to librabbitmq://.
Also `test_prefer_librabbitmq_over_amqp_when_available()` was rewritten to `test_stop_converting_amqp_to_librabbitmq_when_available()`.

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -248,8 +248,6 @@ class Connection(object):
                      insist, ssl, transport, connect_timeout,
                      login_method, heartbeat):
         transport = transport or 'amqp'
-        if transport == 'amqp' and supports_librabbitmq():
-            transport = 'librabbitmq'
         if transport == 'rediss' and ssl_available and not ssl:
             logger.warning(
                 'Secure redis scheme specified (rediss) with no ssl '
--- a/t/unit/test_connection.py
+++ b/t/unit/test_connection.py
@@ -203,12 +203,12 @@ class test_Connection:
         assert connection._transport is None
         assert connection._connection is None
 
-    def test_prefer_librabbitmq_over_amqp_when_available(self):
+    def test_stop_converting_amqp_to_librabbitmq_when_available(self):
         with patch('kombu.connection.supports_librabbitmq',
                    return_value=True):
             connection = Connection('amqp://')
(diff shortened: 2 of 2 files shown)


Alireza2n:
> there is flake8 issue to address

Done. It was caused by an unused import.

thedrow:
@auvipy We're almost ready for Celery 5.
I think we should keep this for that release since this is a breaking change.

Removing the recommendation is a step in the right direction for Celery 4.x.

@Alireza2n What do you think?

auvipy:
yup removing recommendation could be the first move

Alireza2n:
> @auvipy We're almost ready for Celery 5.
> I think we should keep this for that release since this is a breaking change.
> 
> Removing the recommendation is a step in the right direction for Celery 4.x.
> 
> @Alireza2n What do you think?

@thedrow Hi!
Use of librabbitmq as an optimization recommendation is already removed from celery documentation. 
I'm not sure if this is a breaking change, I'm hoping to get the test and flak8 fixed this weekend.

thedrow:
Why isn't this a breaking change?
If people upgrade and they still use librabbitmq they still expect to use it afterwards.

Alireza2n:
> Why isn't this a breaking change?
> If people upgrade and they still use librabbitmq they still expect to use it afterwards.

Well, the existing code switched automatically to `librabbitmq`, I've removed this behavior and it seems it does not break anything (at least in automatic tests), users still can manually specify their library I believe.

thedrow:
Yes but it shouldn't switch without a major version bump.

Alireza2n:
> Yes but it shouldn't switch without a major version bump.

Got it. 
I'll try to fix it, then you can decide on release and other details. :)

PR #1229, opened 2020-07-28
pass timeout on in _extract_failover_opts

### LINKED ISSUE #1223: Passing transport_options to conn_opts in _extract_failover_opts
Is there a reason I am missing why `timeout` option is not passed forward in `_extract_failover_opts`?

https://github.com/celery/kombu/blob/75027490c71b83342f92b5293461c679233dc25e/kombu/connection.py#L835-L847

Reference: https://github.com/celery/celery/issues/5067#issuecomment-656523289

sschiessl-bcp:
Fixes #1223

Please adjust to your desired target branch.

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -844,6 +844,8 @@ class Connection(object):
                 conn_opts['interval_step'] = transport_opts['interval_step']
             if 'interval_max' in transport_opts:
                 conn_opts['interval_max'] = transport_opts['interval_max']
+            if 'timeout' in transport_opts:
+                conn_opts['timeout'] = transport_opts['timeout']                
         return conn_opts
 
     @property
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -278,10 +278,15 @@ class Connection(object):
                          *args, **kwargs)
 
     def connect(self):
-        """Establish connection to server immediately."""
-        return self._ensure_connection(
+        """Establish connection to server immediately, ignore transport options except timeout."""
+        kw_args = dict(
             max_retries=1, reraise_as_library_errors=False
         )
+        conn_opts = self._extract_failover_opts()
+        timeout = conn_opts.get("timeout", None)
+        if timeout is not None:
+            kw_args["timeout"] = timeout
+        return self._ensure_connection(**kw_args)
 
     def channel(self):
         """Create and return a new channel."""
@@ -477,7 +482,7 @@ class Connection(object):
 
     def ensure(self, obj, fun, errback=None, max_retries=None,
                interval_start=1, interval_step=1, interval_max=1,
-               on_revive=None):
+               connection_timeout=None, on_revive=None):
         """Ensure operation completes.
 
         Regardless of any channel/connection errors occurring.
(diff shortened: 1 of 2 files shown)


kombu/connection.py
```diff
@@ -844,6 +844,8 @@ def _extract_failover_opts(self):
                 conn_opts['interval_step'] = transport_opts['interval_step']
             if 'interval_max' in transport_opts:
                 conn_opts['interval_max'] = transport_opts['interval_max']
+            if 'timeout' in transport_opts:
+                conn_opts['timeout'] = transport_opts['timeout']
```
matusvalo, line 848: I prefer to use different name for 'timeout' in `transport_opts` This `timeout` value is overall timeout for all retries:
https://github.com/celery/kombu/blob/56e88dc275831196ca396fce81275ac687d1207b/kombu/utils/functional.py#L322
. We have also `connect_timeout` which specifies timeout of creating connection to single broker.
https://github.com/celery/kombu/blob/56e88dc275831196ca396fce81275ac687d1207b/kombu/connection.py#L81

If not good name is picked, it will be confusing for users.

matusvalo (changes requested):
Thank you @sschiessl-bcp for your PR. Here are my comments:
1. I prefer to adjust `timeout` name to something better as mentioned.
2. Could you add tests that paramters are used correctly in default_channel property?
3. I prefer to document this transport options

kombu/connection.py
sschiessl-bcp: I agree that it is confusing, because I was confused investigating :)

The current references to a timeout in connection.py:

1. https://github.com/celery/kombu/blob/56e88dc275831196ca396fce81275ac687d1207b/kombu/connection.py#L313: Why is there a reference in the documentation? Because it may be in kwargs? Would it be more suitable to reference transport docs for forwarded possible parameters?

2. https://github.com/celery/kombu/blob/56e88dc275831196ca396fce81275ac687d1207b/kombu/connection.py#L390: This references the same timeout as it is forwarded. 

So my suggestion would be to call it `connection_timeout` in the transport options. Thoughts?

sschiessl-bcp: I renamed it accordingly, add also added more functionality where I thought it is usefull to use the new option. Please check

matusvalo:
@sschiessl-bcp unfortunately your changes are breaking current API. Your changes are changing semantics of `connect()` method. Please keep the PR focused only on adding timeout option. Further changes needs to be discussed and breaking API changes needs to be introduced in major release. Hence I propose:
1. revert to the first commit and rename the parameter to something self explaining. I propose to use `default_channel_timeout` (basically it **is** a timeout of default_channel method)
2. add tests for this parameter
3. document the parameter (the best is also to document remaining parameters coming from `_extract_failover_opts`)

sschiessl-bcp:
@matusvalo I have not reacted 

Thank you for the comment @FrankK-1234 
https://github.com/celery/kombu/issues/1223#issuecomment-2180556495

I did check your references, and indeed https://github.com/celery/kombu/pull/1599 gives the same functionality, although with a different name (´connect_retries_timeout´). I do think it also adds to the confusion to the naming together with another change in `_extract_failover_opts`, as we have now 

- `self.connect_timeout` used in `_extract_failover_opts` for the `timeout` option
- `connect_retries_timeout` used in `transport_options` for the `timeout` option, which overrides `connect_timeout` if set
- `timeout` being the only documented wording in `_ensure_connection`

Ultimately, the functionality that I was seeking is now available which renders this PR pointless! Thanks for the discussion, will make a comment on the wording in the issue.

PR #1231, opened 2020-07-29
fix: raise BrokenPipeError exit on brokenpipe instead of hanging around

jagguli:
related to this https://github.com/celery/celery/pull/5581

 #5581

### DIFF
--- a/kombu/message.py
+++ b/kombu/message.py
@@ -129,6 +129,10 @@ class Message(object):
     def ack_log_error(self, logger, errors, multiple=False):
         try:
             self.ack(multiple=multiple)
+        except BrokenPipeError as exc:
+            logger.critical("Couldn't ack %r, reason:%r",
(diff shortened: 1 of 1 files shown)


auvipy:
py 2 test failing https://travis-ci.org/github/celery/kombu/jobs/712766430

auvipy:
mmmm, it's ok on kombu side, but can we do something on the open linked PR as well?

thedrow:
> mmmm, it's ok on kombu side, but can we do something on the open linked PR as well?

We should close it since it's definitely the wrong fix. I already explained this in https://github.com/celery/celery/pull/6528#issuecomment-744013988 and @maybe-sybr seconded in https://github.com/celery/celery/pull/6528#issuecomment-744087715.
We can have a ticket instead to figure out if we actually need to do something else.

PR #1232, opened 2020-07-30
Drop Python 2 Support

thedrow:
We now drop Python 2 support in Kombu as well.

After this is merged we must enter a merge freeze period.

Supersedes #990.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,8 +39,6 @@ jobs:
       dist: xenial
     - env: TOXENV=flake8
       stage: lint
-    - env: TOXENV=flakeplus
-      stage: lint
     - python: '3.8'
       env: TOXENV=apicheck
       stage: lint
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,19 +28,12 @@ _integration_job: &integration_job
 
 jobs:
   include:
-    - python: 2.7
-      env: TOXENV=2.7-unit
-    - python: 3.5
-      env: TOXENV=3.5-unit
     - python: 3.6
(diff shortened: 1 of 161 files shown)


auvipy:
what about another branch? say v5 like celery?

thedrow:
> what about another branch? say v5 like celery?

Since we're going to drop support for Python 2 soon I don't think it is needed.
We can maintain Kombu 4.x in a release branch.
What do you think?

auvipy:
> > what about another branch? say v5 like celery?
> 
> Since we're going to drop support for Python 2 soon I don't think it is needed.
> We can maintain Kombu 4.x in a release branch.
> What do you think?

a maintenance branch before merging this is fair enough. it would be good to hold this merge for this week. I will be in vacation throughout this week. didn't have any vacation or break from the last march Corona lockdown.

thedrow:
> > > what about another branch? say v5 like celery?
> > 
> > Since we're going to drop support for Python 2 soon I don't think it is needed.
> > We can maintain Kombu 4.x in a release branch.
> > What do you think?
> 
> a maintenance branch before merging this is fair enough. it would be good to hold this merge for this week. I will be in vacation throughout this week. didn't have any vacation or break from the last march Corona lockdown.

I'm not sure I can hold off the release but I can guarantee I'll be responsible and will regularly check the repositories for new issues.

Have fun during your vacation.

matusvalo:
Unfortunately, I don't have much time to review this PR. I will do it after next week.

thedrow:
@matusvalo I think it's fine and we can still release this.

PR #1245, opened 2020-09-02
Make `utils.functional.retry_over_time` appear in rendered docs

ashb:
Since this module has `__all__`, autodoc will (by default) only show docs for those
members, meaning retry_over_time is not shown in the rendered docs.

This function is used by the Redis result backend in Celery, and it has a doc
comment so it should be "public".


ashb:
This was discovered cos I wanted to link to it from the new docs in https://github.com/celery/celery/pull/6330

Issue #1201, opened 2020-05-30
pidbox.Mailbox._publish doesn't properly use default channel


tik-stbuehler:
It uses the passed (possible `None`) `channel` in `maybe_declare(self.reply_queue(channel))`, it probably should use `chan` instead.

https://github.com/celery/kombu/blob/5cc4a7688201f4d2e53f468734cc2e7cf177fc40/kombu/pidbox.py#L299-L302

tik-stbuehler:
Same problem in `_collect`:

https://github.com/celery/kombu/blob/5cc4a7688201f4d2e53f468734cc2e7cf177fc40/kombu/pidbox.py#L359-L361

auvipy:
what proposed improvement do you have in mind? can you come with a PR and test?

PR #1256, opened 2020-10-01
Fix pidbox not using default channels on 4.x branch

### LINKED ISSUE #1063: Table empty or key no longer exists

kombu==4.5.0
celery==4.3.0
redis==3.2.1

lambacck:
Not all cases are properly using the default channel even when detecting
for it.

This is directed at the 4.x branch because I'm using celery 4.4.7 and trying to see if this change fixes #1063 for me and I'm not ready to go to Celery 5.

Fixes: #1201


lambacck:
The second commit updates the dev requirements because the master for vine and py-amqp have dropped support for Python 2.7 and 3.5 while the Kombu 4.x branch continues to need to support those versions.

lambacck:
I rearranged commits to show tests going from green to red. I'm feeding the commits one at a time because that's what's required to get test runs to fire. Travis cancels a test run if I push a new commit so I'm waiting for the previous run to complete before I push the next commit.

lambacck:
The test failure in `Update dev requirements for 4.x branch` is windows build of brotli dependency is failing. Travis tests pass.

matusvalo:
Hi @lambacck. Thank you for your PR. I think to accept your PR, you need to create it against master.

.python-version
```diff
@@ -0,0 +1 @@
+3.6.11
```
matusvalo, line 1: What is purpose of this file? Is it needed in PR?

lambacck: nope that's an accidental add because of pyenv.

lambacck: I need it against 4.x because I'm not ready to move to celery 5

lambacck:
> Hi @lambacck. Thank you for your PR. I think to accept your PR, you need to create it against master.

I'm putting it against 4.x because I'm still using Celery 4 and am not yet moving to Celery 5. I'm happy to also open a PR against master with the equivalent changes.

lambacck:
Okay. I was finally able to get tests running locally on my mac so I think this time tests will be green everywhere but windows because install of brotli isn't working on windows tests. Once I see green here I'll also open a PR agains master with the 2 commits related to solving the issue (i.e. without the 4.x dependency pinning).

lambacck:
Okay, I think this is properly ready for review now. I'll go make the PR against master that's the equivalent.

requirements/dev.txt
```diff
@@ -1,2 +1,2 @@
-https://github.com/celery/py-amqp/zipball/master
-https://github.com/celery/vine/zipball/master
+https://github.com/celery/py-amqp/zipball/v2.6
```
auvipy, line 1: is it really need to be changed?

lambacck: master for those projects has dropped support for Python 2.7 and Python 3.5. This pins the dev-dependencies for the 4.x branch here to the branches in those projects so that kombu 4 (and thus Celery 4) can maintain support for python 2.7 and 3.5 in the spirit of the LTS branch declared with the release of Celery 5.

lambacck: Related, it seems that brotli on windows needs some kind of pinning to make it continue to work for python 2.7 but I didn't want to dig too hard into what that was. That's why appveyor tests are failing.

matusvalo:
Closing in favor of #1258. To backport this functionality to 4.x, I prefer to cherrypick commit from master.

lambacck:
That's what this branch is. The master changes cherry-picked. But also the 4.x branch won't pass tests until you do the version pinning things I did in the first commit (which aren't on master).

requirements/dev.txt
hsophie-sf: Seems like it will work with 1.0.7 based on this: https://github.com/google/brotli/issues/848

PR #1262, opened 2020-10-20
Fix 1063 on 4.x branch

### LINKED ISSUE #1063: Table empty or key no longer exists

hsophie-sf:
Fixes #1063 
Details on https://github.com/celery/kombu/issues/1063#issuecomment-704405396

First commit is a failing test which is fixed in second commit.


hsophie-sf:
Unable to build successfully because of two issues, both unrelated to my changes: 

1. Unable to build on python 2.7 because of lzma. The problem I'm getting is similar to: https://github.com/peterjc/backports.lzma/issues/22

> backports/lzma/_lzmamodule.c(32) : warning C4273: 'PyErr_NewExceptionWithDoc' : inconsistent dll linkage
          C:\projects\kombu\.tox\2.7\include\pyerrors.h(226) : see previous definition of 'PyErr_NewExceptionWithDoc'
  backports/lzma/_lzmamodule.c(115) : fatal error C1083: Cannot open include file: 'lzma.h': No such file or directory

2. Flaky test `t/unit/transport/test_redis.py::test_Redis.test_publish__consume` fails when 
run via  `tox -r -e 3.7-linux-unit`  but passes when 
run via `tox -r -e 3.7-linux-unit t/unit/transport/` or more specifically with filename, class, or test method name. 

Seems like it's using the test_redis.Client object which doesn't have `lock()` and other methods implemented. Don't know enough about the code to fix it.

The error is: 

```
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― test_Redis.test_publish__consume ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――

self = <t.unit.transport.test_redis.test_Redis object at 0x117317320>

    def test_publish__consume(self):
        connection = Connection(transport=Transport)
        producer = Producer(channel, self.exchange, routing_key='test_Redis')
        consumer = Consumer(channel, queues=[self.queue])
    
        producer.publish({'hello2': 'world2'})
        _received = []
    
        def callback(message_data, message):
            _received.append(message_data)
            message.ack()
    
        consumer.register_callback(callback)
        consumer.consume()
    
        assert channel in channel.connection.cycle._channels
        try:
>           connection.drain_events(timeout=1)

t/unit/transport/test_redis.py:971: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
kombu/connection.py:324: in drain_events
    return self.transport.drain_events(self.connection, **kwargs)
kombu/transport/virtual/base.py:963: in drain_events
    get(self._deliver, timeout=timeout)
kombu/transport/redis.py:381: in get
    self.maybe_restore_messages()
kombu/transport/redis.py:340: in maybe_restore_messages
    num=channel.unacked_restore_limit,
kombu/transport/redis.py:197: in restore_visible
    self.unacked_mutex_expire):
../../.pyenv/versions/3.7.3/lib/python3.7/contextlib.py:112: in __enter__
    return next(self.gen)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

client = <t.unit.transport.test_redis.Client object at 0x11733c080>, name = 'unacked_mutex', expire = 300

    @contextmanager
    def Mutex(client, name, expire):
        """Acquire redis lock in non blocking way.
    
        Raise MutexHeld if not successful.
        """
>       lock = client.lock(name, timeout=expire)
E       AttributeError: 'Client' object has no attribute 'lock'

kombu/transport/redis.py:114: AttributeError

 t/unit/transport/test_redis.py::test_Redis.test_publish__consume
```

thedrow:
So why are you backporting this?
Is this not relevant for master?

hsophie-sf:
> So why are you backporting this?
> Is this not relevant for master?

We just upgraded to the 4.x branch of Celery and noticed this issue. Since 4.x is LTS, we want it fixed there.
I can also submit a PR against master as well, but would like this fixed on 4.x.

lambacck:
@thedrow you locked #1063 ( https://github.com/celery/kombu/issues/1063#issuecomment-787428515 ) and asked for a PR that reproduces and fixes it. But the PR already exists and is this one.

auvipy:
in any case this should target master branch. in kombu we do not have proper 4.6.x branch like in celery

requirements/extras/brotli.txt
```diff
@@ -1,2 +1,2 @@
 brotlipy>=0.7.0;platform_python_implementation=="PyPy"
-brotli>=1.0.0;platform_python_implementation=="CPython"
+brotli==1.0.7;platform_python_implementation=="CPython"
```
thedrow, line 2: This should not be changed at all.

requirements/dev.txt
```diff
@@ -1,2 +1,2 @@
-https://github.com/celery/py-amqp/zipball/master
-https://github.com/celery/vine/zipball/master
+https://github.com/celery/py-amqp/zipball/v2.6
```
thedrow, line 1: This should not be changed at all.

kombu/pidbox.py
```diff
@@ -284,6 +292,14 @@ def _publish_reply(self, reply, exchange, routing_key, ticket,
                     }, retry=True,
                     **opts
                 )
+            except OperationalError as exc:
+                # Fixes https://github.com/celery/kombu/issues/1063
+
+                if not exc.args and not is_no_route_error_for_reply_celery_pidbox(exc.args[0]):
```
thedrow, line 298: So what exactly happens here?
Why reporting this error is good enough?

thedrow (changes requested):

thedrow:
> 
> 
> in any case this should target master branch. in kombu we do not have proper 4.6.x branch like in celery

@hsophie-sf You should target master anyway. I've done so for you.
We will backport this fix ourselves.

t/unit/test_pidbox.py
```diff
@@ -58,6 +60,26 @@ def test_publish_reply_ignores_InconsistencyError(self):
...
+        mailbox.reply_queue(channel).declare()
+        ticket = uuid()
+        # Using Channel._lookup as a proxy (in absence of a redis integration test) to mock the actual
+        # redis.get_table call which produces the OperationalError, since _lookup is just a level above
+        # redis.get_table on the stack trace.
+        with patch.object(kombu.transport.virtual.Channel, '_lookup') as simulate_redis_get_table_err:
+            # raise the actual redis error
+            simulate_redis_get_table_err.side_effect = OperationalError(
+                redis.NO_ROUTE_ERROR.format(exchange, mailbox.oid))
+            try:
```
thedrow, line 76: Please use `pytest.raises()`.

thedrow:
I'm pretty convinced that this is the wrong place to fix this.
This is a good workaround but the actual fix should be in the Redis transport as it only happens there right?
Also, I don't understand why this error happens yet.

requirements/dev.txt
hsophie-sf: I wasn't able to build locally using tox without this, if I remember correctly.

requirements/extras/brotli.txt
hsophie-sf: I wasn't able to build locally using tox without this, if I remember correctly.

requirements/dev.txt
hsophie-sf: Reverting since unittest is failing on latest commit.

requirements/extras/brotli.txt
hsophie-sf: I'll revert this as well in anticipation of another test failing.

t/unit/test_pidbox.py
hsophie-sf: We're trying to assert that the particular Error is NOT raised now, so can't use `pytest.raises()`.

hsophie-sf:
CI is failing due to unrelated reason now: https://github.com/hsophie-sf/kombu/runs/2667819031?check_suite_focus=true on test `t/unit/transport/test_redis.py::test_Redis::test_publish__consume`

auvipy:
@matusvalo

auvipy (changes requested):
please  rebase

matusvalo:
I agree with @thedrow about this PR:

> I'm pretty convinced that this is the wrong place to fix this. @lambacck Your analysis is 100% correct though.
> This is a good workaround but the actual fix should be in the Redis transport as it only happens there right?
> Also, I don't understand why this error happens yet.

Moreover I miss some kind of warning when it happens.

matusvalo (changes requested):
I agree with @thedrow feedback:

> I'm pretty convinced that this is the wrong place to fix this. @lambacck Your analysis is 100% correct though.
> This is a good workaround but the actual fix should be in the Redis transport as it only happens there right?
> Also, I don't understand why this error happens yet.

Moreover, I miss kind of warning when the exception is ignored.

matusvalo:
Closing in favor of #1394 and #1404

Users affected by #1063 can do hotfix of their environment based on following comment:

https://github.com/celery/kombu/issues/1063#issuecomment-957236477

auvipy:
5.2.0

PR #1264, opened 2020-10-24
Reformat and improve kombu transport reference docs

matusvalo:
This PR adds standard format of transport reference documentation. It also adds missing details like transport options etc... This PR leaves some transport with some information missing (e.g. features for azure services connection string for SQS, ...). This information should be added in separate PR by someone who is using this transports.

Each transport now contains section for:
* Features - which features transport supports
* Connection string - the structure of connection string
* transport options - which transport options transport supports


maybe-sybr (approved):
lgtm

Would it be better to put some placeholder in the empty feature support fields rather than leaving them blank? Something like `*Unreviewed*` perhaps.

PR #1268, opened 2020-11-06
Remove ca_certs for pyamqp

### LINKED ISSUE #1266: kombu still refres to ca_certs ssl param but that param is removed from py-amqp
The py-amqp dependency was bumped from 2.6 to 5.0. The py-amqp modul has changed in [1], the ca_certs parameter is removed from the ssl parameters. But the kombu documentation[2] and test[3] still refers to this parameter.

As far as I see this is not a functional bug in kombu just a doc and test improvement.

[1] https://github.com/celery/py-amqp/commit/ad9697ab2eaa97e153211c82e4bad8d655e63591#diff-8bf65e6879a19d14c2d02835dc148a1b045e2d1c56fa13c6fd89de2426e54d03L339
[2] https://github.com/celery/kombu/blob/master/kombu/connection.py#L64
[3]https://github.com/celery/kombu/blob/master/t/unit/transport/test_pyamqp.py#L94

4383:
This param was recently removed in py-amqp (https://github.com/celery/py-amqp/pull/327/), these changes update related documentation and test.

Will fix #1266

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -61,7 +61,6 @@ class Connection:
 
         >>> import ssl
         >>> Connection('amqp://', login_method='EXTERNAL', ssl={
-        ...    'ca_certs': '/etc/pki/tls/certs/something.crt',
         ...    'keyfile': '/etc/something/system.key',
         ...    'certfile': '/etc/something/system.cert',
         ...    'cert_reqs': ssl.CERT_REQUIRED,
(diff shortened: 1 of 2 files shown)


thedrow:
Since we're fixing it in py-amqp I'm going to close this PR.
Thanks for the contribution anyway.

PR #1280, opened 2020-12-06
Add fanout support to filesystem transport

yhal-nesi:
I am are trying to use celery with filesystem transport, but it doesn't support control commands without the fanout. This patch implements the required functionality. 

* It requires control_folder defined and created. Could probably make it a subfolder of "in" or "out" for compatibility. Or make fanout support conditional on its existence? 
* Exchange bindings are never deleted. Is there any function in Channel class for that? 
* _put_fanout doesn't make use of the routing key. Not sure if this is needed for some functionality? The documentation for FanoutExchange doesn't say anything about this method but celery control doesn't work without it. 

I've also made it create a directory for each queue, so that one huge queue like celeryev doesn't slow everything down.

### DIFF
--- a/kombu/transport/filesystem.py
+++ b/kombu/transport/filesystem.py
@@ -258,8 +258,6 @@ class Channel(virtual.Channel):
 
     def _purge(self, queue):
         """Remove all messages from `queue`."""
-        count = 0
-        queue_find = '.' + queue + '.msg'
         queue_folder = os.path.join(self.data_folder_out, queue)
 
         try:
--- a/kombu/transport/filesystem.py
+++ b/kombu/transport/filesystem.py
@@ -65,7 +65,7 @@ Features
 * Type: Virtual
 * Supports Direct: Yes
 * Supports Topic: Yes
-* Supports Fanout: No
+* Supports Fanout: Yes
 * Supports Priority: No
 * Supports TTL: No
 
@@ -94,6 +94,7 @@ import uuid
 from queue import Empty
 import tempfile
 from time import monotonic
+from collections import namedtuple
 
 from . import virtual
 from kombu.exceptions import ChannelError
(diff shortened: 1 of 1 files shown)


matusvalo:
Hi @yhal-nesi. Thank you for your PR. I will review it soon

matusvalo (changes requested):
Here is my first review of the code:
1. Please make sure that flake8 is happy
2. Please update the doc string of the module. PR Adds additonal Transport Options which are not mentioned there. Also some high level information about directory structure would be great (even the control files)
3. please update the unittests with the tests of newly added code.

kombu/transport/filesystem.py
```diff
@@ -148,43 +149,94 @@ def unlock(file):       # noqa
...
+            
+            lock(d,LOCK_EX)
+            if os.path.isfile(filename):
+                f = open(filename,'r')
+                exchange_table = loads(bytes_to_str(f.read()))
+                f.close()
+            else:
+                exchange_table = []
+            queues = [tuple(q) for q in exchange_table]
+            queue_val = (routing_key or '', pattern or '', queue or '' )
```
matusvalo, line 186: Please use namedtuple here. Define it as global data structure.

kombu/transport/filesystem.py
```diff
@@ -148,43 +149,94 @@ def unlock(file):       # noqa
...
+            return result
+        except FileNotFoundError:
+            return []
+        except OSError:
+            raise ChannelError(
+                f'Cannot open {filename}')
+        finally:
+            f.close()
+
+    def _queue_bind(self, exchange, routing_key, pattern, queue):
```
matusvalo, line 170: Can we call `get_table()` function instead of duplicating of the code?

kombu/transport/filesystem.py
```diff
@@ -148,43 +149,94 @@ def unlock(file):       # noqa
...
+        filename = '{}.exchange'.format(exchange)
+        filename = os.path.join(self.control_folder,filename)
+        try:
+            d_fileno = os.open(self.control_folder,os.O_RDONLY)
+            d = namedtuple('Directory',[])
+            d.fileno = lambda: d_fileno
```
matusvalo, line 176: Can we define structure/adapter in global scope with proper comments?

matusvalo (commented):
Regarding your questions:

> It requires control_folder defined and created. Could probably make it a subfolder of "in" or "out" for compatibility. Or make fanout support conditional on its existence?

I think that we can introduce new directory, but it needs to be documented in module

> Exchange bindings are never deleted. Is there any function in Channel class for that?

It would be great to support also unbinging. There is method for that:
https://github.com/celery/kombu/blob/d62ba3cd96a73a4440c4a62d1e575dee82b5ba9e/kombu/transport/virtual/base.py#L570

> _put_fanout doesn't make use of the routing key. Not sure if this is needed for some functionality? The documentation for FanoutExchange doesn't say anything about this method but celery control doesn't work without it.

To be honest, I cannot answer this question right now. In general, fanout should not need routing key.

auvipy:
fixed in another PR

PR #1283, opened 2020-12-14
Allow specifying session token

flying-sheep:
This is useful for local testing using something like `gimme-aws-creds`


auvipy:
should this change also need mentions in docs?

flying-sheep:
Oh, probably!

We might also want to think about how to mention it, because it’s not something for production code – a session token is only valid for a while, so this is for testing your code locally, while you’d be using roles on AWS.

matusvalo:
@flying-sheep I think at least you can comment the method itself and also you can put a note to docstring of the module.

flying-sheep:
done!

PR #1284, opened 2020-12-17
Support for Azure Service Bus 7.0.0

### LINKED ISSUE #1279: Fix azure service bus integration with service bus library 7.0
Azure Service Bus python library has been upgraded from 0.53 to 7.0 with breaking api changes

version 0.5 is not maintained anymore

See Migration Documentation : [https://github.com/Azure/azure-sdk-for-python/tree/master/sdk/servicebus/azure-servicebus/migration_guide.md](https://github.com/Azure/azure-sdk-for-python/tree/master/sdk/servicebus/azure-servicebus/migration_guide.md)

terrycain:
Azure have completely changed the Azure ServiceBus SDK between 0.50.0 and 7.0.0. It didn't make sense to me to support the old one anymore hence the rather large refactor.

By default ASB will require you to ack messages, so have tried to incorporate that functionality. Also now for each queue the SDK gives you a sender and receiver object which behind the scenes each creates a thread, which one should not forget to close.

The PR is not finished but seems to work in some basic examples. I'm planning to clean up the implementation a bit, but would like some feedback on this as I'm not massively familiar with the Kombu codebase at the moment.

Fixes #1279

### DIFF
--- a/kombu/transport/azureservicebus.py
+++ b/kombu/transport/azureservicebus.py
@@ -35,27 +35,25 @@ Transport Options
 * ``queue_name_prefix``
 * ``wait_time_seconds``
 * ``peek_lock``
+* ``peek_lock_seconds``
 """
 
 import string
 from queue import Empty
+from typing import Dict, Any, Optional, Union, Set
 
 from kombu.utils.encoding import bytes_to_str, safe_str
 from kombu.utils.json import loads, dumps
 from kombu.utils.objects import cached_property
 
-from . import virtual
+import azure.core.exceptions
+import azure.servicebus.exceptions
+import isodate
+from azure.servicebus import ServiceBusClient, ServiceBusMessage, ServiceBusReceiver, ServiceBusSender, \
+    ServiceBusReceiveMode
+from azure.servicebus.management import ServiceBusAdministrationClient
 
-try:
-    # azure-servicebus version <= 0.21.1
-    from azure.servicebus import ServiceBusService, Message, Queue
-except ImportError:
-    try:
-        # azure-servicebus version >= 0.50.0
-        from azure.servicebus.control_client import \
-            ServiceBusService, Message, Queue
-    except ImportError:
-        ServiceBusService = Message = Queue = None
+from . import virtual
 
 # dots are replaced by dash, all other punctuation replaced by underscore.
 CHARS_REPLACE_TABLE = {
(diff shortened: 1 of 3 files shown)


kombu/transport/azureservicebus.py
```diff
@@ -35,122 +35,260 @@
...
-        ServiceBusService = Message = Queue = None
+from . import virtual
 
 # dots are replaced by dash, all other punctuation replaced by underscore.
 CHARS_REPLACE_TABLE = {
     ord(c): 0x5f for c in string.punctuation if c not in '_'
 }
 
 
+class SendReceive(object):
```
auvipy, line 64: explicit inheritance from the object is not needed as we are py 3 only now

auvipy (commented):
thanks for handling this, overall direction is right for this PR

terrycain:
@auvipy Mind reviewing this when you get a chance :smiley:

matusvalo (approved):

matusvalo:
Thank you @terrycain

terrycain:
Thanks @matusvalo, any idea when we can get this released? 😃

AndreCimander:
@terrycain thanks for creating this MR. @matusvalo FYI

I'm running into issues with the connection string, due to the key containing a slash. The old implementation allowed/required an encoded slash, the new implementation now fails with an `InvalidSignature` exception, but using a non-escaped slash in the key results in the Celery worker failing to startup:

> Unrecoverable error: ValueError(\"Port could not be cast to integer value as '[partbeforeslash]'\")", "exc_info": "Traceback (most recent call last):\n  File \"/usr/local/lib/python3.8/site-packages/celery/worker/worker.py\", line 203, in start\n    self.blueprint.start(self)\n  File \"/usr/local/lib/python3.8/site-packages/celery/bootsteps.py\", line 112, in start\n    self.on_start()\n  File \"/usr/local/lib/python3.8/site-packages/celery/apps/worker.py\", line 136, in on_start\n    self.emit_banner()\n  File \"/usr/local/lib/python3.8/site-packages/celery/apps/worker.py\", line 170, in emit_banner\n    ' \\n<file://n>', self.startup_info(artlines=not use_image))),\n  File \"/usr/local/lib/python3.8/site-packages/celery/apps/worker.py\", line 231, in startup_info\n    conninfo=self.app.connection().as_uri(),\n  File \"/usr/local/lib/python3.8/site-packages/kombu/connection.py\", line 672, in as_uri\n    connection_as_uri = maybe_sanitize_url(connection_as_uri)\n  File \"/usr/local/lib/python3.8/site-packages/kombu/utils/url.py\", line 118, in maybe_sanitize_url\n    return sanitize_url(url, mask)\n  File \"/usr/local/lib/python3.8/site-packages/kombu/utils/url.py\", line 111, in sanitize_url\n    return as_url(*_parse_url(url), sanitize=True, mask=mask)\n  File \"/usr/local/lib/python3.8/site-packages/kombu/utils/url.py\", line 76, in url_to_parts\n    parts.port,\n  File \"/usr/local/lib/python3.8/urllib/parse.py\", line 174, in port\n    raise ValueError(message) from None\nValueError: Port could not be cast to integer value as '[partbeforeslash]'", "service": "CeleryMiddleware", "level": "CRITICAL", "timestamp": "2021-01-05T12:14:42Z", "payload": null, "event_id": null, "message_id": null, "request_id": null, "logger": "celery.worker", "filename": "worker.py:207", "fn": "start"}


Not really your fault, since the code base of Celery, imho, is quite a mess. Might be useful to re-implement the support for encoded slashes?

AndreCimander:
Another question: Does this implementation support non-exchange-bound Queues? On first glance I tend to say no, since the the queue name construction on line 240 seems to only look at the prefix and the exchange?

terrycain:
@AndreCimander yeah thats probably my fault, I tried to be smart and not require an encoded slash. Let me have a play around, but yeah I think we'll need to revert to requiring encoded slashes though that would be annoying.

Am not super familiar with kombu, but your assessment of line 240 makes sense to me.

matusvalo:
@terrycain it would be great if you can create PR for it. Please when done, ping me and I will merge it.

terrycain:
@matusvalo will do, I think there is some more bits here I want to fix first, will make a pr for a release after 👍

thedrow:
Is there an issue to track this? @terrycain @matusvalo 
I was about to release a new Kombu 5.1 because of this one.

terrycain:
@thedrow no issue apart from the one referenced in this pr, but probably want to hold off till #1287 is done

olros:
@thedrow Do you have an update on when this could get released? 😄

thedrow:
I have a family emergency so the answer is currently no.

PR #1297, opened 2021-02-22
Stop treating namedtuple as an object when using simplejson

Hardtack:
By default, [simplejson converts namedtuple to JSON object](https://simplejson.readthedocs.io/en/latest/#simplejson.dumps) but json converts that to JSON array. 

This causes unpredictable serialization depends on installation of simplejson. So I added an option for disabling simplejson specific feature.


thedrow:
Why would anyone still use simplejson?

matusvalo (approved):
Kombu prefers simplejson when installed. So I think we should fix issues coming from it. If using simplejson is not having benefits we can drop it later on.

PR #1299, opened 2021-02-25
added initial draft github action workflow

auvipy:
initial experiments

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -39,12 +39,12 @@ jobs:
         ports:
           - 5432:5432
         # set health checks to wait until service has started
-        options: >-
-          --health_cmd = "rabbitmq-diagnostics ping -q"
-          --health_interval = 10
-          --health_timeout = 10
-          --health_retries = 30
-          --health_start_period = 5
+        #options: >-
+         # --health_cmd = "rabbitmq-diagnostics ping -q"
+         # --health_interval = 10
+         # --health_timeout = 10
+         # --health_retries = 30
+         # --health_start_period = 5
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -1,0 +1,88 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Kombu
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+env:
+  GITHUB_WORKFLOW: true
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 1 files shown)


auvipy:
Error: Container operations are only supported on Linux runners -- so integration tests with docker need Linux only

.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,88 @@
...
+  GITHUB_WORKFLOW: true
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04, macos-latest]
```
auvipy, line 22: macos & windows can run unit tests and container based integration can run on llinux only, so separate jobs should be created

auvipy:
related https://github.com/celery/py-amqp/pull/358

.github/workflows/workflow.yaml
thedrow: See this for a solution on MacOS:
https://github.com/actions/virtual-environments/issues/1143#issuecomment-652264388

.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,88 @@
...
+          ${{ matrix.python-version }}-v1-
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install flake8 pytest case pytest-celery pytest-subtests pytest-timeout pytest-cov
+        python -m pip install moto boto3 msgpack PyYAML
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: Run Unit test with tox-docker
+      run: |
+        PYTHONPATH=. pytest -xv --cov=celery --cov-report=xml --cov-report term t/unit
```
thedrow, line 76: Please just use tox since that's what everyone uses when cloning the project.
We need to ensure it functions perfectly.

thedrow (changes requested):

auvipy:
I am a little bit sick. I saw PR of @matusvalo in pyamqp, which I can take as a suggested way. @matusvalo you can push on CI branch as well if got time.

matusvalo:
@auvipy I was checking your branch but your approach is totally different from my PR so I decided to create separate new PR of it. (there is no benefit integrating it in your PR since I would overwrite whole file). Though, your PR contains some great stuff (like coverage, cache etc). Maybe after accepting basic PR of CI you could create PR for adding them to it? (I would like to have simple running CI ASAP, since we introduced some issues/failing unit tests due not having CI running)

auvipy:
> @auvipy I was checking your branch but your approach is totally different from my PR so I decided to create separate new PR of it. (there is no benefit integrating it in your PR since I would overwrite whole file). Though, your PR contains some great stuff (like coverage, cache etc). Maybe after accepting basic PR of CI you could create PR for adding them to it? (I would like to have simple running CI ASAP, since we introduced some issues/failing unit tests due not having CI running)

OK

thedrow:
Should we close this then?

matusvalo:
I think yes. But we can use some parts of this PR which are missing in current CI (cache, covetage etc)

Issue #1181, opened 2020-04-26
SQS: JSONDecodeError when reading message from Amazon SQS predefined_queues


januszm:
When reading messages from SQS using `predefined_queues` (messages come from SNS) I see the following JSON decode error in:

```python
# kombu/transport/SQS.py", line 290
def _message_to_python(self, message, queue_name, queue):
    try:
        body = base64.b64decode(message['Body'].encode())
    except TypeError:
        body = message['Body'].encode()
    payload = loads(bytes_to_str(body)) # <<< HERE line 290
```

Afet dropping `rdb.set_trace()` around those lines I can see that the message looks of course just like regular AWS SNS/SQS message:

```python
# simplified version for clarity
{
  'MessageId': 'abcd-efgh',
  'ReceiptHandle': 'abcdefgh==',
  'MD5OfBody': 'abc123',
  'Body': '''{\n  
    "Type" : "Notification",\n  
    "MessageId" : "abcd-efgh",\n  
    "TopicArn" : "arn:aws:sns:xxx",\n  
    "Subject" : "Amazon S3 Notification",\n  
    "Message" : "{\\"Records\\":[{\\"eventVersion\\":\\"2.1\\",\\"eventSource\\":\\"aws:s3\\",...}"
  }'''
}
```

The `try: body = ` block works fine, it'll produce bytestring:
```
(Pdb) base64.b64decode(message['Body'].encode())
b'O*^6\x8ab~\'\x1c\xd6*\'1
```

but then it errors in line `payload = loads(bytes_...`

```
(Pdb) payload = loads(bytes_to_str(body))
*** json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
(Pdb) bytes_to_str(body)
'O*^6�c~\'\x1b�*\
```

It looks like for messages coming from 'predefined_queues` (from other AWS publishers) we should simply:

```python
loads(bytes_to_str(message['Body'].encode()))
# or even better
loads(message['Body'])
```

januszm:
UPDATE: I found this answer at StackOverflow: https://stackoverflow.com/a/57389394/470512
> The SQS support in Celery is only intended as a transport mechanism for Celery-specific messages. You cannot use Celery to consume arbitrary SQS messages.

> Instead, I'd suggest writing a custom Django management command in which you poll your SQS queue using the boto3 library.

[...]

PR #1306, opened 2021-03-06
fix: non kombu json message decoding in SQS transport

moaddib666:
Fix the case with custom messages decoding in SQS transport
Use cases:
1) Celery with custom consumer
2) PUB/SUB with Custom publisher [SNS+SQS]
3) Non kombu/celery protocols

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -370,11 +370,21 @@ class Channel(virtual.Channel):
         else:
             c.send_message(**kwargs)
 
-    def _message_to_python(self, message, queue_name, queue):
+    @staticmethod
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
can you also add a test for the changes?

moaddib666:
> can you also add a test for the changes?

Hi, sure.

moaddib666:
Updated. Also, I've interested in backporting following changes to the celery 4.* as well, please advise what should be done for it ?

moaddib666:
Also looks like i would fix the issue described above https://github.com/celery/kombu/issues/1181#issue-607011906

auvipy (approved):

PR #1311, opened 2021-03-11
Update connection.py

Krogsager:
Based on my headaches with silent revert to `localhost` I submit this PR.
The developer should be notified if their host settings are not found.
Details on the issue are here: https://github.com/celery/celery/issues/6661
This is just a suggestion and I will not argue if it is rejected.


auvipy:
I think this could be a good addition

thedrow:
Thanks! It's a good idea.

PR #1314, opened 2021-03-14
Update default pickle protocol version to 4

### LINKED ISSUE #1070: Cannot serialize a string larger than 4GiB using pickle
Hi, looks like there is a new protocol ie 4 for supporting large files. Can someone send a PR for this.

illia-v:
Protocol version 4 was added in Python 3.4 and is supported by all Python versions that kombu supports.

https://docs.python.org/3.9/library/pickle.html#data-stream-format

Fixes #1070

### DIFF
--- a/kombu/serialization.py
+++ b/kombu/serialization.py
@@ -35,9 +35,8 @@ else:
 pickle = cpickle or pypickle
 pickle_load = pickle.load
 
-#: Kombu requires Python 2.5 or later so we use protocol 2 by default.
-#: There's a new protocol (3) but this is only supported by Python 3.
-pickle_protocol = int(os.environ.get('PICKLE_PROTOCOL', 2))
+#: We have to use protocol 4 until we drop support for Python 3.6 and 3.7.
+pickle_protocol = int(os.environ.get('PICKLE_PROTOCOL', 4))
 
 codec = namedtuple('codec', ('content_type', 'content_encoding', 'encoder'))
 


auvipy (approved):
I think we are good

clokep:
This might break things during an upgrade though, correct?

illia-v:
> This might break things during an upgrade though, correct? (Or are pickle versions forward compatible?)

Previously pickled objects should be read without a problem because the protocol version is encoded in them

clokep:
> > This might break things during an upgrade though, correct? (Or are pickle versions forward compatible?)
> 
> Previously pickled objects should be read without a problem because the protocol version is encoded in them

Makes sense! (I was concerned about a rolling deploy where you might have producers and consumers on different kombu versions. 👍 )

illia-v:
> I was concerned about a rolling deploy where you might have producers and consumers on different kombu versions.

In this case it may break things if Python is upgraded from version less than 3.4 at the same time

thedrow (approved):

thedrow:
> 
> 
> > I was concerned about a rolling deploy where you might have producers and consumers on different kombu versions.
> 
> In this case it may break things if Python is upgraded from version less than 3.4 at the same time

We'll make sure to mention this in the release notes. @auvipy

auvipy:
> > > I was concerned about a rolling deploy where you might have producers and consumers on different kombu versions.
> > 
> > 
> > In this case it may break things if Python is upgraded from version less than 3.4 at the same time
> 
> We'll make sure to mention this in the release notes. @auvipy

fore sure sire O:)

PR #1316, opened 2021-03-15
Use a thread-safe implementation of cached_property

### LINKED ISSUE #1303: kombu.utils.objects.cached_property is not thread safe.
`cached_property` is heavily used in celery but it is causing issues in multi-threaded code since it is not thread safe. We need to change implementation to store data in thread local storage. 
See following PRs/issues as example: https://github.com/celery/celery/issues/4363, https://github.com/celery/celery/pull/5145, https://github.com/celery/celery/pull/6416
There are multiple ways how we can do it:
1. Fix implementation to store data in thread local storage
2. Use 3rd party component - https://github.com/pydanny/cached-property
3. Use std lib implementation https://docs.python.org/3.9/library/functools.html?highlight=s#functools.cached_property. But this requires python >3.8 and needs to be checked whether is thread-safe.

thedrow:
Fixes #1303.


thedrow:
It turns out that our cached_property implementation needs to support a setter as well.

https://github.com/celery/celery/blob/8c5e9888ae10288ae1b2113bdce6a4a41c47354b/celery/app/amqp.py#L569-L576

auvipy (dismissed):

kombu/utils/objects.py
```diff
@@ -1,64 +1,54 @@
...
-        obj.__dict__[self.__name__] = value
 
-    def __delete__(self, obj, _sentinel=object()):
-        if obj is None:
+        with self.lock:
+            if self.__set is not None:
+                value = self.__set(instance, value)
+
+            cache = instance.__dict__
+            cache[self.attrname] = value
```
xirdneh, line 38: **[question]**: Why do we have to override `set`? `functools.cached_property` and `cached_property` both implement only `get` to calculate the value only once.

Do we really need to re-define the value on cached  properties?

xirdneh (changes requested):

kombu/utils/objects.py
thedrow: Because that's how it's implemented and used in Celery.

matusvalo (changes requested):
I am not sure whether this PR is fixing the issue. The problem of cached_property in celery is that the cache is **shared** between the threads. Current PR is adding just locks but it lacks storing the cache in thread local storage.

```python
>>> from kombu.utils import cached_property
>>> from threading import Thread
>>>
>>> class Foo:
...     @cached_property
...     def foo(self):
...         print('foo')
...         return 5
...
>>> foo = Foo()
>>>
>>> def call_foo():
...     foo.foo
...
>>>
>>> Thread(target=call_foo).start()
foo
>>> foo.foo        # Here we should see `foo` printed out since the cache should not be hit since it is called from different thread than previous call.
5
````

File: kombu/utils/objects.py [hunk 2]
```diff
@@ -1,64 +1,54 @@
...
-class cached_property:
-    """Cached property descriptor.
+try:
+    from functools import _NOT_FOUND, cached_property as _cached_property
+except ImportError:
+    # TODO: Remove this fallback once we drop support for Python < 3.8
+    from cached_property import threaded_cached_property as _cached_property
 
-    Caches the return value of the get method on first call.
+    _NOT_FOUND = object()
```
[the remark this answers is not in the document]
thedrow: But if we inherit from the stdlib `cached_property` we can't use our own sentinel, right?
It's already using `_NOT_FOUND`.

kombu/utils/objects.py
thedrow: @xirdneh Any thoughts?

thedrow:
Wait, we need thread-local storage instead of using locks?
Is this always the case?

matusvalo:
To be honest it is difficult to say but in Celery there is a lot of usage of `cached_property` and at least sometimes it is causing that threads are using global resource which should be thread-local see following comment: https://github.com/celery/celery/pull/5145#issuecomment-786830557 or my previous fix: https://github.com/celery/celery/pull/6416.

Rationale is following:
1. if celery is using multiprocessing parallelisation, each process has 1 thread storing in thread local storage => this should be equal to current state when it is stored as global variable. So moving to thread-local storage **should not** break this scenario.
2. if celery is using multithreading parallelisation, each thread will have it's own resource which should fix problems when there is resource shared which should not shared.

matusvalo:
I was thinking more about `cachedproperty` and I have changed my mind. cachedproperty is storing cached information to the instance of the method. This is correct behavior and we should keep it that way. In celery, we have mainly problem when there is one singleton shared across threads and it contains methods with `cachedproperty` decorator. This is dangerous because we can end up using single resource with multiple thread - see celery/celery#6416..

So I am aproving this PR. @thedrow the only change I am suggesting is adding docstring to the `cachedproperty` class explaning where are cached data stored - in the instance.

matusvalo (approved):

thedrow:
We can add a version of cached_property which is thread-local later on.

PR #1318, opened 2021-03-18
sqla 1.4 migration initialization

auvipy:


thedrow (changes requested):
Please maintain backwards compatibility as it is still possible to do so.

PR #1324, opened 2021-03-24
Azure Service Bus - versatile queue names fix

### LINKED ISSUE #1323: Azure Service Bus transport - queue name prefix with dashes not allowed
## Description
It's not possible to use dashes (`-`) in queue name prefix when using Azure Service Bus transport layer.

## Reproduction scenario:

* pass `queue_name_prefix` into `broker_transport_options` setting dict with values `some-queue-prefix-`

## Expected behaviour:

* celery tries to connect with `some-queue-prefix-celery` queue

## Current behaviour:

* celery tries to connect with `some_queue_prefix_celery`

Issue is blocking while ASB follows naming conventions that uses dashes - which is quite popular

tsunday:
## Fixes #1323 

In current implementation it is not possible to use dashes (`-`) in queue names. All punctuations are converted into underscores (`_`). Although the old commentary text says that dots are replaced by dashes - it wasn't working this way . 

Provided fix allows to use dashes in queue names, translates dots into dashes and translates other punctuations to underscores - just like specified in the comment above the translation table.

## How to verify:
to test/investigate the issue you have to pass `queue_name_prefix` into `broker_transport_options` setting dict on celery.

### DIFF
--- a/kombu/transport/azureservicebus.py
+++ b/kombu/transport/azureservicebus.py
@@ -66,9 +66,10 @@ from azure.servicebus.management import ServiceBusAdministrationClient
 from . import virtual
 
 # dots are replaced by dash, all other punctuation replaced by underscore.
+PUNCTUATIONS_TO_REPLACE = set(string.punctuation) - {'_', '.', '-'}
 CHARS_REPLACE_TABLE = {
     ord('.'): ord('-'),
-    **{ord(c): ord('_') for c in string.punctuation if c not in ['_', '.', '-']}
+    **{ord(c): ord('_') for c in PUNCTUATIONS_TO_REPLACE}
 }
 
 
--- a/kombu/transport/azureservicebus.py
+++ b/kombu/transport/azureservicebus.py
@@ -67,7 +67,8 @@ from . import virtual
 
 # dots are replaced by dash, all other punctuation replaced by underscore.
 CHARS_REPLACE_TABLE = {
-    ord(c): 0x5f for c in string.punctuation if c not in '_'
+    ord('.'): ord('-'),
+    **{ord(c): ord('_') for c in string.punctuation if c not in ['_', '.', '-']}
 }
 
 
--- a/t/unit/transport/test_azureservicebus.py
+++ b/t/unit/transport/test_azureservicebus.py
@@ -278,8 +278,11 @@ def test_purge(mock_queue: MockQueue):
 
 
 def test_custom_queue_name_prefix():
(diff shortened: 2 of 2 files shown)


auvipy (changes requested):
can you also update the tests?

auvipy:
@terrycain can you check

tsunday:
@auvipy, sure I'll add some tests regarding queue name creation with prefix

kombu/transport/azureservicebus.py
```diff
@@ -67,7 +67,8 @@
 
 # dots are replaced by dash, all other punctuation replaced by underscore.
 CHARS_REPLACE_TABLE = {
-    ord(c): 0x5f for c in string.punctuation if c not in '_'
+    ord('.'): ord('-'),
+    **{ord(c): ord('_') for c in string.punctuation if c not in ['_', '.', '-']}
```
thedrow, line 71: We can precalculate the punctuations we need to replace.
This will make the module's load time slightly faster and it is also cleaner.

```python
PUNCTUATIONS_TO_REPLACE = set(string.punctuation) - {'_', '.', '-'}

CHARS_REPLACE_TABLE = {
  ord('.'): ord('-'),
  **{ord(c): ord('_') for c in PUNCTUATIONS_TO_REPLACE}
}
```

thedrow (changes requested):
As mentioned, this needs tests.
I also have a comment regarding the code.

kombu/transport/azureservicebus.py
terrycain: This looks good, should work great.

tsunday: Thanks for suggestion, I'll change it accordingly. And will add tests 😀

tsunday:
@auvipy, I have just provided tests for a queue name generation

kombu/transport/azureservicebus.py
tsunday: @thedrow, I've changed the flow to your suggestions, I also added tests :)

auvipy:
fix the lint errors to trigger unit tests

thedrow:
> 
> 
> fix the lint errors to trigger unit tests

Please check the build log next time. The build failed due to a connection error.

thedrow:
#1327 fixes the CI failure.

thedrow:
I ran the tests manually and this works.
I'm merging this PR to move forward.

If there are any problems, we'll address them before GA.

tsunday:
Thanks @thedrow for the verification and merge 😀

PR #1327, opened 2021-04-04
Fix CI failure

thedrow:
This fixes the problem we have when we install system dependencies.


matusvalo:
LGTM. The failing flake8 can be fixed in next PR...

PR #1330, opened 2021-04-07
Coerce seconds argument to a floating point number.

thedrow:
Celery does not coerce configuration values into the right type (See celery/celery#6696).
This is a workaround. This bug will be fixed in Celery NextGen when we will refactor our configuration subsystem.


auvipy (approved):
will re organise the CI stages and clean the flake8 errors

PR #1332, opened 2021-04-13
re order CI jobs to make lint the last

auvipy:
it was decided long ago and it let the unit and integrations tests run fast for faster feedback loop


auvipy:
will try to address the lint errors in another pr

PR #1337, opened 2021-04-22
SQS - Gossip not supported

galCohen88:
add gossip not supported exception

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -167,6 +167,13 @@ class UndefinedQueueException(Exception):
     """Predefined queues are being used and an undefined queue was used."""
 
 
+class GossipUsingSQSIsNotSupportedException(Exception):
+    def __init__(self):
+        super().__init__("Ping/Gossip is not supported using SQS broker.\n"
+                         "SQS queue is a standalone infrastructure component, "
+                         "creating a new queue will not provide evidence for successful gossip.")
+
+
 class QoS(virtual.QoS):
     """Quality of Service guarantees implementation for SQS."""
 
@@ -314,6 +321,10 @@ class Channel(virtual.Channel):
         """Ensure a queue with given name exists in SQS."""
         if not isinstance(queue, str):
             return queue
+
+        if queue.endswith('reply-celery-pidbox'):
+            raise GossipUsingSQSIsNotSupportedException()
+
         # Translate to SQS name for consistency with initial
         # _queue_cache population.
         queue = self.canonical_queue_name(queue)
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -169,6 +169,7 @@ class UndefinedQueueException(Exception):
 
 class GossipUsingSQSIsNotSupportedException(Exception):
     """Ping/Gossip is not supported using SQS broker."""
+    
     def __init__(self):
         super().__init__("Ping/Gossip is not supported using SQS broker.\n"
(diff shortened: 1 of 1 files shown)


kombu/transport/SQS.py
```diff
@@ -314,6 +322,10 @@ def _new_queue(self, queue, **kwargs):
         """Ensure a queue with given name exists in SQS."""
         if not isinstance(queue, str):
             return queue
+
+        if queue.endswith('reply-celery-pidbox'):
```
thedrow, line 326: Is this the right place to put this?

galCohen88: considering the following stack trace, yes:
`  File "kombu/transport/virtual/base.py", line 528, in queue_declare
    self._new_queue(queue, **kwargs)
  File "kombu/transport/SQS.py", line 233, in _new_queue
    raise UndefinedQueueException`

galCohen88: (https://github.com/celery/celery/issues/6727)

maybe-sybr: I'd guess that @thedrow was questioning whether kombu should be aware of things like this queue suffix which are specific to celery. It feels like a special case. Is there any other property of the queue args or channel instance we can check to reach the same conclusion? Maybe we instead need some property on the class like `does_ping_mean_anything_for_this_type_of_channel() -> bool`?

Bumping this since it's on the in progress list for 5.1.0 still.

galCohen88: @maybe-sybr I think this is a relatively narrow use-case, where the generic use-case you are suggesting is too broad. I don't see a place where this property will be used besides with SQS broker. 
About the transport layer to be aware of the context - I think it's inevitable. The alternative is making Celery aware of SQS specifics, which is the exact same problem

thedrow: Transports have capabilities as far as I recall.
We can add a new one instead and let celery handle raising the error.
I think that would be better.

galCohen88: @thedrow Can you give me some hints around where would you like me to implement this? I don't have an idea where to start

maybe-sybr: https://github.com/celery/kombu/blob/3b6cd13d3b6da22f86a23b93b66c49133776cc75/kombu/transport/base.py#L133-L137

https://github.com/celery/kombu/blob/3b6cd13d3b6da22f86a23b93b66c49133776cc75/kombu/transport/SQS.py#L858-L861

I'd assume a new key in the `default_transport_capabilities` set to `True` so we don't break any existing working behaviours, having that same key set to `False` in the `SQS.Transport`, and then some logic in celery to avoid the ping/gossip behaviour if the transport has that `implements` key set to something falsey? Perhaps something like `implements['multiple_queues']`?

Alternatively, maybe the fact that `SQS.Transport` only supports the `'direct'` exchange type is a valid proxy for this limitation already? We could potentially do something like:
```python
if transport.implements.get('exchange_type', set()) - {'direct'}:
  # support more than just direct queues -> attempt gossip
  do_gossip_stuff()
```
or something like that? Maybe that check could be a fallback for if a new `implements` key which gets added per the paragraph above is missing (to handle celery upgrades with kombu having been held back)?

caveat emptor: I have literally never touched this code so I'm making this all up on the fly. If @thedrow or @auvipy have any further input, I'd trust them more than me :) Hope this helps, and thanks for the work and thought on this so far.

PR #1343, opened 2021-05-21
prepare_accept_content() now raises SerializerNotInstalled instead of KeyError

matusvalo:
When wrong serializer is passed to as accept parameter currently `KeyError` is raised. This can be misleading and confusing. Other calls uses `SerializerNotInstalled` when wrong serializer is used. This PR replaces  `KeyError` with `SerializerNotInstalled`.

After this patch all API calls providing accept parameter will raise `SerializerNotInstalled` instead of `KeyError`.


auvipy (commented):
just minor lint

matusvalo:
> just minor lint

fixed

auvipy (approved):

PR #1349, opened 2021-06-20
Add global key prefix for keys set by Redis transporter

gabor-boros:
This PR continues the [work of @wetneb](https://github.com/celery/kombu/pull/912), adding a global key prefix to queue names addressing https://github.com/celery/kombu/issues/853. In the original PR, there were some unresolved issues that are resolved by this PR.

I believe most of the comments in the original approach were answered/resolved but https://github.com/celery/kombu/pull/912#pullrequestreview-166952378.

By adding the global key prefix to the `unacked_key `, `unacked_index_key`, `unacked_mutex_key` the prefix is applied for the `unacked` and `unacked_index` keys as well when the message is not ack'ed.

Also, the implementation adds the global key prefix even to the queue received by producer. Meaning that in the following example, the Redis transport will handle `a_queue_name` as `<prefix>a_queue_name`.

```python
queue = Queue('a_queue_name', exchange=Exchange('default'))
producer.publish('message_1', exchange=queue.exchange, declare=[queue])
```

**Discussions**:

* https://github.com/celery/kombu/pull/912#pullrequestreview-166952378
* https://github.com/celery/kombu/issues/853

**Dependencies**: None

**Testing instructions**:

1. Start Redis (using docker: `docker run --name kombu_redis --rm -it -p 6379:6379 redis:latest`)
2. Connect to redis using redis-cli in a separate terminal window (`docker exec -it kombu_redis redis-cli`)
3. List all keys (`KEYS *`) and validate empty array returned
4. Run snippet "example 1" (defined below) and validate the message object is printed
5. List all keys (`KEYS *`) and validate (only) `_prefixed__kombu.binding.default` is listed
6. Flush all keys (`FLUSHALL`)
7. Run snippet "example 2" (defined below) in an `ipython` shell and validate the message object is printed
8. List all keys (`KEYS *`) and validate (only) `_prefixed_unacked`, `_prefixed_unacked_index`, and `_prefixed__kombu.binding.default` are listed
9. Quit the shell and list the keys again (`KEYS *`) and validate that the `_prefixed_a_queue_name` is restored and `_prefixed__kombu.binding.default` is still listed
10. Run `SMEMBERS _prefixed__kombu.binding.default` and validate that `1) "\x06\x16\x06\x16_prefixed_a_queue_name"` is present

_Example 1_

```python
from kombu import Exchange, Queue, Producer, Connection, Consumer
connection = Connection('redis://', transport_options={'global_keyprefix': '_prefixed_'})
producer = Producer(channel)
queue = Queue('a_queue_name', exchange=Exchange('default'))

# Publish a message
producer.publish('message_1', exchange=queue.exchange, declare=[queue])

def printer(message):
    print(message)
    # Remove the line below for unacknowledged behavior
    message.ack()

with Consumer(connection, [queue], on_message=printer):
    connection.drain_events()
```

_Example 2_

```python
from kombu import Exchange, Queue, Producer, Connection, Consumer
connection = Connection('redis://', transport_options={'global_keyprefix': '_prefixed_'})
producer = Producer(channel)
queue = Queue('a_queue_name', exchange=Exchange('default'))

# Publish a message
producer.publish('message_1', exchange=queue.exchange, declare=[queue])

def printer(message):
    print(message)

with Consumer(connection, [queue], on_message=printer):
    connection.drain_events()
```

**Author notes and concerns**:

The original PR listed some maintainer concerns, but -- as I mentioned -- I believe those were "discussed" though not confirmed. In case any questions from the original PR are still open and not addressed by this PR, please let me know.

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -38,7 +38,8 @@ Transport Options
 * ``unacked_restore_limit``
 * ``fanout_prefix``
 * ``fanout_patterns``
-* ``global_keyprefix``: (str) The global key prefix to be prepended to all keys used by Kombu
+* ``global_keyprefix``: (str) The global key prefix to be prepended to all keys
+  used by Kombu
 * ``socket_timeout``
 * ``socket_connect_timeout``
 * ``socket_keepalive``
@@ -50,6 +51,7 @@ Transport Options
 * ``priority_steps``
 """
 
+import functools
 import numbers
 import socket
 from bisect import bisect
@@ -178,6 +180,95 @@ def Mutex(client, name, expire):
 
 def _after_fork_cleanup_channel(channel):
     channel._after_fork()
+
+
+class GlobalKeyPrefixMixin:
+    """Mixin to provide common logic for global key prefixing.
+
+    Overriding all the methods used by Kombu with the same key prefixing logic
+    would be cumbersome and inefficient. Hence, we override the command
+    execution logic that is called by all commands.
+    """
+
+    PREFIXED_SIMPLE_COMMANDS = [
+        "HDEL",
+        "HGET",
+        "HSET",
+        "LLEN",
+        "LPUSH",
+        "PUBLISH",
+        "SADD",
+        "SET",
+        "SMEMBERS",
+        "ZADD",
+        "ZREM",
+        "ZREVRANGEBYSCORE",
+    ]
+
+    PREFIXED_COMPLEX_COMMANDS = {
+        "BRPOP": {"args_start": 0, "args_end": -1},
+        "EVALSHA": {"args_start": 2, "args_end": 3},
+    }
+
+    def _prefix_args(self, args):
+        args = list(args)
+        command = args.pop(0)
+
+        if command in self.PREFIXED_SIMPLE_COMMANDS:
+            args[0] = self.global_keyprefix + str(args[0])
+
+        if command in self.PREFIXED_COMPLEX_COMMANDS.keys():
+            args_start = self.PREFIXED_COMPLEX_COMMANDS[command]["args_start"]
+            args_end = self.PREFIXED_COMPLEX_COMMANDS[command]["args_end"]
+
+            pre_args = args[:args_start] if args_start > 0 else []
+
+            if args_end is not None:
+                post_args = args[args_end:]
+            elif args_end < 0:
+                post_args = args[len(args):]
+            else:
+                post_args = []
+
+            args = pre_args + [
+                self.global_keyprefix + str(arg)
+                for arg in args[args_start:args_end]
+            ] + post_args
+
+        return [command, *args]
+
+    def execute_command(self, *args, **kwargs):
+        return super().execute_command(*self._prefix_args(args), **kwargs)
+
+    def pipeline(self, transaction=True, shard_hint=None):
+        return PrefixedRedisPipeline(
+            self.connection_pool,
+            self.response_callbacks,
+            transaction,
+            shard_hint,
+            global_keyprefix=self.global_keyprefix,
+        )
+
+
+class PrefixedStrictRedis(GlobalKeyPrefixMixin, redis.Redis):
+    """Returns a ``StrictRedis`` client that prefixes the keys it uses."""
+
+    def __init__(self, *args, **kwargs):
+        self.global_keyprefix = kwargs.pop('global_keyprefix', '')
+        redis.Redis.__init__(self, *args, **kwargs)
+
+
+class PrefixedRedisPipeline(GlobalKeyPrefixMixin, redis.client.Pipeline):
+    """Custom Redis pipeline that takes global_keyprefix into consideration.
+
+    As the ``PrefixedStrictRedis`` client uses the `global_keyprefix` to prefix
+    the keys it uses, the pipeline called by the client must be able to prefix
+    the keys as well.
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.global_keyprefix = kwargs.pop('global_keyprefix', '')
+        redis.client.Pipeline.__init__(self, *args, **kwargs)
 
 
 class QoS(virtual.QoS):
@@ -569,23 +660,6 @@ class Channel(virtual.Channel):
             # by default.
             self.keyprefix_fanout = ''
 
-        # Prepend the global key prefix
-        self.unacked_key = self._queue_with_prefix(self.unacked_key)
-        self.unacked_index_key = self._queue_with_prefix(
-            self.unacked_index_key
-        )
-        self.unacked_mutex_key = self._queue_with_prefix(
-            self.unacked_mutex_key
-        )
-
-        # The default `keyprefix_queue` starts with an underscore, therefore
-        # adding a prefix ending an undescore will result in double
-        # underscores. Since both `keyprefix_queue` and `global_keyprefix`
-        # can be set by the user, this behavior is better than manipulating
-        # `keyprefix_queue` here.
-        self.keyprefix_queue = self._queue_with_prefix(self.keyprefix_queue)
-        self.keyprefix_fanout = self._queue_with_prefix(self.keyprefix_fanout)
-
         # Evaluate connection.
         try:
             self.client.ping()
@@ -600,10 +674,6 @@ class Channel(virtual.Channel):
 
         if register_after_fork is not None:
             register_after_fork(self, _after_fork_cleanup_channel)
-
-    def _queue_with_prefix(self, queue):
-        """Return the queue name prefixed with `global_keyprefix` if set."""
-        return self.global_keyprefix + queue
 
     def _after_fork(self):
         self._disconnect_pools()
@@ -660,7 +730,6 @@ class Channel(virtual.Channel):
         return self._restore(message, leftmost=True)
 
     def basic_consume(self, queue, *args, **kwargs):
-        queue = self._queue_with_prefix(queue)
         if queue in self._fanout_queues:
             exchange, _ = self._fanout_queues[queue]
             self.active_fanout_queues.add(queue)
@@ -798,7 +867,12 @@ class Channel(virtual.Channel):
         keys = [self._q_for_pri(queue, pri) for pri in self.priority_steps
                 for queue in queues] + [timeout or 0]
         self._in_poll = self.client.connection
-        self.client.connection.send_command('BRPOP', *keys)
+
+        command_args = ['BRPOP', *keys]
+        if self.global_keyprefix:
+            command_args = self.client._prefix_args(command_args)
+
+        self.client.connection.send_command(*command_args)
 
     def _brpop_read(self, **options):
         try:
@@ -875,7 +949,6 @@ class Channel(virtual.Channel):
             self.auto_delete_queues.add(queue)
 
     def _queue_bind(self, exchange, routing_key, pattern, queue):
-        queue = self._queue_with_prefix(queue)
         if self.typeof(exchange).type == 'fanout':
             # Mark exchange as fanout.
             self._fanout_queues[queue] = (
@@ -1055,6 +1128,13 @@ class Channel(virtual.Channel):
             raise VersionMismatch(
                 'Redis transport requires redis-py versions 3.2.0 or later. '
                 'You have {0.__version__}'.format(redis))
+
+        if self.global_keyprefix:
+            return functools.partial(
+                PrefixedStrictRedis,
+                global_keyprefix=self.global_keyprefix,
+            )
+
         return redis.StrictRedis
 
     @contextmanager
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -38,6 +38,7 @@ Transport Options
 * ``unacked_restore_limit``
 * ``fanout_prefix``
 * ``fanout_patterns``
+* ``global_keyprefix``: (str) The global key prefix to be prepended to all keys used by Kombu
 * ``socket_timeout``
 * ``socket_connect_timeout``
 * ``socket_keepalive``
@@ -485,6 +486,11 @@ class Channel(virtual.Channel):
     #: Disable for backwards compatibility with Kombu 3.x.
     fanout_patterns = True
 
+    #: The global key prefix will be prepended to all keys used
+    #: by Kombu, which can be useful when a redis database is shared
+    #: by different users. By default, no prefix is prepended.
+    global_keyprefix = ''
+
     #: Order in which we consume from queues.
     #:
     #: Can be either string alias, or a cycle strategy class
@@ -526,6 +532,7 @@ class Channel(virtual.Channel):
          'unacked_restore_limit',
          'fanout_prefix',
          'fanout_patterns',
+         'global_keyprefix',
          'socket_timeout',
          'socket_connect_timeout',
          'socket_keepalive',
@@ -562,6 +569,23 @@ class Channel(virtual.Channel):
             # by default.
             self.keyprefix_fanout = ''
 
+        # Prepend the global key prefix
+        self.unacked_key = self._queue_with_prefix(self.unacked_key)
+        self.unacked_index_key = self._queue_with_prefix(
+            self.unacked_index_key
+        )
+        self.unacked_mutex_key = self._queue_with_prefix(
+            self.unacked_mutex_key
+        )
+
+        # The default `keyprefix_queue` starts with an underscore, therefore
+        # adding a prefix ending an undescore will result in double
+        # underscores. Since both `keyprefix_queue` and `global_keyprefix`
+        # can be set by the user, this behavior is better than manipulating
+        # `keyprefix_queue` here.
+        self.keyprefix_queue = self._queue_with_prefix(self.keyprefix_queue)
+        self.keyprefix_fanout = self._queue_with_prefix(self.keyprefix_fanout)
+
         # Evaluate connection.
         try:
             self.client.ping()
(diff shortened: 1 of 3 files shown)


kombu/transport/redis.py
```diff
@@ -562,6 +569,23 @@ def __init__(self, *args, **kwargs):
...
             self.keyprefix_fanout = ''
 
+        # Prepend the global key prefix
+        self.unacked_key = self._queue_with_prefix(self.unacked_key)
+        self.unacked_index_key = self._queue_with_prefix(
+            self.unacked_index_key
+        )
+        self.unacked_mutex_key = self._queue_with_prefix(
+            self.unacked_mutex_key
+        )
```
pomegranited, line 579: @gabor-boros 

The final step isn't working for me as described?

> 7. Run snippet "example 2" (defined below) and validate the message object is printed 👍 
> 8. List all keys (`KEYS *`) and validate (only) `_prefixed_unacked`, `_prefixed_unacked_index`, and `_prefixed__kombu.binding.default` are listed

I'm seeing this instead:
```
127.0.0.1:6379> keys *
1) "a_queue_name"
2) "_prefixed__kombu.binding.default"
```

Any idea why?

gabor-boros: @pomegranited I have no idea, as it is running fine for me.

pomegranited: Ok, maybe my virtualenv setup was different somehow?

Here's what I trued -- do you see any differences with yours that might be relevant?

```bash
$ python3.8 -m venv ~/.virtualenvs/redis-celery
$ source ~/.virtualenvs/redis-celery/bin/activate
(redis-celery) $ python --version
Python 3.8.3

# Started by using the versions of `redis` and `celery` taken from
# https://github.com/edx/edx-platform/blob/master/requirements/edx/base.txt 
(redis-celery) $ cat requirements_edx.txt
django-celery-results<2.1  # brings in supported celery
redis==3.5.3
-e git+https://github.com/open-craft/kombu.git@gabor/add-global-keyprefix#egg=kombu==gabor/add-global-keyprefix

# But the same issue happens when I'm using the latest versions of `redis` and `celery`:
(redis-celery) $ cat requirements.txt
(redis-celery) helga:global_keys$ cat requirements.txt 
celery
redis
-e git+https://github.com/open-craft/kombu.git@gabor/add-global-keyprefix#egg=kombu==gabor/add-global-keyprefix

(redis-celery) $ pip install -r requirements.txt
(redis-celery) $ python ex2.py
<Message object at 0x7faeeac2b550 with details {'state': 'RECEIVED', 'content_type': 'text/plain', 'delivery_tag': 'f0a7daa4-7813-4e3d-aeb7-1047b47b6c94', 'body_length': 9, 'properties': {}, 'delivery_info': {'exchange': 'default', 'routing_key': ''}}>
Restoring 1 unacknowledged message(s)

# Meanwhile, over at the redis-cli:
127.0.0.1:6379> keys *
1) "_prefixed__kombu.binding.default"
2) "a_queue_name"
```


```

File: kombu/transport/redis.py [hunk 2]
```diff
@@ -485,6 +486,11 @@ class Channel(virtual.Channel):
     #: Disable for backwards compatibility with Kombu 3.x.
     fanout_patterns = True
 
+    #: The global key prefix will be prepended to all keys used
+    #: by Kombu, which can be useful when a redis database is shared
+    #: by different users. By default, no prefix is prepended.
+    global_keyprefix = ''
```
auvipy, line 492: this needs to be documened

kombu/transport/redis.py
gabor-boros: I did find the issue. Will push a commit later today fixing the problem.

kombu/transport/redis.py
gabor-boros: @auvipy Should I regenerate the documentation somehow or is this something I'd need to do somewhere else?

kombu/transport/redis.py
pomegranited: @gabor-boros You're brilliant, that fixed it! Will do another code review pass and approve this. Thank you!

pomegranited (approved):
:+1: This is clean and functioning beautifully, thank you @gabor-boros !

- [x] I tested this using the PR test instructions.
- [x] I read through the code.
- [x] Includes documentation for the new argument -- [open question](https://github.com/celery/kombu/pull/1349#pullrequestreview-689303433) about where/how to update the [docs generated](https://docs.celeryproject.org/projects/kombu/en/stable/reference/kombu.transport.redis.html) from these comments.

kombu/transport/redis.py
```diff
@@ -38,6 +38,7 @@
 * ``unacked_restore_limit``
 * ``fanout_prefix``
 * ``fanout_patterns``
+* ``global_keyprefix``
```
matusvalo, line 41: Could you please add description here and type of the options?

gabor-boros: @matusvalo Sure I can and I [would](https://github.com/celery/kombu/pull/1349#discussion_r657200096). I have some questions though:

* What kind of documentation we mean here? If you mean to have a description of what the `global_keyprefix` means, it is described [here](https://github.com/celery/kombu/blob/6d2ea9a0e7139b77100607c7c07eb246b39a8036/kombu/transport/redis.py#L489-L492) that is used by sphinx as I can see. If you meant another documentation, please advise where I can find that as I'm not a regular contributor at kombu. I just generated the documentation locally and the changes are reflected.
* What options do you refer to? Since it is a string variable, the value could be anything that the user sets.

gabor-boros:
@matusvalo @auvipy The linter issues are fixed and I had a [question](https://github.com/celery/kombu/pull/1349#discussion_r660783267) regarding documentation. Could you please take a look?

kombu/transport/redis.py
matusvalo: > What kind of documentation we mean here? If you mean to have a description of what the global_keyprefix means, it is described here that is used by sphinx as I can see.

The users of this transport are mainly interested in Transport Option they should be documented first.

> What options do you refer to?

Just add type and type and short description of ``global_keyprefix`` transport option
```suggestion
* ``global_keyprefix``: (str) The global key prefix to be prepended to all keys used by Kombu
```

gabor-boros: Thank you for the suggestion @matusvalo; now I see the way you wanted to write the docs.

matusvalo:
Hi @gabor-boros thank you for your PR. The CI seems to be failing with following:
```
kombu/transport/redis.py:41:80: E501 line too long (93 > 79 characters)
```

matusvalo:
Unfortunatelly, I will also need more time for PR review so please be patient.

gabor-boros:
@matusvalo The linter error is fixed. Let me know if you need me to update anything else.

matusvalo (changes requested):
@gabor-boros unfortunately, the code misses the case when default exchange is used:
```
127.0.0.1:6379> keys *
(empty list or set)
127.0.0.1:6379>
(kombu38) matus@dev:~/dev/kombu$ cat p.py
import kombu
with kombu.Connection('redis://') as conn:
    with conn.channel() as channel:
        producer = kombu.Producer(channel)

        producer.publish(
                {'hello': 'world'},
                retry=True,
                routing_key='rkey',
                retry_policy={
                    'interval_start': 0, # First retry immediately,
                    'interval_step': 2,  # then increase by 2s for every retry.
                    'interval_max': 30,  # but don't exceed 30s between retries.
                    'max_retries': 30,   # give up after 30 tries.
                },
                transport_options={'global_keyprefix': 'myprefix'}
            )
(kombu38) matus@dev:~/dev/kombu$ python p.py
(kombu38) matus@dev:~/dev/kombu$ redis-cli
127.0.0.1:6379> keys *
1) "rkey"
127.0.0.1:6379>
```

gabor-boros:
@matusvalo Oh! Thank you for pointing it out. I'll update the PR today to include that case too.

gabor-boros:
> @gabor-boros unfortunately, the code misses the case when default exchange is used:

@matusvalo I was curious why did the `global_keyprefix` not passed. I may overlook something, but the `transport_options` is **not** a valid `kwarg` for the `publish` method. In fact, the `**properties` is available on `publish`, but nothing reads `transport_options` from it. Therefore **none** of the transport options are can be set from here. If you try `unacked_key` you can see that that's not changed either.

matusvalo:
@gabor-boros yes it was typo but it is still not working (this time I have ensured that transport options were filled as expected):
```
(kombu38) matus@dev:~/dev/kombu$ redis-cli keys '*'
(empty list or set)
(kombu38) matus@dev:~/dev/kombu$ cat p.py
import kombu
with kombu.Connection('redis://', transport_options={'global_keyprefix': 'myprefix'}) as conn:
    with conn.channel() as channel:
        producer = kombu.Producer(channel)

        producer.publish(
                {'hello': 'world'},
                retry=True,
                routing_key='rkey',
                retry_policy={
                    'interval_start': 0, # First retry immediately,
                    'interval_step': 2,  # then increase by 2s for every retry.
                    'interval_max': 30,  # but don't exceed 30s between retries.
                    'max_retries': 30,   # give up after 30 tries.
                }
            )
(kombu38) matus@dev:~/dev/kombu$ python p.py
(kombu38) matus@dev:~/dev/kombu$ redis-cli keys '*'
1) "rkey"
```

BTW: I you look into codebase of kombu it cannot work since you are not adjusting key value here (method `_put()` is used when data are published):
https://github.com/celery/kombu/blob/e5dbfed8ae9fb5e7d2d76be682517a9779137c77/kombu/transport/redis.py#L829-L834

Moreover, I have also problem with the PR itself. I think that it does changes on wrong layer. Key name adjustments should be done in lower layer in client itself. I propose a solution by :

1. creating a wrapper of redis client which will adjust key name with the prefix:

```python
class PrefixedClient:

    def __init__(self, redis_client, prefix=''):
        self.prefix = prefix
        self.client = client

    def lpush(queue, *args, **kwargs):
        return self.client.lpush(prefix + queue, *args, **kwargs)
    ... # All other used methods needs to be implemented in similar way
```

2. creating a wrapper of pipeline doing the same as redis client (prefixing the queues).
3. the codebase of redis transport in this case will not need any changes (except creating instance of PrefixedClient).

The reason behind this is following:
* all logic around prefixing is in single place, so the readability will be better.
* in this way we can guarantee that all key will be prefixed.
* unittests will be much easer to create covering all cases.

gabor-boros:
@matusvalo Thank you for your feedback and for investigating the root cause of the issue. Adding `queue = self._queue_with_prefix(queue)` to the `_put` method fixes the issue and afterward the `rkey` in the example prefixed as expected.

Do you mean to create the `PrefixedClient` in Kombu or in the Redis client implementation?

gabor-boros:
@matusvalo Could you please help me to get unblocked and sort out the remaining bits of this PR?

matusvalo:
Sorry @gabor-boros for late reply. I was thinking about creating wrapper around `PrefixedClient`in kombu. It means you create new class in `kombu/transport/redis.py` with all used methods wrapped. Each method do just prefixing. Does it make sense?

gabor-boros:
@matusvalo Yes, it makes. I'll update the PR with the requested changes soon.

kombu/transport/redis.py
pomegranited: @gabor-boros Thanks for updating this PR, but I'm unable to get the example scripts you wrote to print out any messages.

I used the [same setup as before](https://github.com/celery/kombu/pull/1349#discussion_r655881808), but when I run the scripts, I get no messages dequeued, and no output printed from either ex1.py or ex2.py:

```
(redis-celery) $ python ex1.py
(no output)
```

The keys are showing properly prefixed in redis though:
```
127.0.0.1:6379> keys *
1) "_prefixed__kombu.binding.default"
2) "_prefixed_a_queue_name"
```

Are you seeing the same thing?

gabor-boros: @pomegranited I see the same, though it is strange. I tested the changes only with the latest script provided by @matusvalo.

@matusvalo do you have any clue why the `Consumer` is not calling the `on_message` callback at all in the scripts that are provided as part of the PR description?

gabor-boros: Or @matusvalo do you have any idea how to properly test the scenarios other than the script you provided?

auvipy (approved):
please fix the lint failures

gabor-boros:
@auvipy I wasn't satisfied with the constant names I introduced so renaming them solves two problems at once.

kombu/transport/redis.py
```diff
@@ -179,6 +181,91 @@ def _after_fork_cleanup_channel(channel):
...
+    """Custom Redis pipeline that takes global_keyprefix into consideration.
+
+    As the ``PrefixedStrictRedis`` client uses the `global_keyprefix` to prefix
+    the keys it uses, the pipeline called by the client must be able to prefix
+    the keys as well.
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.global_keyprefix = kwargs.pop('global_keyprefix', '')
+        redis.client.Pipeline.__init__(self, *args, **kwargs)
```
auvipy, line 266: can we use super() call here instead?

thedrow: I don't think we can as this class is using multiple inheritance.
I'm not sure why it has to be both a client and a pipeline.

gabor-boros: @thedrow is right, we cannot use only `super()` call here as far as I remember correctly.

The thing is that [`Pipeline`](https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L4245) inherits from `Redis`. Both has the `execute_command` method which must be overridden to prefix the keys it is using. We could do three things here as far as I understand:

1. Create a mixin that overrides the `execute_command` method and use that for both the pipeline and the custom redis class
2. Reimplement the exact same key prefixing logic in both pipeline and custom redis classes
3. Do what I did and create the custom pipeline using the original Pipeline class and the already patched redis class (`PrefixedStrictRedis`) and override the init function to call only the Pipeline's init

Probably we have other options here too, so if you would prefer something else, please let me know and I'll adjust the PR.

cc: @auvipy

thedrow: I think that a mixin would be far more obvious.
This code confused me as I was unaware of the fact that a Pipeline inherits from StrictRedis.

gabor-boros: @thedrow Right, I'll refactor to have a mixin instead.

gabor-boros: @thedrow the mixin was introduced as part of df8dfdc91b4ea5ed0a1b50c9f6cee01216d6e607

auvipy (approved):

matusvalo (commented):
I am missing more unittests for `_prefix_args()` method. Mainly for case of `PREFIXED_COMPLEX_COMMANDS`. Otherwise LGTM.

gabor-boros:
Thank you @auvipy, @matusvalo, and @thedrow for the reviews! I'll do the necessary remaining changes and add more tests for the prefixing logic. The `functools.partial` change is already applied.

The only thing that worries me and @pomegranited is [this comment](https://github.com/celery/kombu/pull/1349#discussion_r689349840). It was resolved, though it is alarming that the message dequeue may not happen. Could one of you help to take a closer look at that? Although I'm working on this PR for a while and getting more and more familiar with Kombu, this is something that would take a lot more time for me to investigate/check/validate.

matusvalo:
Thank you @gabor-boros . I will check the comment.

auvipy (changes requested):
tests are failing after new changes

gabor-boros:
@auvipy Thanks for noticing. I just squashed the commits and will fix the failing test

gabor-boros:
@auvipy @matusvalo @thedrow I believe this is now ready for your review.

auvipy (dismissed):

matusvalo (changes requested):
It seems that example from PR description (at least first one) is not working with the patch. When executed nothing is printed:

```
(kombu37) matus@matus-debian:~/zmaz$ python ex1.py
(kombu37) matus@matus-debian:~/zmaz$
```

But when I have adjusted the example in a way that I have removed prefixed option - hence connection looked like following:
```
connection = Connection('redis://')
```

The script worked fine:
```
(kombu37) matus@matus-debian:~/zmaz$ python ex1.py
<Message object at 0x7fa5df224c18 with details {'state': 'RECEIVED', 'content_type': 'text/plain', 'delivery_tag': 'c7bf6578-4fb7-4c52-a921-4a975046ebd2', 'body_length': 9, 'properties': {}, 'delivery_info': {'exchange': 'default', 'routing_key': ''}}>
```

I think this PR needs more love... I will try to allocate some time for investigation - please be patient...

kombu/transport/redis.py
```diff
@@ -179,6 +182,95 @@ def _after_fork_cleanup_channel(channel):
...
+            else:
+                post_args = []
+
+            args = pre_args + [
+                self.global_keyprefix + str(arg)
+                for arg in args[args_start:args_end]
+            ] + post_args
+
+        return [command, *args]
+
```
matusvalo, line 239: I have found out that PR is missing removal prefixes from keys returned by Redis Command. Following snippet is fixing it:

```suggestion

    def parse_response(self, connection, command_name, **options):
        """Parses a response from the Redis server.

        Method wraps ``redis.parse_response()`` to remove prefixes of keys
       returned by redis command.
        """
        ret = super().parse_response(connection, command_name, **options)
        if command_name == 'BRPOP' and ret:
            key, value = ret
            key = key[len(self.global_keyprefix):]
            return key, value
        return ret
```

matusvalo (commented):
@gabor-boros I have found out the reason why the examples were failing - see my code review. The missing piece are unittests covering it. Unfortunately, I am not able to push changes directly to PR branch since you have unchecked this possibility.

kombu/transport/redis.py
pomegranited: Oh brilliant @matusvalo , that fixes all the issues I was seeing. Thank you!

pomegranited (approved):
:+1:

- [x] I tested this using the PR test instructions, and everything works perfectly now!
- [x] I read through the code
- [x] <strike>I checked for accessibility issues</strike> N/A
- [x] Includes documentation

matusvalo (approved):

PR #1351, opened 2021-06-20
Add support for setting redis username

gabor-boros:
Although the underlying Redis package supports setting username for connections, the connection params for username are not passed to the `Redis` client.

This PR extracts the `userid` from connection info (or host) if and passes to Redis client ensuring the connection can be established for other users than the default `default` user (set when username is `None`).

**Dependencies**: None

**Testing instructions**:

1. Start Redis (using docker: `docker run --name kombu_redis --rm -it -p 6379:6379 redis:latest`)
2. Connect to redis using redis-cli in a separate terminal window (`docker exec -it kombu_redis redis-cli`)
3. List ACLs (`ACL LIST`) and validate only `default` user is listed
4. Run snippet "example 1" (defined below) and validate the message object is printed
5. Set a new password for `default` by running ` acl setuser default on >securepass`
6. List ACLs (`ACL LIST`) and validate `default` user now has a password hash set 
7. Set a new user by running ` acl setuser testuser on >password allcommands`
8. List ACLs (`ACL LIST`) and validate `testuser` is created with a password hash
9. Run snippet "example 2" (defined below) and validate the message object is printed

_Example 1_

```python
from kombu import Exchange, Queue, Producer, Connection, Consumer
connection = Connection('redis://')
producer = Producer(channel)
queue = Queue('a_queue_name', exchange=Exchange('default'))

# Publish a message
producer.publish('message_1', exchange=queue.exchange, declare=[queue])

def printer(message):
    print(message)
    # Remove the line below for unacknowledged behavior
    message.ack()

with Consumer(connection, [queue], on_message=printer):
    connection.drain_events()
```

_Example 2_

```python
from kombu import Exchange, Queue, Producer, Connection, Consumer
connection = Connection('redis://testuser:password@127.0.0.1:6379/0')
producer = Producer(channel)
queue = Queue('queue_name', exchange=Exchange('default'))

# Publish a message
producer.publish('message_1', exchange=queue.exchange, declare=[queue])

def printer(message):
    print(message)
    # Remove the line below for unacknowledged behavior
    message.ack()

with Consumer(connection, [queue], on_message=printer):                      
    connection.drain_events()
```

**Author notes and concerns**:

Redis' "new" ACL features released as part of Redis 6 introduced the possibility of defining additional users identified by passwords. Therefore it would be a great achievement if Kombu (and Celery) would support connecting to Redis using an ACL limited user.

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -944,6 +944,7 @@ class Channel(virtual.Channel):
             'host': conninfo.hostname or '127.0.0.1',
             'port': conninfo.port or self.connection.default_port,
             'virtual_host': conninfo.virtual_host,
+            'username': conninfo.userid,
             'password': conninfo.password,
             'max_connections': self.max_connections,
             'socket_timeout': self.socket_timeout,
@@ -974,7 +975,7 @@ class Channel(virtual.Channel):
                 pass
         host = connparams['host']
         if '://' in host:
-            scheme, _, _, _, password, path, query = _parse_url(host)
+            scheme, _, _, username, password, path, query = _parse_url(host)
             if scheme == 'socket':
                 connparams = self._filter_tcp_connparams(**connparams)
                 connparams.update({
@@ -984,6 +985,7 @@ class Channel(virtual.Channel):
                 connparams.pop('socket_connect_timeout', None)
                 connparams.pop('socket_keepalive', None)
                 connparams.pop('socket_keepalive_options', None)
+            connparams['username'] = username
             connparams['password'] = password
 
             connparams.pop('host', None)
--- a/t/unit/transport/test_redis.py
+++ b/t/unit/transport/test_redis.py
@@ -695,6 +695,10 @@ class test_Channel:
         self.channel.connection.client.hostname = 'george.vandelay.com'
         assert self.channel._connparams()['host'] == 'george.vandelay.com'
 
+    def test_connparams_username(self):
+        self.channel.connection.client.userid = 'kombu'
+        assert self.channel._connparams()['username'] == 'kombu'
+
     def test_connparams_password_for_unix_socket(self):
         self.channel.connection.client.hostname = \
             'socket://:foo@/var/run/redis.sock'
@@ -1414,7 +1418,7 @@ class test_RedisSentinel:
                 min_other_sentinels=0, password=None, sentinel_kwargs=None,
                 socket_connect_timeout=None, socket_keepalive=None,
                 socket_keepalive_options=None, socket_timeout=None,
-                retry_on_timeout=None)
+                username=None, retry_on_timeout=None)
 
             master_for = patched.return_value.master_for
             master_for.assert_called()
@@ -1437,7 +1441,7 @@ class test_RedisSentinel:
                 min_other_sentinels=0, password=None, sentinel_kwargs=None,
                 socket_connect_timeout=None, socket_keepalive=None,
                 socket_keepalive_options=None, socket_timeout=None,
-                retry_on_timeout=None)
+                username=None, retry_on_timeout=None)
 
             master_for = patched.return_value.master_for
             master_for.assert_called()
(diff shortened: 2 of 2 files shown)


t/unit/transport/test_redis.py
```diff
@@ -695,6 +695,10 @@ def test_connparams_regular_hostname(self):
         self.channel.connection.client.hostname = 'george.vandelay.com'
         assert self.channel._connparams()['host'] == 'george.vandelay.com'
 
+    def test_connparams_username(self):
+        self.channel.connection.client.userid = 'kombu'
+        assert self.channel._connparams()['username'] == 'kombu'
+
     def test_connparams_password_for_unix_socket(self):
         self.channel.connection.client.hostname = \
             'socket://:foo@/var/run/redis.sock'
```
pomegranited, line 704: Could you also add a test here to `test_connparams_password_for_unix_socket` that tests the parsing of the username from the hostname string?

pomegranited (approved):
:+1: Have requested 1 additional test, and I had to use a slightly modified command to set up the `testuser`, but this works great, thank you @gabor-boros !  

`acl setuser testuser on >password allkeys allcommands`

- [x] I tested this using the instructions in the PR description. 
- [x] I read through the code
- [x] <strike>Includes documentation</strike> N/A -- standard connection string

t/unit/transport/test_redis.py
gabor-boros: @pomegranited I went with a different approach but still testing to get the username and password from the connection url.

kombu/transport/redis.py
```diff
@@ -944,6 +944,7 @@ def _connparams(self, asynchronous=False):
             'host': conninfo.hostname or '127.0.0.1',
             'port': conninfo.port or self.connection.default_port,
             'virtual_host': conninfo.virtual_host,
+            'username': conninfo.userid,
```
auvipy, line 947: should we also update the related docs?

auvipy (changes requested):

kombu/transport/redis.py
gabor-boros: @auvipy Actually, this is a good question. I'd say yes if the documentation explicitly says that the username part of the connection URL won't be parsed (before this change), otherwise, I'd say that this should be "mainstream" that a connection URL can contain a username.

If you would like me, I'm happy to find the relevant section in the documentation and explicitly state that, from now on, usernames will be parsed too.

thedrow: Please do.

gabor-boros: @thedrow @auvipy I'm happy to extend any documentation, if you could advise a bit. The documentation has no mention at all about usernames in terms of the Redis transport. In fact, the documentation is defined as
```rst
.. code-block::

    redis://REDIS_ADDRESS[:PORT][/VIRTUALHOST]
    rediss://REDIS_ADDRESS[:PORT][/VIRTUALHOST]

To use sentinel for dynamic Redis discovery,
the connection string has the following format:

.. code-block::

    sentinel://SENTINEL_ADDRESS[:PORT]
```
I do believe -- but correct me if I'm wrong -- the reason behind this is that the host parameter (`REDIS_ADDRESS`) may contain username and password as per RFC 2617. If you would like me, I can describe that in the inline [documentation](https://github.com/celery/kombu/blob/master/kombu/transport/redis.py#L12-L27), though that would be a bit verbose.

If you would like me to extend another documentation that I did not find (yet), please help me with the link.

matusvalo:
@gabor-boros thank you for our PR. Let me review this PR. I try to do it in next few days.

kombu/transport/redis.py
matusvalo: The documentation is not 100% complete. I am not 100% whether is good idea to refer to some standard because URL parsing is done by `urllib.parse.urlparse` but some magic is involved during the parsing - see `kombu/utils/url.py:url_to_parts()`. Let it keep simple. Please extend the documentation as follows:

```
redis://[USER:PASSWORD@]REDIS_ADDRESS[:PORT][/VIRTUALHOST]
rediss://[USER:PASSWORD@]REDIS_ADDRESS[:PORT][/VIRTUALHOST]
```

Maybe we can also add some examples to the documentation.

Otherwise, the PR seems to be fine for me.

matusvalo (approved):

matusvalo:
Thank you @gabor-boros for your PR. I have accepted the PR. The documentation update is committed in next commit.

PR #1368, opened 2021-08-13
Use Python's built-in json module by default, instead of simplejson

nikolas:
Only use simplejson if it's absolutely necessary - Python's built-in json module is better if it's available.

### DIFF
--- a/kombu/utils/json.py
+++ b/kombu/utils/json.py
@@ -12,18 +12,18 @@ except ImportError:  # pragma: no cover
         """Dummy object."""
 
 try:
+    import json
+    _json_extra_kwargs = {}
+
+    class _DecodeError(Exception):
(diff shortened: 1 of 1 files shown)


auvipy (approved):

kombu/utils/json.py
```diff
@@ -12,18 +12,18 @@ class DjangoPromise:
         """Dummy object."""
 
 try:
+    import json
+    _json_extra_kwargs = {}
+
+    class _DecodeError(Exception):
+        pass
+except ImportError:                 # pragma: no cover
```
MelnykR, line 20: As far as I understand, ImportError will be raised only for python<2.6, which is not supported.

Maybe remove simplejson related code at all?

PR #1373, opened 2021-08-23
Add support for custom yaml encoding

ticosax:
if `__yaml__` method is implemented, then use it
to instruct the yaml Encoder on how to perform arbitrary custom instances.

Related to https://github.com/celery/celery/issues/6901

### DIFF
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -101,3 +101,4 @@ Kombu utils
     kombu.utils.time
     kombu.utils.url
     kombu.utils.uuid
+    kombu.utils.yaml
--- a/docs/reference/kombu.utils.yaml.rst
+++ b/docs/reference/kombu.utils.yaml.rst
@@ -1,0 +1,12 @@
+==========================================================
+ Yaml Utilities - ``kombu.utils.yaml``
+==========================================================
+
+.. contents::
+    :local:
+.. currentmodule:: kombu.utils.yaml
+
+.. automodule:: kombu.utils.yaml
+    :members:
+    :undoc-members:
+
--- a/kombu/utils/yaml.py
+++ b/kombu/utils/yaml.py
@@ -1,0 +1,43 @@
+from typing import Type
+
+from yaml import SafeDumper
+
+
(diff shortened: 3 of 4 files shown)


matusvalo (changes requested):
This PR is introducing new dependency to a library - yaml. To be honest I am against introducing another serialiser to the kombu. This can be done via separate library.

ticosax:
> This PR is introducing new dependency to a library - yaml. To be honest I am against introducing another serialiser to the kombu. This can be done via separate library.

Thanks for your wisdom and sympathy
yaml is supported since the inception of the project, it goes back to 2010.

https://github.com/celery/kombu/commit/da3e75459489e2069b4de487eb578c99bc6a19ed#diff-7b3ed02bc73dc06b7db906cf97aa91dec2b2eb21f2d92bc5caa761df5bbc168fR217

It's a bug I reported https://github.com/celery/celery/issues/6901 and before making the proposal I asked if the community is interested in receiving the fix. I was under the impression that @thedrow  was [okay](https://github.com/celery/celery/issues/6901#issuecomment-894808847) with it.
I'm not spending time on opensource projects for fun and glory, It cost times. And I know maintainers time is scarce and precious, this is why I took precautions to ask first.

Before we continue, @matusvalo do you want that fix or not ? I'm asking again because I don't waste more anybody's time.

thedrow:
This is fine as long as we make yaml an optional dependency.
If it can't be imported, just skip importing the module.

ticosax:
> This is fine as long as we make yaml an optional dependency.
> If it can't be imported, just skip importing the module.

I agree, and I think that's what this PR does. If not, please give me some pointers.

matusvalo:
@ticosax I am sorry I missed that celery is mentioning yaml as supported. In this case I won't be against adding it to kombu when it is optional. In this case, I think that even we can go with this to master. @ticosax please just fix the linter errors and I will accept your PR.

ticosax:
Nice, thanks for the feedback.
do you think this is going in to the right direction regarding https://github.com/celery/celery/pull/6925 ?

Issue #1377, opened 2021-08-29
Broker State of Virtual transport is global


matusvalo:
State of Broker in kombu is global:
https://github.com/celery/kombu/blob/24b0820d0bddcad074209cc1381d92d17207448c/kombu/transport/virtual/base.py#L874-L875

**Note**:  This issue is affecting virtual transport **only**. It is not affecting e.g. AMQP transport.

This is causing two issues occurring mainly in virtual transport (Redis):
1. User cannot connect to multiple Redis brokers in single process.
```python
import kombu

>>> with kombu.Connection('redis://localhost') as conn:
...     with conn.channel() as channel:
...         queue = kombu.Queue('queue', kombu.Exchange('ex1'), routing_key='queue')
...         producer = kombu.Producer(channel, exchange=queue.exchange)
...         producer.publish({'hello': 'world'}, retry=True,
...             routing_key=queue.name,
...             declare=[queue]
...         )
...
>>> # Here we see Connection state coming from previous connection. We should not see it since we are connecting
>>> # to a broker with different option. The same holds also for connecting to totally different broker.
>>> kombu.Connection('redis://localhost', transport_options={'global_keyprefix':'test'}).transport.state.exchanges
{'ex1': {'type': 'direct', 'durable': True, 'auto_delete': False, 'arguments': {}, 'table': [('queue', None, 'queue')]}}
>>> # Now to show the exception we try to connect the different borker and publish a message. The publishing should be 
>>> # successful but due non-existing exchange it fails.
>>> with kombu.Connection('redis://localhost', transport_options={'global_keyprefix':'test'}) as conn:
...     with conn.channel() as channel:
...         queue2 = kombu.Queue('queue', kombu.Exchange('ex1'), routing_key='queue')
...         producer = kombu.Producer(channel, exchange=queue2.exchange)
...         producer.publish({'hello': 'world'}, retry=True,
...             routing_key=queue2.name,
...             declare=[queue2]
...         )
...
Traceback (most recent call last):
  File "/home/matus/dev/kombu/kombu/connection.py", line 447, in _reraise_as_library_errors
    yield
  File "/home/matus/dev/kombu/kombu/connection.py", line 524, in _ensured
    return fun(*args, **kwargs)
  File "/home/matus/dev/kombu/kombu/messaging.py", line 199, in _publish
    return channel.basic_publish(
  File "/home/matus/dev/kombu/kombu/transport/virtual/base.py", line 600, in basic_publish
[...]

PR #1378, opened 2021-08-29
Make BrokerState Transport specific

matusvalo:
This PR changes BrokerState from being global to be Transport specific. This fixes issues when client is reconnecting and old BrokerState is used not matching the actual broker state. Memory and filesystem transport are still keeping global state since the implementation is not compatibile with per-transport broker state. Pyro transport is kept with global broker state for backward compatibility. Pyro transport should be retested and ported to per-transport state.

Fixes: #1377

### DIFF
--- a/kombu/transport/filesystem.py
+++ b/kombu/transport/filesystem.py
@@ -271,10 +271,15 @@ class Transport(virtual.Transport):
     """Filesystem Transport."""
 
     Channel = Channel
-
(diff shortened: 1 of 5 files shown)


matusvalo:
Only unittests are pending...

matusvalo:
Unittest added, ready for review.

auvipy (approved):

PR #1381, opened 2021-09-07
stop mentioning librabbitmq

auvipy:


TheBubblePopped:
Hi there,

why did this get removed?

auvipy:
py-amqp is the recommended package. librabbit do not offer the performance we thought it would be. we will deprecate it in near future.

PR #1384, opened 2021-09-08
Make BrokerState Transport specific

graingert:


auvipy:
we no longer maintain 4.x branch.

PR #1394, opened 2021-09-29
Prevent caching of oid in pidbox

### LINKED ISSUE #1063: Table empty or key no longer exists

matusvalo:
oid is not cached anymore due race conditions of oid in celery. Caching
oid is causing following exception:

```
OperationalError("
Cannot route message for exchange 'reply.celery.pidbox': Table empty or key no longer exists.
Probably the key ('_kombu.binding.reply.celery.pidbox') has been removed from the Redis database.
",)
```

This exception seems to be occuring when multiple celery workers
contains same oid.

Fixes #1063

### DIFF
--- a/kombu/pidbox.py
+++ b/kombu/pidbox.py
@@ -6,7 +6,6 @@ from collections import defaultdict, deque
 from contextlib import contextmanager
 from copy import copy
 from itertools import count
-from threading import local
 from time import time
 
 from . import Consumer, Exchange, Producer, Queue
@@ -187,7 +186,6 @@ class Mailbox:
         self.clock = LamportClock() if clock is None else clock
         self.exchange = self._get_exchange(self.namespace, self.type)
         self.reply_exchange = self._get_reply_exchange(self.namespace)
-        self._tls = local()
         self.unclaimed = defaultdict(deque)
         self.accept = self.accept if accept is None else accept
         self.serializer = self.serializer if serializer is None else serializer
@@ -404,13 +402,9 @@ class Mailbox:
                         durable=False,
                         delivery_mode='transient')
 
-    @cached_property
+    @property
     def oid(self):
-        try:
-            return self._tls.OID
-        except AttributeError:
-            oid = self._tls.OID = oid_from(self)
-            return oid
+        return oid_from(self)
 
     @cached_property
     def producer_pool(self):


matusvalo:
Marking as draft for now. Let's wait for confirmation of this fix.

auvipy:
looks good

thedrow:
Hold up, how does this work if the OID changes every time?
How do multiple workers end up with the same OID?

thedrow (changes requested):
I still have questions about this.

matusvalo:
> Hold up, how does this work if the OID changes every time?

OID won't change everytime. The generation of OID is bound to CPU, thread and object ID - see 

https://github.com/celery/kombu/blob/5ef5e22638035a6c412e949a2fbc5d44b7b088b2/kombu/common.py#L49-L55

Hence in the same process and thread and for each pidbox object the `oid_from()` function will return **the same** oid.

> How do multiple workers end up with the same OID?

The problem is with caching. The cache is stored in the heap and hence it is replicated across processes during fork() when new worker is spawned. Even thread local storage won't help -see following proof:

```python
import threading
import multiprocessing
mydata = threading.local()
mydata.x = 1

def f():
    print('f: ', mydata.x)

print('main process:', mydata.x)
t1 = threading.Thread(target=f)
t2 = threading.Thread(target=f)
print('starting t1')
t1.start()
print('starting t2')
t2.start()

p = multiprocessing.Process(target=f)
print('starting p')
p.start()
```

Running this script will print one for subprocess and raise attribute error for each thread. It means subthreads of process has local variable but subprocess **not**.

The same holds for pidbox oid - see example below:
```python
import multiprocessing
from kombu import pidbox

m = pidbox.Mailbox('unittest')
print(m.oid)

def getoid():
    print('getoid: ', m.oid)

p = multiprocessing.Process(target=getoid)
print('starting p')
p.start()

print(m.oid)
```

In this case both master process and subprocess **will share the oid** which is wrong:

```python
>>> import multiprocessing
>>> from kombu import pidbox
>>>
>>> m = pidbox.Mailbox('unittest')
>>> print(m.oid)
b8001cb8-1ae5-36f5-87f4-2660008969ec
>>>
>>> def getoid():
...     print('getoid: ', m.oid)
...
>>> p = multiprocessing.Process(target=getoid)
>>> print('starting p')
starting p
>>> p.start()
>>>
>>> print(m.oid)getoid:  b8001cb8-1ae5-36f5-87f4-2660008969ec

b8001cb8-1ae5-36f5-87f4-2660008969ec
>>>
```

> Each thread generates a new OID, so how there is a clash?

See my explanation above

thedrow (approved):
Thank you for the analysis.
This is a bug in `cached_property`'s implementation, it seems.

thedrow:
Note that the PyPy integration tests build is failing because we now only support 3.7 and above.

matusvalo:
> This is a bug in cached_property's implementation, it seems.

The bug in celery is due combination of cached_property, wrong usage of thread local storage.

I will add also unittests for this PR (99% done)

thedrow:
Oh, sorry.
I already merged the PR.
You can add the unit tests in another PR. :smile:

PR #1404, opened 2021-10-21
Return empty list instead of InconsistencyError when exchange table is empty

### LINKED ISSUE #1063: Table empty or key no longer exists

matusvalo:
Missing redis key containg set of queues bound to queue is caused
by removal all queues from the exchange. Hence, we should not raise an
exception but just return empty list of queues instead. This commit
fixes e.g. case publishing message against empty exchange:

```python
import kombu
conn = kombu.Connection('redis://')
exchange = kombu.Exchange('name', type='direct')
exchange.declare(channel=conn.default_channel)

producer = conn.Producer()
producer.publish(
     {'hello': 'world'},
     exchange=exchange,
     routing_key='queue1'
)
```

But it also fixes the case when last queue is unbound from exchange and
after publishing to this exchange:

```python
import kombu
conn = kombu.Connection('redis://')
exchange = kombu.Exchange('name', type='direct')
queue1 = kombu.Queue('queue1', exchange=exchange, routing_key='queue1')
exchange.declare(channel=conn.default_channel)
queue1 = queue1.bind(channel=conn.default_channel)
queue1.declare()

producer = conn.Producer()
producer.publish(
     {'hello': 'world'},
     exchange=exchange,
     routing_key='queue1'
)

queue1.delete()

producer.publish(
     {'hello': 'world'},
     exchange=exchange,
     routing_key='queue1'
)
```

Fixes #1063

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -100,10 +100,6 @@ error_classes_t = namedtuple('error_classes_t', (
     'connection_errors', 'channel_errors',
 ))
 
-NO_ROUTE_ERROR = """
-Cannot route message for exchange {0!r}: Table empty or key no longer exists.
-Probably the key ({1!r}) has been removed from the Redis database.
-"""
 
 # This implementation may seem overly complex, but I assure you there is
 # a good reason for doing it this way.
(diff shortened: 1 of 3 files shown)


auvipy:
yeah but im talking about other integrations tests. the tools that run CI needs some minor tweaks. like pypy should be pypy-3.7 etc. I have to check docs and try to fix them this week

PR #1427, opened 2021-11-07
python !=3.12.1

auvipy:


auvipy (commented):
all green except pypy which needs CI adjustment later

PR #1428, opened 2021-11-08
try latest sqs dependencies

auvipy:


auvipy (commented):
now passing

sparrowt:
I ended up here because `safety` is now reporting `kombu<5.2.1` as vulnerable - could you perhaps provide some more detail please?

Under "kombu" in https://github.com/pyupio/safety-db/blob/master/data/insecure_full.json (which references this change) it mentions `PVE-2021-42497` however I can't seem to find anything about that online - how/where does a "PVE" get allocated and what is the actual vulnerability which is now causing older kombu versions to be considered vulnerable?

requirements/extras/sqs.txt
```diff
@@ -1,3 +1,3 @@
-boto3>=1.4.4
-pycurl==7.43.0.2  # Latest build with wheels provided
-urllib3<1.26      # Unittests are faiing with urllib3>=1.28
+boto3>=1.9.12
+pycurl~=7.44.1  # Latest build with wheels provided
+urllib3>=1.26.7    # Unittests are faiing with urllib3>=1.28
```
sparrowt, line 3: Is the comment here still true regarding issues with `urllib3>=1.28`? If so should there still be an upper bound on this (if not the comment can be removed)

felipsmartins:
Same question asked for @sparrowt: What is the vulnerability? I have kombu `4.6.7` and celery `4.4.0` installed but I'm not sure whether upgrading the packages will break something or not. Is it upgrade safe?

Thanks in advance!

thedrow:
It should be safe to upgrade but since this is open source, your mileage may vary.

requirements/extras/sqs.txt
auvipy: https://pypi.org/project/urllib3/#history says there were no 1.28 and its a typo. and should be 1.26. if you do no use windows, you should be pretty safe to upgrade as pycurl latest dont provide any wheel in pypi. you/we have to generate pycurl wheel for windows.

felipsmartins:
Thank you!

Alex-ley:
The latest build with wheels is now 7.45.1 I believe: https://github.com/pycurl/pycurl/tags https://pypi.org/project/pycurl/

What would need to be done to modify the [requirements/extras/sqs.txt](https://github.com/celery/kombu/pull/1428/files/a77e70d9dec46e85ee5c210151a010cda65db242#diff-d229b45a35f019d718901cd33e806e8d2fbecb8eb66d81fb00f8e481c56ea0b4) to read something like:
```
boto3>=1.9.12
pycurl~=7.45.1  # Latest build with wheels provided
urllib3>=1.26.7    # Unittests are faiing with urllib3>=1.28
```

Is there anything the community could do for that? Cheers.

auvipy:
are you sure latest release provide wheel? if so I will use them

auvipy:
se https://github.com/celery/celery/pull/7429 windows CI is failing

Alex-ley:
my bad: https://github.com/pycurl/pycurl/issues/657

PR #1442, opened 2021-11-21
drop as we dont use travis

auvipy:


auvipy:
pypy failures are not related

PR #1444, opened 2021-11-23
Ensure that restore is atomic in redis transport

### LINKED ISSUE #1405: Tasks Sporadically Lost in Redis Broker
Hi, we've recently started running celery/kombu in an environment with autoscaling workers and with Redis as a broker. 
What we've noticed is that tasks mysteriously disappear from time to time even though we are using acks_late.

I'm suspecting that they may be getting lost somewhere in the process of moving tasks from `unacked` back onto a queue (after the visibility timeout)

Here's the line of code that schedules this periodic task
https://github.com/celery/kombu/blob/master/kombu/transport/redis.py#L1242

and the function that does the actual restore:
https://github.com/celery/kombu/blob/master/kombu/transport/redis.py#L359-L366

I believe an issue may arise if a worker is in the process of restoring tasks, and receives a SIGTERM. Since `restore_by_tag` isn't pipelining the entire set of Redis commands, tasks may potentially be lost if they are in the middle of being moved.

matusvalo:
Fixes #1405


matusvalo:
This PR is just preview. Not tested yet, tests to be updated.

matusvalo:
PR ready for review.

auvipy:
I was away for vacaton

auvipy:
there is a regression report

thedrow:
Where?

auvipy:
here but need to verify https://github.com/celery/kombu/issues/1568

PR #1450, opened 2021-11-30
SQS transport: detect FIFO queue properly by checking queue URL

### LINKED ISSUE #1449: Failure to send message to SQS FIFO queue when using `predefined_queues`
# Preconditions
* use SQS transport
* FIFO queue must be created in SQS
* use `predefined_queues` setting (queue name MUST not end with `.fifo` and queue URL naturally ends with `.fifo`) as per [documentation](https://docs.celeryproject.org/projects/kombu/en/stable/reference/kombu.transport.SQS.html#predefined-queues), e.g.
```python
'predefined_queues': {
    'celery': {
        'url': 'https://sqs.us-east-1.amazonaws.com/xxx/queue.fifo',
        'access_key_id': 'a',
        'secret_access_key': 'b',
    }
}
```

# Actual result
Sending any message will fail because it will be missing `MessageGroupId` attribute and it is required for FIFO queues.
This happens because such configuration in `predefined_queues` is not treated as a FIFO queue based on the queue name while the actual queue is FIFO:

https://github.com/celery/kombu/blob/9a91e8b7807aa634031e369f05682df4820766ce/kombu/transport/SQS.py#L372-L382

# Expected result
Since nothing prevents to have a predefined FIFO queue with a name that does not end with `.fifo` the easiest way to fix would be to detect FIFO queues by checking queue URL and not the name (which is just a routing key).

Other way would be to validate the `predefined_queues` and force user to name FIFO queues properly.

n0061q:
Fixes #1449

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -59,8 +59,8 @@ exist in AWS) you can tell this transport about them as follows:
           'backoff_policy': {1: 10, 2: 20, 3: 40, 4: 80, 5: 320, 6: 640}, # optional
           'backoff_tasks': ['svc.tasks.tasks.task1'] # optional
         },
-        'queue-2': {
-          'url': 'https://sqs.us-east-1.amazonaws.com/xxx/bbb',
+        'queue-2.fifo': {
+          'url': 'https://sqs.us-east-1.amazonaws.com/xxx/bbb.fifo',
           'access_key_id': 'c',
           'secret_access_key': 'd',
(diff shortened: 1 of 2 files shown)


matusvalo (changes requested):
Thank you @n0061q  for your PR.

Let first finish discussion in bug report first to understand how we can fix the issue.

matusvalo (approved):

matusvalo:
Thank you @n0061q for your PR!

PR #1466, opened 2021-12-20
try pining setuptools

auvipy:

PR #1467, opened 2021-12-20
prevent KeyError: 'purelib' 

pawl:
This might be another potential fix for the issues in: https://github.com/celery/kombu/pull/1462

Currently we're seeing this error:
```
File "setup.py", line 60, in
scheme['data'] = scheme['purelib']
KeyError: 'purelib'
```

This seems to describe what the code with the error is doing: https://stackoverflow.com/a/3042436

That code was added in the initial commit for kombu and `data_files` is no longer defined anymore. So, maybe the hack to copy data files to the module directory isn't necessary anymore? Or maybe this should be handled in `MANIFEST.in` instead?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,6 @@
 import os
 import re
 import sys
-from distutils.command.install import INSTALL_SCHEMES
 
 import setuptools
 import setuptools.command.test
@@ -55,9 +54,6 @@ def fullsplit(path, result=None):
         return result
     return fullsplit(head, [tail] + result)
 
-
-for scheme in list(INSTALL_SCHEMES.values()):
-    scheme['data'] = scheme['purelib']
 
 # if os.path.exists('README.rst'):
 #    long_description = codecs.open('README.rst', 'r', 'utf-8').read()


pawl:
The root cause of this issue seems to be this commit: https://github.com/pypa/setuptools/commit/f32af4e73a3f994f3a76898bb404919c5722508d

That new scheme that got added didn't have purelib defined.

auvipy:
I was thinking of removing this as well but should we open an issue in setup tolls as well?

auvipy:
https://github.com/celery/kombu/pull/1466/files doesn't seem to work

pawl:
@auvipy I created an issue with setuptools here: https://github.com/pypa/setuptools/issues/2936

pawl:
@auvipy the setuptools maintainer suggests:
> My preference would be for the hack to adjust its expectations or to remove the hack altogether if possible, but I realize that may not be possible in all scenarios.

So maybe we should go ahead with this?

We could also potentially compare the files from the packages built before and after this change to be extra careful.

auvipy:
what about using 

try adding this to your pip install:

SETUPTOOLS_USE_DISTUTILS=stdlib pip install .... 

As mentioned from setuptools issue

auvipy:
I will also come back to PR, but need some more careful analysis before merging this :D

matusvalo (approved):
Based what I read in https://stackoverflow.com/questions/1612733/including-non-python-files-with-setup-py/3042436#3042436 we should be fine since we are not using `data_files` as mentioned in the link right? But to be honest this topic is not my expertise :-(

auvipy:
> Based what I read in https://stackoverflow.com/questions/1612733/including-non-python-files-with-setup-py/3042436#3042436 we should be fine since we are not using `data_files` as mentioned in the link right? But to be honest this topic is not my expertise :-(

no worries. I have been studying these type of topics  for last few weeks.

pawl:
@matusvalo @auvipy I just compared the contents of `site-packages/kombu-5.2.3-py3.8.egg/` before and after this change (with the most recent changes from master), and I confirmed this change doesn't affect which files end up in the package. I also created a fresh virtualenv for each `python setup.py install`. 

Here's the output I was looking at: https://gist.github.com/pawl/4828c02abba078a154bf7b77f22ba0dd

So, I think that confirms the lines this PR removes are no longer necessary since we're not using `data_files`.

auvipy (approved):

auvipy:
thank you @pawl

PR #377, opened 2014-07-05
Added pattern/matcher to Mailbox

malinoff:
Refs [celery#2122](https://github.com/celery/celery/issues/2122)

PR #1470, opened 2021-12-21
reduce memory usage of Transport

pawl:
While investigating a redis broker memory leak, I noticed the same issue as py-amqp's [#377](https://github.com/celery/py-amqp/pull/377):
```
Top 10 lines
#1: /celery_app/kombu/kombu/transport/virtual/base.py:911: 782.4 KiB
    ARRAY_TYPE_H, range(self.channel_max, 0, -1),
<truncated>
```

The array that's used to store available connection ids is taking up an unnecessary amount of memory (more than anything else while running celery), and it's making memory leaks worse.

This PR reduces the memory usage of the `Transport` by reversing that logic and turning `_avail_channel_ids` into `_used_channel_ids`. This will only store the channel ids that are currently used, which is almost always going to be a much smaller array.

### DIFF
--- a/kombu/transport/virtual/base.py
+++ b/kombu/transport/virtual/base.py
@@ -462,21 +462,7 @@ class Channel(AbstractChannel, base.StdChannel):
             typ: cls(self) for typ, cls in self.exchange_types.items()
         }
 
-        # Cast to a set for fast lookups, and keep stored as an array
-        # for lower memory usage.
-        used_channel_ids = set(self.connection._used_channel_ids)
-
-        for channel_id in range(1, self.connection.channel_max + 1):
-            if channel_id not in used_channel_ids:
-                self.connection._used_channel_ids.append(channel_id)
-                self.channel_id = channel_id
-                break
-        else:
-            raise ResourceError(
-                'No free channel ids, current={}, channel_max={}'.format(
-                    len(self.connection.channels),
-                    self.connection.channel_max), (20, 10),
-            )
+        self.channel_id = self._get_free_channel_id()
 
         topts = self.connection.client.transport_options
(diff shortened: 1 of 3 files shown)


kombu/transport/virtual/base.py
```diff
@@ -462,9 +462,15 @@ def __init__(self, connection, **kwargs):
             typ: cls(self) for typ, cls in self.exchange_types.items()
         }
 
-        try:
-            self.channel_id = self.connection._avail_channel_ids.pop()
-        except IndexError:
+        # Cast to a set for fast lookups, and keep stored as an array for lower memory usage.
+        used_channel_ids = set(self.connection._used_channel_ids)
```
pawl, line 466: Maybe we should put this `_used_channel_ids` stuff into its own method like it is in py-amqp?: https://github.com/celery/py-amqp/blob/6433be13199938f47bbc2826e6585e746baf4f9b/amqp/connection.py#L484-L492

auvipy: if that improves the codebase you could give it a try

auvipy:
also can you and your team please check the two open PR in pyamqp with this two PR in kombu & celery that nothing create any regression? I like your efforts very much, just being too cautious as it's holiday season coming.

kombu/transport/virtual/base.py
pawl: Ok, I went ahead and made this change: https://github.com/celery/kombu/pull/1470/commits/a439adaceae6c7e56e0d7c192ce63f2434a079e8

auvipy:
@michael-lazar

auvipy (approved):

PR #1471, opened 2021-12-22
allow getting recoverable_connection_errors without an active transport 

pawl:
I was looking into solutions for the [celery + redis memory leak issue](https://github.com/celery/celery/issues/4843#issuecomment-999168967) and I was attempting to reduce the references to initialized `Transport`s.

One thing I noticed was that you currently need at least one initialized `Transport` with an active connection to allow filling the cache for `recoverable_connection_errors`/`recoverable_channel_errors`/`connection_errors`/`channel_errors`. It seems like it makes more sense to get the errors from the `Transport` class without needing to initialize it (in case there are connection errors on the first `Transport` initialization).

This doesn't fix the memory leak issue, but maybe this change is still worth making?

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -932,7 +932,7 @@ class Connection:
         but where the connection must be closed and re-established first.
         """
         try:
-            return self.transport.recoverable_connection_errors
+            return self.get_transport_cls().recoverable_connection_errors
         except AttributeError:
             # There were no such classification before,
             # and all errors were assumed to be recoverable,
@@ -948,19 +948,19 @@ class Connection:
         recovered from without re-establishing the connection.
         """
         try:
-            return self.transport.recoverable_channel_errors
+            return self.get_transport_cls().recoverable_channel_errors
         except AttributeError:
             return ()
 
     @cached_property
     def connection_errors(self):
         """List of exceptions that may be raised by the connection."""
-        return self.transport.connection_errors
+        return self.get_transport_cls().connection_errors
 
     @cached_property
     def channel_errors(self):
         """List of exceptions that may be raised by the channel."""
-        return self.transport.channel_errors
+        return self.get_transport_cls().channel_errors
 
     @property
     def supports_heartbeats(self):
--- a/kombu/transport/consul.py
+++ b/kombu/transport/consul.py
@@ -276,23 +276,24 @@ class Transport(virtual.Transport):
     driver_type = 'consul'
     driver_name = 'consul'
 
-    def __init__(self, *args, **kwargs):
-        if consul is None:
-            raise ImportError('Missing python-consul library')
-
-        super().__init__(*args, **kwargs)
-
-        self.connection_errors = (
+    if consul:
+        connection_errors = (
             virtual.Transport.connection_errors + (
                 consul.ConsulException, consul.base.ConsulException
             )
         )
 
-        self.channel_errors = (
+        channel_errors = (
             virtual.Transport.channel_errors + (
                 consul.ConsulException, consul.base.ConsulException
             )
         )
+
+    def __init__(self, *args, **kwargs):
+        if consul is None:
+            raise ImportError('Missing python-consul library')
+
+        super().__init__(*args, **kwargs)
 
     def verify_connection(self, connection):
         port = connection.client.port or self.default_port
--- a/t/unit/test_connection.py
+++ b/t/unit/test_connection.py
@@ -574,7 +574,9 @@ class test_Connection:
         assert conn.channel_errors == (KeyError, ValueError)
 
     def test_channel_errors__exception_no_cache(self):
-        """Ensure the channel_errors can be retrieved without an initialized transport."""
+        """Ensure the channel_errors can be retrieved without an initialized
+        transport.
+        """
 
         class MyTransport(Transport):
             channel_errors = (KeyError,)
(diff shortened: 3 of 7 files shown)


kombu/transport/consul.py
```diff
@@ -276,24 +276,25 @@ class Transport(virtual.Transport):
...
     driver_name = 'consul'
 
-    def __init__(self, *args, **kwargs):
-        if consul is None:
-            raise ImportError('Missing python-consul library')
-
-        super().__init__(*args, **kwargs)
-
-        self.connection_errors = (
+    if consul:
```
auvipy, line 279: only 3 transports are subject to this adjustments?

t/unit/test_connection.py
```diff
@@ -571,6 +573,16 @@ class MyTransport(Transport):
         conn = Connection(transport=MyTransport)
         assert conn.channel_errors == (KeyError, ValueError)
 
+    def test_channel_errors__exception_no_cache(self):
```
auvipy, line 576: beside the unit tests I would love to see integration tests for this

auvipy (changes requested):

kombu/transport/consul.py
pawl: Yeah, I did a search for `self.connection_errors` and `self.channel_errors`, and looked at each of the transports.

For some reason, the librabbitmq and mongodb backends don't have the `try/except ImportError:` logic and they don't use `__init__` to fill in `connection_errors` and `channel_errors`.

auvipy (changes requested):
with pytest.raises(KeyError):
            self.channel._receive_one(self.channel.subclient)
>       assert not self.channel._in_listen
E       assert not True
E        +  where True = <t.unit.transport.test_redis.Channel object at 0x7f8bf8d594d0>._in_listen
E        +    where <t.unit.transport.test_redis.Channel object at 0x7f8bf8d594d0> = <t.unit.transport.test_redis.test_Channel object at 0x7f8bf8d591d0>.channel

t/unit/transport/test_redis.py:582: AssertionError

auvipy (changes requested):
I would like to take some time & more eyes and trials before merging this. in the mean the you can push any improvements in mind. and most possibly the more integration tests the better IMHO

auvipy:
besides, do we need any docs changes to mention the adjustments being made?

auvipy:
@ michael-lazar

auvipy:
@pawl did you try this in any project with celery? i want to have some integration test for this

pawl:
@auvipy I added an integration test here: https://github.com/celery/kombu/pull/1471/commits/448cdb9a4bf3364eef4025da69f339e954b116ea

> do we need any docs changes to mention the adjustments being made?

Oops, should I be updating the changelog?

I'm not aware of any documentation changes since old code should continue to work the same way, because the errors are still a class attribute on the same class.

I'm not sure we need to document the removal of `_get_errors` outside of the release notes since it's an internal method. It also doesn't look like it's overridden in any open source projects: https://github.com/search?q=transport+_get_errors+NOT+MultiChannelPoller+NOT+test_purge+NOT+%22by+Ask+Solem%22&type=Code

pawl:
> did you try this in any project with celery?

I've tested the change with [this branch](https://github.com/pawl/celery_pyamqp_memory_leak/tree/test_errors_fix) of my celery memory leak example repo. It's still able to recover from connection errors successfully after I restart redis. Also, it connects successfully on start-up.

auvipy:
ok we are good to go

auvipy (approved):

auvipy:
decided to land this change on 5.3 for now, as it changes some behaviours, but if other core team members agree will end landing in 5.2.x bug fix release

kombu/connection.py
```diff
@@ -948,19 +948,19 @@ def recoverable_channel_errors(self):
         recovered from without re-establishing the connection.
         """
         try:
-            return self.transport.recoverable_channel_errors
+            return self.get_transport_cls().recoverable_channel_errors
```
matusvalo, line 951: Is it correct to put parameter of `get_transport_cls()` as None?

kombu/connection.py
```diff
@@ -948,19 +948,19 @@ def recoverable_channel_errors(self):
...
         except AttributeError:
             return ()
 
     @cached_property
     def connection_errors(self):
         """List of exceptions that may be raised by the connection."""
-        return self.transport.connection_errors
+        return self.get_transport_cls().connection_errors
```
matusvalo, line 958: Here the same.

kombu/connection.py
```diff
@@ -948,19 +948,19 @@ def recoverable_channel_errors(self):
...
 
     @cached_property
     def channel_errors(self):
         """List of exceptions that may be raised by the channel."""
-        return self.transport.channel_errors
+        return self.get_transport_cls().channel_errors
```
matusvalo, line 963: Here the same.

matusvalo (changes requested):
I am not 100% sure but if I read correctly the code `get_transport_cls()` needs a transport as string or type. Default value is `None` which I am not sure will work.

https://github.com/celery/kombu/blob/b6b44085755c121fbcc8d954bd8f62ebe56ebc12/kombu/transport/__init__.py#L47-L86

pawl:
@matusvalo `get_transport_cls` is already used without arguments in `create_transport` (used by the `transport` property): https://github.com/celery/kombu/blob/a9c4f9837c1ad2296a9b21311f79c8cdbddade6b/kombu/connection.py#L606

It looks like `_init_params` will populate `self.transport_cls` and it will be used by `get_transport_cls`.

matusvalo:
> @matusvalo `get_transport_cls` is already used without arguments in `create_transport` (used by the `transport` property):
> 
> https://github.com/celery/kombu/blob/a9c4f9837c1ad2296a9b21311f79c8cdbddade6b/kombu/connection.py#L606
> 
> It looks like `_init_params` will populate `self.transport_cls` and it will be used by `get_transport_cls`.

Yeah you are right :-/. I missed that it is wrapped in the same class. Please ignore my comment.

matusvalo (approved):

PR #1477, opened 2021-12-24
prevent redis event loop stopping on 'consumer: Cannot connect'

pawl:
I did some more testing with the change in #1476 and noticed that the event loop will stop polling after a "consumer: Cannot connect to redis" error from `Consumer.ensure_connected._error_handler`. This will cause celery to stop processing work. You need to stop redis for more than a few seconds to make this happen.

It can still re-connect when Redis restarts after the event loop already started, but the error above seems to happen when it's starting the `Consumer` and before `on_poll_start` runs for the first time on the new consumer.

This PR fixes the issue by checking whether event loop polling has started (based on file descriptors) and will only remove `on_poll_start` from the event loop if polling has started.

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -397,6 +397,9 @@ class MultiChannelPoller:
     #: Set of one-shot callbacks to call after reading from socket.
     after_read = None
 
+    #: Set by :meth:`on_poll_start` after successfully running once.
+    poll_started = False
+
     def __init__(self):
         # active channels
         self._channels = set()
@@ -474,6 +477,7 @@ class MultiChannelPoller:
                     self._register_BRPOP(channel)
             if channel.active_fanout_queues:    # LISTEN mode?
                 self._register_LISTEN(channel)
+        self.poll_started = True
 
     def on_poll_init(self, poller):
         self.poller = poller
@@ -1238,11 +1242,13 @@ class Transport(virtual.Transport):
             if connection._sock:
                 loop.remove(connection._sock)
 
-            # stop polling in the event loop
-            try:
-                loop.on_tick.remove(on_poll_start)
-            except KeyError:
-                pass
+            # must have started polling or this will break reconnection
+            if cycle.poll_started:
+                # stop polling in the event loop
+                try:
+                    loop.on_tick.remove(on_poll_start)
+                except KeyError:
+                    pass
         cycle._on_connection_disconnect = _on_disconnect
 
         def on_poll_start():
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -397,9 +397,6 @@ class MultiChannelPoller:
     #: Set of one-shot callbacks to call after reading from socket.
     after_read = None
 
-    #: Set by :meth:`on_poll_start` after successfully running once.
-    poll_started = False
-
     def __init__(self):
         # active channels
         self._channels = set()
@@ -477,7 +474,6 @@ class MultiChannelPoller:
                     self._register_BRPOP(channel)
             if channel.active_fanout_queues:    # LISTEN mode?
                 self._register_LISTEN(channel)
-        self.poll_started = True
 
     def on_poll_init(self, poller):
         self.poller = poller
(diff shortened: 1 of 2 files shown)


auvipy (approved):

auvipy:
you want to add anything to this?

auvipy:
did you check with 2 commit here which is not on pypi yet? https://github.com/celery/py-amqp/commits/master

pawl:
@auvipy 

> did you check with 2 commit here which is not on pypi yet? https://github.com/celery/py-amqp/commits/master

This change only affects redis broker code, so those py-amqp changes shouldn't matter for this bug.

However, I tried to reproduce the same issue by stopping rabbitmq for a while on (this branch)[https://github.com/pawl/celery_pyamqp_memory_leak/tree/test_errors_fix] of my memory leak example (with the most recent code from py-amqp), and it's able to re-connect successfully.

> you want to add anything to this?

I don't have plans to add anything else at the moment, unless you have some suggestions.

fcovatti:
@pawl is this somehow should fix a issue I am experiencing https://github.com/celery/celery/discussions/7276
I am testing on celery 5.2.3 and still experiencing a issue where the worker stops consuming tasks, but it does not happen in every redis restart. Many thanks

pawl:
@fcovatti This was a fix to a specific issue from #1476 that never made it into a release by itself. So, I doubt this will fix [your issue](https://github.com/celery/celery/discussions/7276).

I've had similar issues with workers not processing work, and I've ended up enabling `without_gossip` and `without_heartbeat` too. I can't remember if that's what fixed it though. The sentry project does the [same thing](https://github.com/getsentry/sentry/blob/fc50e54b38d44015102c53e7503a4ba667e17285/src/sentry/runner/commands/run.py#L157-L162) and I think they've got a pretty battle-tested setup. CloudAMQP's [suggested settings](https://www.cloudamqp.com/docs/celery.html#commandline-arguments
) are similar too.
 
Your issue might have something to do with the worker losing a connection to the broker and not reconnecting automatically. This comment also has a potential explanation: https://github.com/celery/celery/issues/5037#issuecomment-508028060

I think it would be a good contribution to add more debug logging (especially around broker connection errors) to make these sorts of issues easier to figure out.

thedrow:
@pawl I strongly agree that Celery lacks debug logs.

zyhgithub:
> I did some more testing with the change in #1476 and noticed that the event loop will stop polling after a "consumer: Cannot connect to redis" error from `Consumer.ensure_connected._error_handler`. This will cause celery to stop processing work. You need to stop redis for more than a few seconds to make this happen.
> 
> It can still re-connect when Redis restarts after the event loop already started, but the error above seems to happen when it's starting the `Consumer` and before `on_poll_start` runs for the first time on the new consumer.
> 
> This PR fixes the issue by checking whether event loop polling has started (based on file descriptors) and will only remove `on_poll_start` from the event loop if polling has started.

This change may cause loop.on_tick to be an empty set when _on_disconnect function implements. When loop.on_tick  is an empty set, there is no fds in kombu.asynchronous.hub.Hub#create_loop function to poll, then celery worker will get no task to consume. This can reappear when broker（redis） lost and reconnect with 30s interval when celery worker with threads mode. By the way, I find that python GC will cause redis.connection.PythonParser#__del__ called and then  _on_disconnect function in kombu.transport.redis.Transport#register_with_event_loop will be called to.
I think code loop.on_tick.remove(on_poll_start) is not need to do, because kombu.transport.redis.MultiChannelPoller#close clear   the cycle.fds.

auvipy:
@zyhgithub can you please share/contribute a failing testcase? or do you have anything else to share with us to improve the codes?

zyhgithub:
I think the commit (https://github.com/celery/kombu/pull/1477) is no need to do. This commit will cause that there is no fd for gevent hub to poll so that  the celery worker will get no task to consume when broker（redis） lost and reconnect after 30s as celery worker in threads mode.

image

auvipy:
do you have any better fix? because we can not simply revert that

zyhgithub:
I don't know the reason for the commit (https://github.com/celery/kombu/pull/1477). Could you tell me what problems the commit resolved? 3ks.

auvipy:
> I don't know the reason for the commit (#1477). Could you tell me what problems the commit resolved? 3ks.

did you checked the tests?

PR #1481, opened 2022-01-12
Revert "try pining setuptools (#1466)"

pawl:
This reverts commit 6ddf9409401e290e267c9c5041d92fdf5f47170d.

Since #1467 was merged, we no longer need to pin setuptools to `<59.7.0` to avoid this error:
```
File "setup.py", line 60, in
scheme['data'] = scheme['purelib']
KeyError: 'purelib'
```

This also fixes these issues:
* https://github.com/celery/celery/discussions/7202
* https://github.com/celery/celery/pull/7194#issuecomment-1002928015
* https://github.com/celery/celery/commit/d4b97bedc79aed0b45dd3720b683d8d8572da2a9#r63108656

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,7 +18,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
-        run: pip install --upgrade pip 'setuptools<59.7.0' wheel tox tox-docker
+        run: pip install --upgrade pip setuptools wheel tox tox-docker
       - name: Run unittest
         run: tox -v -e ${{ matrix.python-version }}-linux-unit -- -v
   #################### Integration tests ####################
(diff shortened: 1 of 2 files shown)


auvipy:
what is the least safer version? like with python 3.6+ only?

pawl:
@auvipy looks like version v47.0.0:
> #2094: Setuptools now actively crashes under Python 2. Python 3.5 or later is required. Users of Python 2 should use setuptools<45.

Source: https://setuptools.pypa.io/en/latest/history.html#v47-0-0

I'll go ahead and switch it to requiring at least that version.

auvipy:
I am OK with v47/48, @jaraco can you please shed some light on the least safe version to be used?

jaraco:
I don’t understand the question. Can you describe what you’re trying to achieve with “safe”?

auvipy:
the minimal version of setuptools we should enforce in 2022

auvipy (approved):

PR #1482, opened 2022-01-22
The times of retrying is not correct

lemonshir:
The times of retrying was 'max_retries' plus 1 which is fixed now.


auvipy (approved):

lemonshir:
No other related issues were detected from my side.

thedrow:
There should be a unit test to ensure this doesn't regress. @lemonshir Mind writing one?

PR #1483, opened 2022-01-25
Remove use of OrderedDict in various places

atombrella:


kombu/connection.py
```diff
@@ -658,7 +657,7 @@ def _info(self, resolve=True):
 
     def info(self):
         """Get connection info."""
-        return OrderedDict(self._info())
+        return dict(self._info())
```
auvipy, line 660: can you find out some more OrderedDict in the code base please?

auvipy (changes requested):

kombu/connection.py
atombrella: Will do. There are more.

kombu/transport/virtual/base.py
```diff
@@ -160,7 +160,7 @@ class QoS:
     #: current prefetch count value
     prefetch_count = 0
 
-    #: :class:`~collections.OrderedDict` of active messages.
```
atombrella, line 163: Should be reverted.

kombu/transport/virtual/base.py
```diff
@@ -177,6 +177,8 @@ def __init__(self, channel, prefetch_count=0):
         self.channel = channel
         self.prefetch_count = prefetch_count or 0
 
+        # Because we set an attribute on the OrderedDict, which
+        # is not possible with an ordinary dictionary
```
thedrow, line 181: Please rephrase the comment as it isn't using proper English.

thedrow (changes requested):

kombu/transport/virtual/base.py
atombrella: Is this better? I'd prefer not to trigger pipelines just for this.

```
        # Standard Python dictionaries do not support setting attributes
        # on the object, hence the use of OrderedDict
```

atombrella: @thedrow

atombrella: I pushed that change.

auvipy (approved):

PR #1484, opened 2022-01-28
Set redelivered property for Celery with Redis

Dehax:
Fixed setting `redelivered` value for Celery when Redis broker is used.

According to the Celery implementation, `redelivered` should be inside `delivery_info`:

- https://github.com/celery/celery/blob/25ca389039f667cdf940c9efe44bdefd1cf70f30/celery/app/trace.py#L406-L407
- https://github.com/celery/celery/blob/25ca389039f667cdf940c9efe44bdefd1cf70f30/celery/app/control.py#L358-L362
- https://github.com/celery/celery/blob/25ca389039f667cdf940c9efe44bdefd1cf70f30/celery/worker/request.py#L153-L157

It's also mentioned in the User Guide:
https://github.com/celery/celery/blob/25ca389039f667cdf940c9efe44bdefd1cf70f30/docs/userguide/tasks.rst#reject
```
from celery.exceptions import Reject

@app.task(bind=True, acks_late=True)
def requeues(self):
    if not self.request.delivery_info['redelivered']:
        raise Reject('no reason', requeue=True)
    print('received two times')
```

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -717,6 +717,7 @@ class Channel(virtual.Channel):
         try:
             try:
                 payload['headers']['redelivered'] = True
+                payload['properties']['delivery_info']['redelivered'] = True
             except KeyError:
                 pass
             for queue in self._lookup(exchange, routing_key):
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
could you please add test for the proposed change?

Dehax:
@auvipy Sure, I've added a test with a real payload from my Celery project.

auvipy:
thanks for handling this. I have started the build

auvipy (approved):

PR #1489, opened 2022-02-02
Protect set of ready tasks by lock to avoid concurrent updates

olii:
When there is no locking then there is a possibility that multiple threads manipulate with the same object at the same time.

The issue is manifested when the celery worker is run in the threading model and fails with the following exception:
```
RuntimeError: Set changed size during iteration
```

I don't think we can create reliable tests for this change.

See: https://github.com/celery/celery/issues/7162

### DIFF
--- a/kombu/asynchronous/hub.py
+++ b/kombu/asynchronous/hub.py
@@ -1,6 +1,7 @@
 """Event loop implementation."""
 
 import errno
+import threading
 from contextlib import contextmanager
 from queue import Empty
 from time import sleep
@@ -78,6 +79,7 @@ class Hub:
         self.on_tick = set()
         self.on_close = set()
         self._ready = set()
+        self._ready_lock = threading.Lock()
 
         self._running = False
         self._loop = None
@@ -198,7 +200,8 @@ class Hub:
     def call_soon(self, callback, *args):
         if not isinstance(callback, Thenable):
             callback = promise(callback, args)
-        self._ready.add(callback)
+        with self._ready_lock:
+            self._ready.add(callback)
         return callback
 
     def call_later(self, delay, callback, *args):
@@ -242,6 +245,12 @@ class Hub:
         except (AttributeError, KeyError, OSError):
             pass
 
+    def _pop_ready(self):
+        with self._ready_lock:
+            ready = self._ready
+            self._ready = set()
+            return ready
+
     def close(self, *args):
         [self._unregister(fd) for fd in self.readers]
         self.readers.clear()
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
Thanks for the PR, can you please add unit and integration test for the changes.

olii:
I guess we can write a unit test with mocked `threading.Lock()` and test if the lock was acquired and released but I don't think an integration test is possible. The race condition requires a context switch of at least 2 threads.

auvipy:
unit test is the minimal requirement so go with it please

thedrow:
> I guess we can write a unit test with mocked `threading.Lock()` and test if the lock was acquired and released but I don't think an integration test is possible. The race condition requires a context switch of at least 2 threads.

@olii If you synchronize a worker thread A using `condition.wait()` at the same point where a context switch should occur to reproduce the data race, if you know how it reproduces and by triggering `condition.notify()` at the end of the execution of worker thread B, thread A should raise the exception you're observing without the mutex you added.

kombu/asynchronous/hub.py
```diff
@@ -242,6 +245,12 @@ def _unregister(self, fd):
         except (AttributeError, KeyError, OSError):
             pass
 
+    def _pop_ready(self):
+        with self._ready_lock:
+            ready = self._ready
+            self._ready = set()
```
thedrow, line 251: Is this an optimization or a source of future problems?
A set is theoretically unordered but a list is ordered which is what we used before.

In any case, I think the main problem with this code is it's design.
ready should not be a set but a priority queue with unique items. See [implementation example](https://gist.github.com/Dobiasd/37705392b4aaa3a3539ba1a61efec6b6).
This data structure will allow us to re-order the execution of events later on. The bonus is that it is thread-safe and it doesn't pollute this class with more responsibilities it shouldn't have.

thedrow (changes requested):
I'd accept the PR if I have to but I think that there's a much better way to implement this PR which will resolve more problems in the future.

kombu/asynchronous/hub.py
olii: > In any case, I think the main problem with this code is it's design.

I agree. The reason I decided to take the approach in this PR is that it is a small change. I can't implement a better approach due to a lack of experience in Kombu.

olii: Feel free to close this PR and implement it in a much better way. Or merge this PR, release a bugfix and prepare a better implementation later.

thedrow: I think I see your point of view better now. It should work. I'm going to merge this.

Some promises have to be called as soon as possible since their ETA related for example so they have to be prioritized over regular tasks. I guess it doesn't matter now and we can open a ticket about this later

thedrow (approved):

Issue #1358, opened 2021-07-22
SQS worker errors our when Rejected


inai-shreyas:
I get the following error when I raise a Reject exception

'str' object has no attribute 'headers'

full stacktrace:
```
[2021-07-22 16:01:04,309: CRITICAL/MainProcess] Unrecoverable error: AttributeError("'str' object has no attribute 'headers'")
Traceback (most recent call last):
  File "/home/shreyas/.local/share/virtualenvs/inai-backend-mwrd66wz/lib/python3.8/site-packages/celery/worker/worker.py", line 203, in start
    self.blueprint.start(self)
  File "/home/shreyas/.local/share/virtualenvs/inai-backend-mwrd66wz/lib/python3.8/site-packages/celery/bootsteps.py", line 116, in start
    step.start(parent)
  File "/home/shreyas/.local/share/virtualenvs/inai-backend-mwrd66wz/lib/python3.8/site-packages/celery/bootsteps.py", line 365, in start
    return self.obj.start()
  File "/home/shreyas/.local/share/virtualenvs/inai-backend-mwrd66wz/lib/python3.8/site-packages/celery/worker/consumer/consumer.py", line 326, in start
    blueprint.start(self)
  File "/home/shreyas/.local/share/virtualenvs/inai-backend-mwrd66wz/lib/python3.8/site-packages/celery/bootsteps.py", line 116, in start
    step.start(parent)
  File "/home/shreyas/.local/share/virtualenvs/inai-backend-mwrd66wz/lib/python3.8/site-packages/celery/worker/consumer/consumer.py", line 618, in start
    c.loop(*c.loop_args())
  File "/home/shreyas/.local/share/virtualenvs/inai-backend-mwrd66wz/lib/python3.8/site-packages/celery/worker/loops.py", line 81, in asynloop
    next(loop)
  File "/home/shreyas/.local/share/virtualenvs/inai-backend-mwrd66wz/lib/python3.8/site-packages/kombu/asynchronous/hub.py", line 298, in create_loop
    item()
  File "/home/shreyas/.local/share/virtualenvs/inai-backend-mwrd66wz/lib/python3.8/site-packages/vine/promises.py", line 160, in __call__
    return self.throw()
  File "/home/shreyas/.local/share/virtualenvs/inai-backend-mwrd66wz/lib/python3.8/site-packages/vine/promises.py", line 157, in __call__
    retval = fun(*final_args, **final_kwargs)
[...]

PR #1496, opened 2022-02-25
SQS: Fix message arg in backoff policy

matmarczak:
Hi!

I just found an error in SQS implementation regarding backoff tasks. There was already notion of this [here](https://github.com/celery/kombu/pull/1372) or [here](https://github.com/celery/kombu/issues/1358). I still experience this problem when backoff policy is defined explicitly. [Prepared small example to reproduce issue](https://gitlab.com/amonowy/celery-sqs-reproduce). Fix is in this PR.

I'm open to comments and happy to improve this PR!
Error is here:
[err.txt](https://github.com/celery/kombu/files/8142965/err.txt)

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -215,8 +215,8 @@ class QoS(virtual.QoS):
                 VisibilityTimeout=policy_value
             )
 
-    @staticmethod
-    def extract_task_name_and_number_of_retries(message):
+    def extract_task_name_and_number_of_retries(self, delivery_tag):
+        message = self._delivered[delivery_tag]
         message_headers = message.headers
         task_name = message_headers['task']
         number_of_retries = int(


auvipy (changes requested):
can you fix the lint error please?

kombu/transport/SQS.py
```diff
@@ -200,10 +201,10 @@ def _extract_backoff_policy_configuration_and_message(self, delivery_tag):
         return routing_key, message, backoff_tasks, backoff_policy
 
     def apply_backoff_policy(self, routing_key, delivery_tag,
-                             backoff_policy, backoff_tasks):
+                             backoff_policy, backoff_tasks, message):
         queue_url = self.channel._queue_cache[routing_key]
         task_name, number_of_retries = \
-            self.extract_task_name_and_number_of_retries(delivery_tag)
+            self.extract_task_name_and_number_of_retries(message)
```
thedrow, line 207: Shouldn't we get the message using `message = self._delivered[delivery_tag]`?
This PR breaks the class' public API and I'd rather avoid doing that.

thedrow (changes requested):

thedrow (approved):
This is much better. Thanks!

auvipy (approved):

PR #1506, opened 2022-03-28
Kafka support

CountRedClaw:
Based on #600 but switched from pykafka to https://github.com/confluentinc/confluent-kafka-python as pykafka project was archived. 
Also aiokafka was mentioned [here](https://github.com/celery/kombu/issues/301#issuecomment-772577882), but I didn't find any usages of asyncio libraries in Kombu. Thus I assumed that the best option for me would be confluent-kafka-python.

Integration tests are green, but had to exclude BaseTimeToLive and BasePriority tests because Kafka does not allow to set expiration time per message, as well as it does not provide message priority feature (Kafka partition is an immutable ordered log of messages).

I'm not sure about the relevance of my PR (maybe you meant to use another library), so let me know if it makes sense.


matusvalo:
Can you add also a documentation to head of file - see redis transport as an example:
https://github.com/celery/kombu/blob/aab2588c4ab2f8fd67c6d1c6be3243504b028504/kombu/transport/redis.py#L1-L52

gudipudipradeep (approved):

auvipy (changes requested):
a rebase & docs update needed

CountRedClaw:
Seems like confluent-kafka-python [doesn't support PyPy](https://github.com/confluentinc/confluent-kafka-python/issues/298#issuecomment-357904764). So had to disable PyPy tests for kafka.

auvipy:
> Seems like confluent-kafka-python [doesn't support PyPy](https://github.com/confluentinc/confluent-kafka-python/issues/298#issuecomment-357904764). So had to disable PyPy tests for kafka.

that's OK, we just need to mention it on docs/release notes

kombu/transport/confluentkafka.py
```diff
@@ -0,0 +1,379 @@
...
+        self._kafka_consumers = {}
+
+
+class Transport(virtual.Transport):
+    """Kafka Transport."""
+
+    def as_uri(self, uri: str, include_password=False, mask='**') -> str:
+        pass
+
+    Channel = Channel
```
slenart, line 356: hmmmm

Issue #1511, opened 2022-04-08
Plan to annotate - Kombu


Kludex:
# Fully Type Annotation - Kombu

Analogous to #7394.

|  Status  |                  Filename                  |  PR   |
|  :----:  |  ----------------------------------------  |  ---  |
|   :x:    |  **\_\_init\_\_.py**                       |       |
|   :x:    |  **abstract.py**                           |       |
|   :x:    |  **asynchronous/\_\_init\_\_.py**          |       |
|   :x:    |  **asynchronous/aws/\_\_init\_\_.py**      |       |
|   :x:    |  **asynchronous/aws/connection.py**        |       |
|   :x:    |  **asynchronous/aws/ext.py**               |       |
|   :x:    |  **asynchronous/aws/sqs/\_\_init\_\_.py**  |       |
|   :x:    |  **asynchronous/aws/sqs/connection.py**    |       |
|   :x:    |  **asynchronous/aws/sqs/ext.py**           |       |
|   :x:    |  **asynchronous/aws/sqs/message.py**       |       |
|   :x:    |  **asynchronous/aws/sqs/queue.py**         |       |
|   :x:    |  **asynchronous/debug.py**                 |       |
|   :x:    |  **asynchronous/http/\_\_init\_\_.py**     |       |
|   :x:    |  **asynchronous/http/base.py**             |       |
|   :x:    |  **asynchronous/http/curl.py**             |       |
|   :x:    |  **asynchronous/hub.py**                   |       |
|   :x:    |  **asynchronous/semaphore.py**             |       |
|   :x:    |  **asynchronous/timer.py**                 |       |
|   :x:    |  **clocks.py**                             |       |
|   :x:    |  **common.py**                             |       |
|   :x:    |  **compat.py**                             |       |
|   :x:    |  **compression.py**                        |       |
|   :x:    |  **connection.py**                         |       |
|   :x:    |  **entity.py**                             |       |
|   :x:    |  **exceptions.py**                         |       |
|   :x:    |  **log.py**                                |       |
|   :x:    |  **matcher.py**                            |       |
|   :x:    |  **message.py**                            |       |
|   :x:    |  **messaging.py**                          |       |
|   :x:    |  **mixins.py**                             |       |
|   :x:    |  **pidbox.py**                             |       |
|   :x:    |  **pools.py**                              |       |
|   :x:    |  **resource.py**                           |       |
|   :x:    |  **serialization.py**                      |       |
|   :x:    |  **simple.py**                             |       |
[...]

PR #1513, opened 2022-04-08
chore: Annotate semaphore.py

sondrelg:
Related to #1511

### DIFF
--- a/kombu/asynchronous/semaphore.py
+++ b/kombu/asynchronous/semaphore.py
@@ -1,8 +1,20 @@
 """Semaphores and concurrency primitives."""
-
+import sys
 from collections import deque
+from typing import TYPE_CHECKING, Callable, Deque, Optional, Tuple, Type
+
+if sys.version_info < (3, 10):
+    from typing_extensions import ParamSpec
+else:
+    from typing import ParamSpec
+
+if TYPE_CHECKING:
+    from types import TracebackType
+
 
 __all__ = ('DummyLock', 'LaxBoundedSemaphore')
+
+P = ParamSpec("P")
 
 
 class LaxBoundedSemaphore:
@@ -28,13 +40,18 @@ class LaxBoundedSemaphore:
         HELLO 3
     """
 
-    def __init__(self, value):
+    def __init__(self, value: int) -> None:
         self.initial_value = self.value = value
-        self._waiting = deque()
+        self._waiting: Deque[Tuple] = deque()
         self._add_waiter = self._waiting.append
         self._pop_waiter = self._waiting.popleft
 
-    def acquire(self, callback, *partial_args, **partial_kwargs):
+    def acquire(
+        self,
+        callback: Callable[P, None],
+        *partial_args: P.args,
+        **partial_kwargs: P.kwargs
+    ) -> bool:
         """Acquire semaphore.
 
         This will immediately apply ``callback`` if
@@ -54,7 +71,7 @@ class LaxBoundedSemaphore:
             callback(*partial_args, **partial_kwargs)
             return True
 
-    def release(self):
+    def release(self) -> None:
         """Release semaphore.
 
         Note:
@@ -68,23 +85,24 @@ class LaxBoundedSemaphore:
         else:
             waiter(*args, **kwargs)
 
-    def grow(self, n=1):
+    def grow(self, n: int = 1) -> None:
         """Change the size of the semaphore to accept more users."""
         self.initial_value += n
         self.value += n
-        [self.release() for _ in range(n)]
+        for _ in range(n):
+            self.release()
 
-    def shrink(self, n=1):
+    def shrink(self, n: int = 1) -> None:
         """Change the size of the semaphore to accept less users."""
         self.initial_value = max(self.initial_value - n, 0)
         self.value = max(self.value - n, 0)
 
-    def clear(self):
+    def clear(self) -> None:
         """Reset the semaphore, which also wipes out any waiting callbacks."""
         self._waiting.clear()
         self.value = self.initial_value
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return '<{} at {:#x} value:{} waiting:{}>'.format(
             self.__class__.__name__, id(self), self.value, len(self._waiting),
         )
@@ -93,8 +111,13 @@ class LaxBoundedSemaphore:
 class DummyLock:
     """Pretending to be a lock."""
 
-    def __enter__(self):
+    def __enter__(self) -> 'DummyLock':
         return self
 
-    def __exit__(self, *exc_info):
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional['TracebackType']
+    ) -> None:
         pass
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,7 +21,8 @@ show_error_codes = True
 disallow_untyped_defs = True
 ignore_missing_imports = True
 files =
-    kombu/utils/time.py
+    kombu/utils/time.py,
+    kombu/asynchronous/semaphore.py
 
 [pep257]
 ignore = D102,D104,D203,D105,D213
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,4 +1,5 @@
 importlib-metadata>=0.18; python_version<"3.8"
 cached_property; python_version<"3.8"
+typing_extensions; python_version<"3.10"
 amqp>=5.0.9,<6.0.0
 vine


kombu/asynchronous/semaphore.py
```diff
@@ -93,8 +100,8 @@ def __repr__(self):
 class DummyLock:
     """Pretending to be a lock."""
 
-    def __enter__(self):
+    def __enter__(self) -> 'DummyLock':
         return self
 
-    def __exit__(self, *exc_info):
+    def __exit__(self, *exc_info: Any) -> None:
```
auvipy, line 106: https://sethmlarson.dev/blog/tests-arent-enough-case-study-after-adding-types-to-urllib3#anything-but-any

auvipy (changes requested):

kombu/asynchronous/semaphore.py
Kludex: We can replace exc_info by the three parameters and annotate them properly.

kombu/asynchronous/semaphore.py
```diff
@@ -28,13 +30,18 @@ class LaxBoundedSemaphore:
...
         self._waiting = deque()
         self._add_waiter = self._waiting.append
         self._pop_waiter = self._waiting.popleft
 
-    def acquire(self, callback, *partial_args, **partial_kwargs):
+    def acquire(
+        self,
+        callback: Callable,
+        *partial_args: Any,
+        **partial_kwargs: Any
```
Kludex, line 43:
```suggestion
        callback: Callable[P, None]
        *partial_args: P.args,
        **partial_kwargs: P.kwargs
```
image

kombu/asynchronous/semaphore.py
```diff
@@ -4,6 +4,8 @@
 
 __all__ = ('DummyLock', 'LaxBoundedSemaphore')
 
+from typing import Any, Callable
```
Kludex, line 7: I guess we'd want to move the imports closer to the other import statements?

Kludex:
The file needs to be added on the `setup.cfg`, so mypy can look at it.

sondrelg:
I'll add it shortly 😊

kombu/asynchronous/semaphore.py
sondrelg: I'll replace it with this 👍 
```python
def __exit__(self, exc_type: type, exc_val: Exception, exc_tb: 'TracebackType') -> None:
```

kombu/asynchronous/semaphore.py
sondrelg: I'd never heard of `ParamSpec` before - that's really cool!

kombu/transport/virtual/base.py
```diff
@@ -722,7 +726,8 @@ def _restore(self, message):
         message = message.serializable()
         message['redelivered'] = True
         for queue in self._lookup(
-                delivery_info['exchange'], delivery_info['routing_key']):
+            delivery_info['exchange'],
+                delivery_info['routing_key']):
```
sondrelg, line 730: flake forced me to update indentation after touching the file 🙂

sondrelg:
I went ahead and updated all `__exit__` signatures since they should all be the same. Hope that's ok.

Should be ready for another looks now @Kludex 👍

kombu/asynchronous/semaphore.py
```diff
@@ -1,9 +1,15 @@
 """Semaphores and concurrency primitives."""
 
 from collections import deque
+from typing import TYPE_CHECKING, Callable, ParamSpec
```
Kludex, line 4: ParamSpec was introduced in 3.10, so typing_extensions is needed for the other versions.

sondrelg: Should I then also put this inside the `TYPE_CHECKING` condition, in case the package is run in a runtime that doesn't have `typing_extensions` installed? I see it's not listed in the package dependencies, and it isn't built-in is it?

sondrelg: Added 76e6c7a7607bb3c5b87e6983c5ed0335ae38015b as an initial change

sondrelg:
I see I might've changed one `__exit__` signature too many. I'll fix 👍

sondrelg:
Fixed broken tests by changing the tests themselves. Instead of calling `__exit__` explicitly, I changed to using `with` statements.

kombu/asynchronous/semaphore.py
Kludex: I guess we'd want to include the package as dependency, but only for the versions that is not supported (with a marker). And in code we need to check the version with sys.version_info

kombu/asynchronous/http/base.py
```diff
@@ -253,5 +257,10 @@ def on_header(self, headers, line):
     def __enter__(self):
         return self
 
-    def __exit__(self, *exc_info):
+    def __exit__(
+        self,
+        exc_type: type,
+        exc_val: Exception,
+        exc_tb: 'TracebackType'
```
Kludex, line 264: Ref.: https://github.com/python/typeshed/blob/354787f6e0037f7b8d37d11eb851b8f4e0931a42/stdlib/contextlib.pyi#L187-L189
```suggestion
        exc_type: Optional[Type[BaseException]],
        exc_val: Optional[BaseException],
        exc_tb: Optional['TracebackType']
```

kombu/asynchronous/semaphore.py
```diff
@@ -93,8 +108,13 @@ def __repr__(self):
...
-    def __enter__(self):
+    def __enter__(self) -> 'DummyLock':
         return self
 
-    def __exit__(self, *exc_info):
+    def __exit__(
+        self,
+        exc_type: type,
+        exc_val: Exception,
+        exc_tb: 'TracebackType'
```
Kludex, line 118: Same as previous.

kombu/asynchronous/timer.py
```diff
@@ -101,7 +105,12 @@ def __init__(self, max_interval=None, on_error=None, **kwargs):
     def __enter__(self):
         return self
 
-    def __exit__(self, *exc_info):
+    def __exit__(
+        self,
+        exc_type: type,
+        exc_val: Exception,
+        exc_tb: 'TracebackType'
```
Kludex, line 112: Same as previous.
```suggestion
        exc_type: type,
        exc_val: Exception,
        exc_tb: 'TracebackType'
```

kombu/compat.py
```diff
@@ -65,7 +69,12 @@ def close(self):
     def __enter__(self):
         return self
 
-    def __exit__(self, *exc_info):
+    def __exit__(
+        self,
+        exc_type: type,
+        exc_val: Exception,
+        exc_tb: 'TracebackType'
```
Kludex, line 76: Same as previous.
```suggestion
        exc_type: type,
        exc_val: Exception,
        exc_tb: 'TracebackType'
```

Kludex (commented):
I didn't comment all the `__exit__`, but that's the right annotation according to typeshed.

kombu/asynchronous/semaphore.py
```diff
@@ -28,13 +37,18 @@ class LaxBoundedSemaphore:
         HELLO 3
     """
 
-    def __init__(self, value):
+    def __init__(self, value: int) -> None:
         self.initial_value = self.value = value
-        self._waiting = deque()
+        self._waiting: deque = deque()
```
Kludex, line 42: `Deque[<Type>]`?

sondrelg: Yeah you're right. Think the correct type can use the ParamSpec actually:

```python
Deque[Tuple[P, P.args, P.kwargs]]
```

sondrelg:
Cleaned up the `__exit__` and deque annotations, thanks! Let me know if there's anything else 👍

control/test_transport_filesystem.exchange
```diff
@@ -0,0 +1 @@
+[["test_transport_filesystem2", "", "test_transport_filesystem2"], ["test_transport_filesystem", "", "test_transport_filesystem"]]
```
Kludex, line 1: Too much?

kombu/asynchronous/semaphore.py
```diff
@@ -28,13 +37,18 @@ class LaxBoundedSemaphore:
         HELLO 3
     """
 
-    def __init__(self, value):
+    def __init__(self, value: int) -> None:
         self.initial_value = self.value = value
-        self._waiting = deque()
+        self._waiting: Deque[Tuple[P, P.args, P.kwargs]] = deque()
```
Kludex, line 42: I don't think this is a possible use case, because the second argument is actually a list, and the third is actually a dictionary, also ParamSpec is only used with a callable. This passes on mypy? In any case, it should be at least:
```suggestion
        self._waiting: Deque[Tuple[Callable[P, None], P.args, P.kwargs]] = deque()
```

Kludex (commented):
Besides what I've commented, everything else LGTM.

kombu/asynchronous/semaphore.py
sondrelg: Mypy did not like either, so I reverted to `Deque[tuple]` for now.

sondrelg:
Rebase to make this PR a little more concise. Will open a separate PR for the remaining `__exit__` signature changes.

setup.cfg
```diff
@@ -22,6 +22,7 @@ disallow_untyped_defs = True
 ignore_missing_imports = True
 files =
     kombu/utils/time.py
```
Kludex, line 24: I guess the comma is needed 🤔 
```suggestion
    kombu/utils/time.py,
```

Kludex (approved):
Besides the comma, everything looks great to me 👍

kombu/asynchronous/semaphore.py
```diff
@@ -28,13 +40,18 @@ class LaxBoundedSemaphore:
         HELLO 3
     """
 
-    def __init__(self, value):
+    def __init__(self, value: int) -> None:
         self.initial_value = self.value = value
-        self._waiting = deque()
+        self._waiting: Deque[tuple] = deque()
```
Kludex, line 45: Maybe? 🤔
```suggestion
        self._waiting: Deque[Tuple[callable, list, dict] = deque()
```

sondrelg: This unfortunately causes issues with the ParamSpec annotations further down:

```
kombu/asynchronous/semaphore.py:67: error: Argument 1 has incompatible type "Tuple[Callable[P, None], P.args, P.kwargs]"; expected "Tuple[Callable[..., Any], List[Any], Dict[Any, Any]]"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

auvipy:
> Besides the comma, everything looks great to me +1
> 
> Not that my approval counts, but after that many comments, is fair to at least show it hahaha

have another look before i merge this please

Kludex (approved):
LGTM 👍 

@auvipy 👍

auvipy:
> Related to #1511

congrats on first contributions

PR #1514, opened 2022-04-09
Annotate `utils/uuid.py` and `utils/text.py`

Kludex:
Related to #1511

### DIFF
--- a/kombu/utils/text.py
+++ b/kombu/utils/text.py
@@ -2,7 +2,6 @@
 # flake8: noqa
 
 
-from ctypes import cast
 from difflib import SequenceMatcher
 from typing import Iterator, Optional, Sequence, Tuple, Union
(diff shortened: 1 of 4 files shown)


kombu/utils/text.py
```diff
@@ -2,7 +2,9 @@
 # flake8: noqa
 
 
+from ctypes import cast
```
thedrow, line 5: Do we need to import this?

Kludex: No, it's wrong.

Kludex, line 5:
```suggestion
```

Kludex: Thanks 🙏

auvipy (changes requested):
needs a rebase

PR #1516, opened 2022-04-11
Added possibility to serialize and deserialize binary messages in json

dobosevych:
Signature args in callback_msg should be serializable, so
```
callback_msg = str(uuid.uuid4()).encode()
redis_key = str(uuid.uuid4())
callback = redis_echo.si(callback_msg, redis_key=redis_key)
```
,so should be
as they will be passed inside as_task_v2 that would return message that would be passed to a send_task_message, then to publish (kombu) and _prepare (kombu) in messaging that would cll a serialization (kombu).
```
from kombu.serialization import dumps
s = ((42,), {}, {'callbacks': [{'task': 't.integration.tasks.redis_echo', 'args': (b'7a108737-0b0a-4acc-a32f-f8a5e6726775',), 'kwargs': {'redis_key': 'ef3bc79f-a04c-4de1-88c0-f13d9d999ca8'}, 'options': {}, 'subtask_type': None, 'immutable': True}], 'errbacks': None, 'chain': None, 'chord': None})
print(dumps(s))
```
to reproduce serialization issue

### DIFF
--- a/kombu/utils/json.py
+++ b/kombu/utils/json.py
@@ -4,6 +4,7 @@ import datetime
 import decimal
 import json as stdjson
 import uuid
+import base64
 
 try:
     from django.utils.functional import Promise as DjangoPromise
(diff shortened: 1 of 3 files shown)


kombu/utils/json.py
```diff
@@ -55,6 +55,8 @@ def default(self, o,
                 return o.isoformat()
             elif isinstance(o, textual):
                 return text_t(o)
+            elif isinstance(o, bytes):
+                return {"bytes": o.decode("utf-8"), "__bytes__": True}
```
dobosevych, line 59: try: 
except UnicodeDecoder:
 raise DecodeError

auvipy (approved):

PR #1519, opened 2022-04-11
Annotate `utils/url.py`

nils-van-zuijlen:
Related to https://github.com/celery/kombu/issues/1511

### DIFF
--- a/kombu/utils/url.py
+++ b/kombu/utils/url.py
@@ -4,7 +4,7 @@
 
 from collections.abc import Mapping
 from functools import partial
-from typing import NamedTuple
+from typing import Any, Dict, NamedTuple, Optional, TypeVar, Union
 from urllib.parse import parse_qsl, quote, unquote, urlparse
 
 try:
@@ -30,8 +30,7 @@ class urlparts(NamedTuple):
     query: Mapping
 
 
-def parse_url(url):
-    # type: (str) -> Dict
+def parse_url(url: str) -> Dict[str, Any]:
     """Parse URL into mapping of components."""
     scheme, host, port, user, password, path, query = _parse_url(url)
     if query:
@@ -53,8 +52,7 @@ def parse_url(url):
                 password=password, virtual_host=path, **query)
 
 
-def url_to_parts(url):
-    # type: (str) -> urlparts
+def url_to_parts(url: str) -> urlparts:
     """Parse URL into :class:`urlparts` tuple of components."""
     scheme = urlparse(url).scheme
     schemeless = url[len(scheme) + 3:]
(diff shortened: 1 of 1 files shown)


kombu/utils/url.py
```diff
@@ -97,22 +96,19 @@ def as_url(scheme, host=None, port=None, user=None, password=None,
...
-    # type: (str, str) -> str
+def sanitize_url(url: str, mask: str='**') -> str:
     """Return copy of URL with password removed."""
     return as_url(*_parse_url(url), sanitize=True, mask=mask)
 
-
-def maybe_sanitize_url(url, mask='**'):
-    # type: (Any, str) -> Any
+_T = TypeVar("_T")
+def maybe_sanitize_url(url: _T, mask: str='**') -> Union[_T, str]:
```
Kludex, line 104:
```suggestion
_T = TypeVar("_T")

@overload
def maybe_sanitize_url(url: str, mask: str = ...) -> str:
    ...
    
@overload
def maybe_sanitize_url(url: _T, mask: str = ...) -> _T:
    ...

def maybe_sanitize_url(url, mask='**'):
```

kombu/utils/url.py
```diff
@@ -4,7 +4,7 @@
 
 from collections.abc import Mapping
 from functools import partial
-from typing import NamedTuple
+from typing import Any, Dict, NamedTuple, Optional, TypeVar, Union
```
Kludex, line 7:
```suggestion
from typing import Any, Dict, NamedTuple, Optional, TypeVar
```

kombu/utils/url.py
```diff
@@ -30,8 +30,7 @@ class urlparts(NamedTuple):
     query: Mapping
 
 
-def parse_url(url):
-    # type: (str) -> Dict
+def parse_url(url: str) -> Dict[str, Any]:
```
Kludex, line 33: Unfortunately, `TypedDict` doesn't support extra keys, so this is the best we can get... More information on: https://github.com/python/mypy/issues/4617

kombu/utils/url.py
nils-van-zuijlen: Actually, the `str -> str` case is already covered by `_T -> _T`, so it it is not required. Should I specify it anyway ?

Kludex: Oh! Then no. 😅

kombu/utils/url.py
auvipy: anything but Any

nils-van-zuijlen: I'm still unsure about `Dict[str, Union[str, int, Dict[str, str], 'ssl.VerifyMode']]`. If someone could double-check that these are the right types in that union it would be great.

Kludex: On my side, I think `Dict[str, Any]` is the right call.

Another possibility would be (which I'm not sure...):
```python
class ParsedURL(TypedDict):
    transport: str
    hostname: str
    port: int
    userid: str
    password: str
    virtual_host: str

def parse_url(url: str) -> Union[ParsedURL, Dict[str, Any]]
```

Kludex: Other than that, you can check if you can create the exact type with some information provided on this issue: https://github.com/python/mypy/issues/4617

PR #1522, opened 2022-04-11
Annotate `abstract.py`

Kludex:
Related to #1511


Kludex:
Why is there a bot that commits on my branch..........

auvipy (changes requested):
please rebase & try to reduce the usage of Any types https://sethmlarson.dev/blog/tests-arent-enough-case-study-after-adding-types-to-urllib3#anything-but-any

kombu/abstract.py
```diff
@@ -1,19 +1,33 @@
...
+_T = TypeVar("_T")
+_ObjectType = TypeVar("_ObjectType", bound="Object")
+_MaybeChannelBoundType = TypeVar(
+    "_MaybeChannelBoundType", bound="MaybeChannelBound"
+)
+
 
-def unpickle_dict(cls, kwargs):
+def unpickle_dict(
+    cls: Type[_ObjectType], kwargs: Dict[str, Any]
```
Kludex, line 25: This `Any` is alright.

kombu/abstract.py
```diff
@@ -23,9 +37,9 @@ class Object:
     Supports automatic kwargs->attributes handling, and cloning.
     """
 
-    attrs = ()
+    attrs: Tuple[Tuple[str, Any], ...] = ()
```
Kludex, line 40: The thing here is that each `attr` can really be `Any`, and `str` is the name of the `attr`. It looks fine as well.

kombu/abstract.py
```diff
@@ -36,48 +50,57 @@ def __init__(self, *args, **kwargs):
                 except AttributeError:
                     setattr(self, name, None)
 
-    def as_dict(self, recurse=False):
-        def f(obj, type):
+    def as_dict(self, recurse: bool = False) -> Dict[str, Any]:
+        def f(obj: Any, type: Callable[[Any], Any]) -> Any:
```
Kludex, line 54: Here is the only place that we can try to improve... I'm not sure what can be a good replacement tho...

Kludex:
rebased 👍

auvipy (approved):

PR #1524, opened 2022-04-12
Update `__exit__` function signatures

sondrelg:
Follow up from #1513

Relates to #1511

PR updates all `__exit__` function signatures, and adjusts (now) failing tests that were calling `__exit__` without any arguments (which I seems incorrect)

### DIFF
--- a/kombu/asynchronous/http/base.py
+++ b/kombu/asynchronous/http/base.py
@@ -2,6 +2,7 @@
 
 import sys
 from http.client import responses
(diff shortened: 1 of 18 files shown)


auvipy (commented):
we are changing the signature from *exc_info/*args to more explicit ones, but is it breaking the API?

Kludex:
It should be equivalent, considering `typeshed`: https://github.com/python/typeshed/blob/master/stdlib/contextlib.pyi

PR #1525, opened 2022-04-12
Annotate and refactor limits.py

sondrelg:
Relates to #1511 

Adds annotations to `utils/limits.py`, but notably also removes and refactors some code. 

Would understand if the refactoring isn't wanted, but AFAICT the file is not a part of the public API, and the code is completely unused. Annotating this code is also made a little harder by this fact. 

This seems to be the commit where the `add`, `pop`, and `clear` methods were added - but they're not implemented here either: https://github.com/celery/kombu/commit/946c59c55953375ab72cd821335e21adc3293109


kombu/utils/limits.py
```diff
@@ -21,32 +21,21 @@ class TokenBucket:
...
+        capacity: int = 1
+    ) -> None:
+        #: Maximum number of tokens in the bucket.
+        self.capacity = self._tokens = float(capacity)
 
-    #: Maximum number of tokens in the bucket.
-    capacity = 1
-
-    #: Timestamp of the last time a token was taken out of the bucket.
-    timestamp = None
```
sondrelg, line 31: These three class variables would always be overwritten in the `__init__`, so they seem redundant to me. I could be wrong, but there are no classmethods or staticmethods or anything else to suggest this is code we want to keep.

kombu/utils/limits.py
```diff
@@ -21,32 +21,21 @@ class TokenBucket:
...
 
-    #: The rate in tokens/second that the bucket will be refilled.
-    fill_rate = None
+    def __init__(
+        self,
+        fill_rate: Union[str, int, float],
+        capacity: int = 1
+    ) -> None:
+        #: Maximum number of tokens in the bucket.
+        self.capacity = self._tokens = float(capacity)
```
sondrelg, line 30: token capacity is discrete and logically is an int, but cast it to float here to make `min` and `max` calls simpler further down on the class.

auvipy (changes requested):
please rebase

auvipy:
@Kludex can you have a look

kombu/utils/limits.py
```diff
@@ -74,10 +62,10 @@ def expected_time(self, tokens=1):
...
 
-    def _get_tokens(self):
+    def _get_tokens(self) -> int:
         if self._tokens < self.capacity:
             now = monotonic()
             delta = self.fill_rate * (now - self.timestamp)
             self._tokens = min(self.capacity, self._tokens + delta)
             self.timestamp = now
-        return self._tokens
+        return int(self._tokens)
```
Kludex, line 71: From where are we inferring it should be `int`?

Kludex (commented):
> Specifically this removes the add, pop, and clear_pending methods on the TokenBucket class. This seems to be the commit where they were added - but they're not implemented here either: https://github.com/celery/kombu/commit/946c59c55953375ab72cd821335e21adc3293109. No tests were broken by the removal and I can only see one place where the class is accessed. None of the downstream uses from that one property call any of these methods.

I'll let the part of the code that changes the above to you @auvipy.

About the rest of the code, I've added a comment. As I'm a little confused about the `self._tokens` cast.

kombu/utils/limits.py
sondrelg: This looks like an oversight on my part. Both the `tokens` args in the `expected_time` method and the return type of `_get_tokens` look like they should be floats.

Updated in 7589dc54

kombu/utils/limits.py
```diff
@@ -23,32 +22,21 @@ class TokenBucket:
         section of any multithreaded code.
     """
 
-    #: The rate in tokens/second that the bucket will be refilled.
-    fill_rate = None
+    def __init__(
+        self,
+        fill_rate: str | int | float,
+        capacity: int = 1
```
Kludex, line 28: I guess `capacity` follows the same type as `fill_rate`?

sondrelg: My understanding is that capacity is a discrete measure, and should be an int. At the same time, it's used like this in the class (unchanged code):

```python
self._tokens = min(self.capacity, self._tokens + delta)
```

Since mixing ints and floats causes mypy errors, I decided to cast `self.capacity` and `self._tokens` to floats in the init instead of ignoring the mypy error. Don't think they are used in a manner that require them to be ints anywhere else.

Does that make sense?

PR #1526, opened 2022-04-12
Annotate `exceptions.py` and `clocks.py`

Kludex:
Related to #1511


setup.cfg
```diff
@@ -15,7 +15,6 @@ extend-ignore = W504, N806, N802, N801, N803
 [mypy]
 warn_unused_configs = True
 strict = False
-warn_return_any = True
```
Kludex, line 18: This rule is problematic on operators, like on `__gt__`, you cannot assume that `<` will return a boolean.

I've also disabled on celery for the same reason some days ago.

auvipy (changes requested):
rebase please

auvipy (approved):

PR #1527, opened 2022-04-13
Annotate init files

Kludex:
Related to #1511

### DIFF
--- a/kombu/transport/__init__.py
+++ b/kombu/transport/__init__.py
@@ -1,10 +1,12 @@
 """Built-in transports."""
 
+from typing import Optional
+
 from kombu.utils.compat import _detect_environment
 from kombu.utils.imports import symbol_by_name
 
 
-def supports_librabbitmq():
+def supports_librabbitmq() -> Optional[bool]:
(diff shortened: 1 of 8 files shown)


kombu/transport/__init__.py
```diff
@@ -58,7 +61,7 @@ def resolve_transport(transport=None):
         except KeyError:
             if '.' not in transport and ':' not in transport:
                 from kombu.utils.text import fmatch_best
-                alt = fmatch_best(transport, TRANSPORT_ALIASES)
+                alt = fmatch_best(transport, list(TRANSPORT_ALIASES.keys()))
```
auvipy, line 64: why  list(TRANSPORT_ALIASES.keys()) ?

Kludex: It was my mistake!

Kludex:
The `fmatch_best` was actually wrong annotated. Thanks for the review!

auvipy (approved):

Kludex:
Thanks @auvipy ! 🙇

Issue #1528, opened 2022-04-13
JSON utils looks redundant


dobosevych:
At top of https://github.com/celery/kombu/blob/master/kombu/utils/json.py. 
We import `import json as stdjson`, but then we try again to `import json` and otherwise do `import simplejson`. This looks like legacy of Python 2.7 and there is no need to support that. I propose to clean top of file and keep only `import json` and remove this `stdjson`.

auvipy:
yes, just using json is alright. simple json is old.

PR #1533, opened 2022-04-14
json.py cleaning from outdated libs

dobosevych:
We import `import json as stdjson`, but then we try again to `import json` and otherwise do `import simplejson`. This is the legacy of Python 2.7 and there is no need to support that.
Issue: #1528

### DIFF
--- a/kombu/utils/json.py
+++ b/kombu/utils/json.py
@@ -13,9 +13,6 @@ except ImportError:  # pragma: no cover
         """Dummy object."""
 
 
-_json_extra_kwargs = {}
-
-
 class _DecodeError(Exception):
     pass
 
@@ -65,8 +62,7 @@ _default_encoder = JSONEncoder
 
 def dumps(s, _dumps=json.dumps, cls=None, default_kwargs=None, **kwargs):
     """Serialize object to json string."""
-    if not default_kwargs:
-        default_kwargs = _json_extra_kwargs
+    default_kwargs = default_kwargs or {}
     return _dumps(s, cls=cls or _default_encoder,
                   **dict(default_kwargs, **kwargs))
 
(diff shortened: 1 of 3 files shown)


clokep:
Note that some of the docs refer to this still, I'm unsure if those references should be cleaned up?

dobosevych:
> Note that some of the docs refer to this still, I'm unsure if those references should be cleaned up?

Can you point me to these docs?

clokep:
`docs/userguide/serialization.rst` and `docs/reference/kombu.serialization.rst` have references to `simplejson` still. I suspect the one in `kombu.serialization.rst` is a link target that's completely unused now. The one in `serialization.rst` seems just unnecessary. 🤷

kombu/utils/json.py
```diff
@@ -12,19 +12,12 @@
...
-    class _DecodeError(Exception):
-        pass
-except ImportError:                 # pragma: no cover
-    import simplejson as json
-    from simplejson.decoder import JSONDecodeError as _DecodeError
-    _json_extra_kwargs = {
-        'use_decimal': False,
-        'namedtuple_as_object': False,
-    }
+_json_extra_kwargs = {}
```
MelnykR, line 16: This variable seems unnecessary. Since it is private -- maybe you can remove it as well ?

dobosevych: Removed and usage replaced with or

dobosevych:
> `docs/userguide/serialization.rst` and `docs/reference/kombu.serialization.rst` have references to `simplejson` still. I suspect the one in `kombu.serialization.rst` is a link target that's completely unused now. The one in `serialization.rst` seems just unnecessary. 🤷

I removed `simplejson` and `cjson` from documentation, as both libs aren't supported anymore.

auvipy (approved):

PR #1536, opened 2022-04-17
Support pymongo 4.x

Jakski:
I recently tried to use `kombu` with 4.x `pymongo` driver, but it's not possible. New major version broke compatibility by introducing [a lot of changes](https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html). This patch adds support for `pymongo` 4.x and integration tests coverage. I also tried to run patch with `pymongo` 3.x and it seems to work as well. Most of breaking changes in 4.x are about removing previously deprecated high level helper functions.

I tried to use `BaseTimeToLive` test suite as well, but it's not compatible with MongoDB TTL indices. Expired documents are removed [approximately once every 60 seconds](https://www.mongodb.com/docs/manual/core/index-ttl/). To test it reliably we would need to use undocumented [ttlMonitorSleepSecs parameter](https://jira.mongodb.org/browse/SERVER-30388) or fix expiration on query level, e.g.: queries would exclude expired entries based on `expire_at` field and tailable cursors would skip them.


auvipy:
thank you, I thin we have to make change in celery as well

auvipy:
related https://github.com/celery/celery/pull/7469

auvipy (approved):

PR #1539, opened 2022-04-22
Upgrade Azure Storage Queues transport to version 12

jonasmiederer:
Fixes https://github.com/celery/kombu/issues/1468

Fixed Azure Storage Queue transport to work with the latest version 12 of the [azure-storage-queue](https://pypi.org/project/azure-storage-queue/) library, which is incompatible with the current implementation.

Since the `QueueServiceClient` now requires a mandatory `account_url` for connection, this may be a breaking change for anyone using the Azure Storage Queue transport  as the connection string has to be updated.


auvipy (approved):

thedrow:
Couldn't we maintain backwards compatibility?

PR #1540, opened 2022-04-25
Added HLEN to the list of prefixed redis commands

vinayinvicible:
Although HLEN is not used anywhere inside kombu or celery, it is the method that is used to get the count of unacked messages.
This can be used similar to monitoring celery queue length.

The unacked_key that is used to store info about the unacknowledged tasks is a hash.
If I wish to identify whether all celery tasks have been processed, I need to ensure that the unacked_key is empty along with queues length also.
When using a prefix, HLEN should also use the same key prefix.

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -191,6 +191,7 @@ class GlobalKeyPrefixMixin:
     PREFIXED_SIMPLE_COMMANDS = [
         "HDEL",
         "HGET",
+        "HLEN",
         "HSET",
         "LLEN",
         "LPUSH",


thedrow (changes requested):
HLEN can return the number of fields in a hash. Are you confusing it with another command?

vinayinvicible:
@thedrow 
No. I mean `HLEN` only.
The `unacked_key` that is used to store info about the unacknowledged tasks is a `hash`.
If I wish to identify whether all celery tasks have been processed, I need to ensure that the unacked_key is empty along with queues length also.

thedrow (approved):

PR #1549, opened 2022-05-09
Fix mongodb connection parameters not passed correctly

zdyxry:
In Kombu 3.x versions, the MongoDB transport allows accepting the Celery `transport_options` parameters(BROKER_TRANSPORT_OPTIONS) as connection parameters.

In https://github.com/celery/kombu/pull/537 , removed support for parameter passing, maybe we should add it back.

### DIFF
--- a/kombu/transport/mongodb.py
+++ b/kombu/transport/mongodb.py
@@ -313,6 +313,7 @@ class Channel(virtual.Channel):
                                  if self.connect_timeout else None),
         }
         options.update(parsed['options'])
+        options.update(client.transport_options)
         options = self._prepare_client_options(options)
 
         return hostname, dbname, options


kombu/transport/mongodb.py
```diff
@@ -313,6 +313,7 @@ def _parse_uri(self, scheme='mongodb://'):
                                  if self.connect_timeout else None),
         }
         options.update(parsed['options'])
+        options.update(client.transport_options)
```
auvipy, line 316: can you please add test for the proposed change?

auvipy (changes requested):

kombu/transport/mongodb.py
zdyxry: OK, i will add test later.

auvipy:
also integration tests are failing

kombu/transport/mongodb.py
auvipy: can you address the tests please?

auvipy (changes requested):

thedrow (changes requested):
Please add test coverage to the proposed fix

zdyxry:
@auvipy  Hi, Are there any documents about unit tests I can refer to?

auvipy:
> @auvipy Hi, Are there any documents about unit tests I can refer to?

you can found existing tests here https://github.com/celery/kombu/blob/main/t/unit/transport/test_mongodb.py

PR #1553, opened 2022-05-30
Annotate utils amp manager

khanh96le:
Related to #1511

### DIFF
--- a/kombu/utils/amq_manager.py
+++ b/kombu/utils/amq_manager.py
@@ -1,22 +1,28 @@
 """AMQP Management API utilities."""
 
-
 from __future__ import annotations
 
+from typing import Optional, Any
 
-def get_manager(client, hostname=None, port=None, userid=None,
-                password=None):
+from kombu import Connection
+
+DEFAULT_RABBIT_MQ_PORT = 15672
+
+
+def get_manager(client: Connection, hostname: Optional[str] = None,
+                port: Optional[int] = None, userid: Optional[str] = None,
+                password: Optional[str] = None):
     """Get pyrabbit manager."""
     import pyrabbit
     opt = client.transport_options.get
 
-    def get(name, val, default):
+    def get(name: str, val: Any, default: Any):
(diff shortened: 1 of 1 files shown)


kombu/utils/amq_manager.py
```diff
@@ -1,22 +1,34 @@
...
+
 
-def get_manager(client, hostname=None, port=None, userid=None,
-                password=None):
+def get_manager(
+    client: Connection,
+    hostname: str | None = None,
+    port: int | None = None,
+    userid: str | None = None,
+    password: str | None = None
```
matusvalo, line 17: Based on PEP 604 Unions in annotations are supported from python version 3.10. Kombu currently is supporting also older versions of python se we need to use here `Optional`

kombu/utils/amq_manager.py
```diff
@@ -1,22 +1,34 @@
...
-    def get(name, val, default):
-        return (val if val is not None
-                else opt('manager_%s' % name) or
-                getattr(client, name, None) or default)
+    def get(name: str, val: Any, default: Any) -> Any:
+        return (
+            val if val is not None
+            else opt(
+                'manager_%s' % name) or getattr(client, name, None) or default
+        )
```
matusvalo, line 28: This PR should keep the diff as small as possible. Please remove this reformatting.

matusvalo (changes requested):

kombu/utils/amq_manager.py
Kludex: As we are using `from __future__ import annotations`, we don't. As annotations are not evaluated at runtime.

Kludex (commented):
This PR also needs to include the name of the file in the mypy file list, I guess.

kombu/utils/amq_manager.py
auvipy: this is OK. we can use python 3.10 stuffs in typing

Issue #1074, opened 2019-08-02
Support for SQS DelaySeconds


ikonst:
SQS has a built-in "DelaySeconds" feature that could support Celery's ETA/Countdown feature for up to 15 minutes.

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-send-message-with-timer.html

auvipy:
could you send a patch for this?

ikonst:
It's on my radar once I have some free time. Just wanted to record an issue first to gather feedback.

edmundlam:
Hello @auvipy, my colleague @erikaloranger recently monkey-patched our local kombu library to add the `DelaySeconds` to the kombu SQS transport, similar to how the `MessageDeduplicationID` and `MessageGroupId` is handled for fifo queues.

Can we fork and open a PR with a patch for review?

auvipy:
please open and ask me for review

PR #1567, opened 2022-07-06
Add support to SQS DelaySeconds

erikaloranger:
SQS has a built-in "DelaySeconds" feature currently not supported by kombu. If `DelaySeconds` is present in the properties, then add it to the kwargs passed to the method `send_message`. Also added a test for DelaySeconds.

Related to celery/kombu#1074

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -413,6 +413,9 @@ class Channel(virtual.Channel):
                     message['properties']['MessageDeduplicationId']
             else:
                 kwargs['MessageDeduplicationId'] = str(uuid.uuid4())
+        else:
+            if "DelaySeconds" in message['properties']:
+                kwargs['DelaySeconds'] = message['properties']['DelaySeconds']
 
         c = self.sqs(queue=self.canonical_queue_name(queue))
         if message.get('redelivered'):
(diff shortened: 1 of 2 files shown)


auvipy:
yes

kombu/transport/SQS.py
```diff
@@ -402,18 +402,23 @@ def _put(self, queue, message, **kwargs):
...
-                    message['properties']['MessageGroupId']
-            else:
-                kwargs['MessageGroupId'] = 'default'
-            if 'MessageDeduplicationId' in message['properties']:
-                kwargs['MessageDeduplicationId'] = \
-                    message['properties']['MessageDeduplicationId']
-            else:
-                kwargs['MessageDeduplicationId'] = str(uuid.uuid4())
 
+        if 'properties' in message:
```
edmundlam, line 406: We added this check because one of the tests checks whether a raw message not created by kombu (hence not having a `properties` value) will still get passed properly to SQS. 

```
        # In test_messages_to_python()
        # json formatted message NOT created by kombu
        for i in range(json_message_count):
            message = {'foo': 'bar'}
            self.channel._put(self.producer.routing_key, message)
```
ref: https://github.com/celery/kombu/blob/master/t/unit/transport/test_SQS.py#L359

Added the check to include the checks for the FIFO params, because a raw message like this sent to a fifo queue would have errored out as well when checking for `properties`.

erikaloranger:
@auvipy We've fixed the test and made sure the tests are passing locally, can you restart the workflow please?

erikaloranger:
@auvipy Could you review this pr? Thanks!

auvipy (approved):

PR #1574, opened 2022-07-14
chore: handle kafka transport with confluentkafka ✨

meysam81:
Based on my local & confluent tests, the same transport class can be used for both `confluentkafka` and `kafka` itself. I've done countless tests and it works.

Based on the works on #1506


kombu/transport/__init__.py
```diff
@@ -24,6 +24,7 @@ def supports_librabbitmq() -> bool | None:
     'pyamqp': 'kombu.transport.pyamqp:Transport',
     'librabbitmq': 'kombu.transport.librabbitmq:Transport',
     'confluentkafka': 'kombu.transport.confluentkafka:Transport',
+    'kafka': 'kombu.transport.confluentkafka:Transport',
```
auvipy, line 27: we might add another back end in near future, like aiokafka, so i am not in favor of adding this change right now :)

auvipy (changes requested):

kombu/transport/__init__.py
meysam81: Well @auvipy , for the moment, you would want to give some API to the already-implemented logic. If in the future, you plan to change the logic, you'd just change one line here. No hard feelings :)

meysam81: Because you see, at the moment, I'm giving `confluentkafka://kafka.kafka.svc.cluster.local` to my app. which is ugly when you think about it.

auvipy (approved):

PR #1575, opened 2022-07-14
Avoid losing type of UUID when serializing/deserializing

el-chogo:
Serializing UUIDs as strs and deserializing them as strs can lead to
somewhat obscure bugs such as the one that led to the creation of this
PR in django-cacheback
https://github.com/codeinthehole/django-cacheback/pull/100 which some
would call unexpected behaviour.

After all, an UUID is altogether a different type from strs and if
bytes got their own serializer/deserializer for UUID the same logic
should apply.

I added a case for every uuidX constructor found in CPython.

Lemme know what you think of it and if there's anything I'm misreading :D.

P.D: I understand that UUIDs are now serialized as a convenience for users
but perhaps it's a bit non-intuitive that after versions of not supporting them
now they're converted to a different type when deserializing?

### DIFF
--- a/kombu/utils/json.py
+++ b/kombu/utils/json.py
@@ -29,7 +29,7 @@ class JSONEncoder(_encoder_cls):
     def default(self, o,
                 dates=(datetime.datetime, datetime.date),
                 times=(datetime.time,),
-                textual=(decimal.Decimal, uuid.UUID, DjangoPromise),
+                textual=(decimal.Decimal, DjangoPromise),
                 isinstance=isinstance,
                 datetime=datetime.datetime,
                 text_t=str):
@@ -44,6 +44,8 @@ class JSONEncoder(_encoder_cls):
                 return {"datetime": r, "__datetime__": True}
             elif isinstance(o, times):
                 return o.isoformat()
+            elif isinstance(o, uuid.UUID):
+                return {"uuid": str(o), "__uuid__": True, "version": o.version}
             elif isinstance(o, textual):
                 return text_t(o)
             elif isinstance(o, bytes):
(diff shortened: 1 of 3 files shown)


kombu/utils/json.py
```diff
@@ -29,7 +29,7 @@ class JSONEncoder(_encoder_cls):
     def default(self, o,
                 dates=(datetime.datetime, datetime.date),
                 times=(datetime.time,),
-                textual=(decimal.Decimal, uuid.UUID, DjangoPromise),
+                textual=(decimal.Decimal, DjangoPromise),
```
auvipy, line 32: should we document this change?

auvipy (changes requested):

kombu/utils/json.py
el-chogo: Hmmmm, there's an entry for `- JSON serializer now handles datetimes, Django promise, UUID and Decimal.` in 4.0. It doesn't specify how it's handled but maybe (if accepted) this change could be documented as part of the next release?

el-chogo: I pushed [6b89388](https://github.com/celery/kombu/pull/1575/commits/6b89388870271e23c31f6c2845f248c832fb7c6f) to keep the userguide (the only place I could find where JSON serialization is discussed) updated c:

thedrow (approved):

auvipy (approved):

youtux:
This PR is quite dangerous as if you have celery workers running 5.2.4, and you are deploying a new codebase using celery 5.3.1, the old workers cannot decode the arguments for the new tasks if there was a `datetime` object!
This should have at least been mentioned in the backwards incompatible changes.

youtux:
cc @auvipy

youtux:
Sorry, this is not the actual PR that introduced the change. I'll try to find the original one and post the warning there.

PR #1584, opened 2022-08-18
Not actually calling the intended function

seto2103:
Causes an error where broker transport options won't actually change the visibility timeout.

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -363,7 +363,7 @@ class Channel(virtual.Channel):
                     "defined in 'predefined_queues'."
                 ).format(queue))
 
-            attributes = {'VisibilityTimeout': str(self.visibility_timeout)}
+            attributes = {'VisibilityTimeout': str(self.visibility_timeout())}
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
can you please change the relevant tests as well?

auvipy:
>           attributes = {'VisibilityTimeout': str(self.visibility_timeout())}
E           TypeError: 'int' object is not callable

kombu/transport/SQS.py:366: TypeError

kombu/transport/SQS.py
```diff
@@ -363,7 +363,7 @@ def _new_queue(self, queue, **kwargs):
                     "defined in 'predefined_queues'."
                 ).format(queue))
 
-            attributes = {'VisibilityTimeout': str(self.visibility_timeout)}
+            attributes = {'VisibilityTimeout': str(self.visibility_timeout())}
```
auvipy, line 366:
```suggestion
            attributes = {'VisibilityTimeout': self.visibility_timeout()}
```

PR #1587, opened 2022-08-21
hub: tick delay fix

shahar-lev:
todo and timer callbacks can perform actions that
require a tick callback to be executed right away
without polling.

the current order can cause issues
when using single worker with no prefetch (acks late).

related issue in celery:
https://github.com/celery/celery/issues/7718


auvipy:
can you add some test cases?

shahar-lev:
added test case

auvipy (approved):

PR #1595, opened 2022-09-09
Revert "Solve Kombu filesystem transport not thread safe"

karajan1001:
Reverts celery/kombu#1593 
Sorry, the PR created by me works even worse than previous on our program. It only solved part of the problem and introduce a new one. I'll create a new PR later. 

The previous problem comes from: 
1. in `_queue_bind`
```python
    def _queue_bind(self, exchange, routing_key, pattern, queue):
        queues = self.get_table(exchange)
        queue_val = exchange_queue_t(routing_key or "", pattern or "",
                                     queue or "")
        if queue_val not in queues:
            queues.insert(0, queue_val)
        with self._get_exchange_file_obj(exchange, "wb") as f_obj:
            f_obj.write(str_to_bytes(dumps(queues)))
```
We first read the table and then add a new line to it. We do add a lock during both progress but released it too quickly and the file might change between these two progress. 
2. in `_get` and `_purge` we modified the data but didn't add a lock to them.

3. The new problem introduced by this PR. 
`signal` only works in the main thread. 

Actually, I didn't expect it to be merged so quickly. Next time I'll add draft sign to any PR that hadn'd been carefully tested. 

I'll create a new PR next Monday.

### DIFF
--- a/kombu/transport/filesystem.py
+++ b/kombu/transport/filesystem.py
@@ -93,7 +93,6 @@ from __future__ import annotations
 
 import os
 import shutil
-import signal
 import tempfile
 import uuid
 from collections import namedtuple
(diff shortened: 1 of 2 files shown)


auvipy (approved):

auvipy:
better luck next time

Issue #398, opened 2014-08-29
Kombu filesystem transport not thread safe


harlowja:
From looking it over it appears that this transport isn't thread safe.

I've hit errors like http://paste.openstack.org/show/102226/ a few times and am thinking these class of errors are due to files getting moved by more than one thread. Perhaps some basic locking is in order?

mpenalver:
Attached the version we use, which also avoids the copy of the msgs to tmp. Credit to https://github.com/benediktschmitt/py-filelock.
[filesystem.py.tar.gz](https://github.com/celery/kombu/files/3493492/filesystem.py.tar.gz)

jhofeditz:
Thanks! @mpenalver I also have this problem and will try your fix. Can you submit it as a PR?

PR #1596, opened 2022-09-12
Solve Kombu filesystem transport not thread safe

karajan1001:
partly fix: #398
1. add an exclusive lock during the whole exchange file update.
2. add some unit tests for the file lock

This time I tested it on my local machine, which works well with 20 workers. 

I'm struggling with how to test the lock status during the progress, I tried to mock some of the functions `kombu.utils.encoding.str_to_bytes` want to inject a lock operation into it. But what I found is that the actual running function was not the same one I mocked. ( I print the function info in the `_queue_bind`, it is not modified. but if I import the function in `test_filesystem.py` and print it out, it will show a mocked function. Any ideas on this?

### DIFF
--- a/kombu/transport/filesystem.py
+++ b/kombu/transport/filesystem.py
@@ -96,17 +96,15 @@ import shutil
 import tempfile
 import uuid
 from collections import namedtuple
-from contextlib import contextmanager
 from pathlib import Path
 from queue import Empty
 from time import monotonic
 
 from kombu.exceptions import ChannelError
+from kombu.transport import virtual
 from kombu.utils.encoding import bytes_to_str, str_to_bytes
 from kombu.utils.json import dumps, loads
 from kombu.utils.objects import cached_property
-
-from . import virtual
 
 VERSION = (1, 0, 0)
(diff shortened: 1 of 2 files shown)


auvipy:
how is the progress?

karajan1001:
> how is the progress?

The current situation is that I can verify this PR on my local computer ( really easy). And the demo is below

[asciicast](https://asciinema.org/a/VpoYUBuW8iXrQfbI4TOcUqbFN)

We can see that in the current `master` if I start 10 workers together, 5 of them will error out. After switching to the `fix_398` branch, all of the 10 workers are started. 

But the current difficulty is that I'm not sure how to write a test for this. The previous problem raises from that between the 
`reading` and `writing` progress there is no file lock between them and will probably got a content change when we want to write to it(the current status had already been different from when we read it). A rigorous test should be something like testing the status between this two progress. I tried to mock some of the functions inside this progress. But I found some weird behavior when using `celery` together with `mock`,  I successfully mocked a function but when celery runs it, it still uses the old real one.

auvipy (approved):

PR #1599, opened 2022-09-27
Add separate transport option for retry loop timeout

### LINKED ISSUE #1594: Unable to configure connect_timeout and failover at the same time
Hi. My use case is this: I want to use Celery with multiple clustered RabbitMQ servers **without** a load balancer in between.

When the first RabbitMQ server is unreachable, Celery should wait until connection timeout and retry connecting to the next server. From the surface it seems like that configuration is possible: I can pass multiple URLs to `kombu.Connection` and there are more settings to fine-tune failover/retry configuration: `max_retries`, `failover_strategy`, etc.

However, this does not seem to work in practice: when connection to the 1st server **timeouts**, Kombu never tries the 2nd server in the list.

When using `Connection.default_channel`, it calls `_extract_failover_opts()`:

```python
    def _extract_failover_opts(self):
        conn_opts = {'timeout': self.connect_timeout}
```

This `timeout` is then passed `Connection.default_channel` -> `_ensure_connection()` -> `retry_over_time(self._connection_factory, ..., timeout=(timeout from above))`

`self._connection_factory()` and `retry_over_time()` use the exact same `connect_timeout` value when connecting. By the time `_connection_factory` raises a TimeoutError, `retry_over_time`'s timeout has also expired. I don't see any way to pass different values for these timeouts.

I also notice that the `Connection.connect()` method hard-codes `max_retries=1`. This method doesn't seem to be called in my usage, but still I have to ask: WHY? Shouldn't the retry and failover policy be up to user configuration?

----

As a solution, I think there should be two separate settings for timeouts: One to set TCP-level timeout for a single connection attempt, and the other timeout for the retry loop in `retry_over_time()`.

intgr:
Fixes #1594. This only applies when using `Connect.default_channel`.

Before this change, the retry loop timeout was set equal to TCP connect timeout (`connect_timeout`), meaning when first connection attempt timeouted, no retry would be attempted.

Now, if a new transport option `connect_total_timeout` is provided, this overrides `connect_timeout` for the retry loop (but not for TCP connect).

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -868,7 +868,8 @@ class Connection:
             if 'interval_max' in transport_opts:
                 conn_opts['interval_max'] = transport_opts['interval_max']
             if 'connect_retries_timeout' in transport_opts:
-                conn_opts['timeout'] = transport_opts['connect_retries_timeout']
+                conn_opts['timeout'] = \
+                    transport_opts['connect_retries_timeout']
         return conn_opts
 
     @property
--- a/t/unit/test_connection.py
+++ b/t/unit/test_connection.py
@@ -719,8 +719,8 @@ class test_Connection:
             ['server1', 'server2'],
             transport=TimeoutingTransport,
             connect_timeout=1,
-            transport_options={'connect_retries_timeout': 2, 'interval_start': 0,
-                               'interval_step': 0},
+            transport_options={'connect_retries_timeout': 2,
+                               'interval_start': 0, 'interval_step': 0},
         ) as conn:
             conn._establish_connection = Mock(
                 side_effect=conn._establish_connection


intgr:
`Connection._ensure_connection()` gets called from many different places with different sets of arguments, but for some reason `transport_options` only apply when using `Connect.default_channel`. So I didn't touch other call paths.

I tested that this solves my issue when using Celery.

kombu/connection.py
```diff
@@ -867,6 +867,8 @@ def _extract_failover_opts(self):
                 conn_opts['interval_step'] = transport_opts['interval_step']
             if 'interval_max' in transport_opts:
                 conn_opts['interval_max'] = transport_opts['interval_max']
+            if 'connect_total_timeout' in transport_opts:
```
auvipy, line 870: would love to see test coverage for the change

auvipy (changes requested):

intgr:
Also I'm not sure about the naming of `connect_total_timeout` transport option, ideas welcome.

Alternatives I could come up with: `connect_retry_timeout`, `connect_retry_total_timeout`, `connect_retry_loop_timeout`, `total_connect_timeout`...

kombu/connection.py
intgr: Added two tests, for behavor with and without `connect_total_timeout`.

intgr: Added test cases with and without the new option. I hope it's OK to use `time.sleep()` in tests.

auvipy (commented):
how does connect_retry_timeouts sound?

auvipy (changes requested):
Maybe connect_retries_timeout then seems ok. but the tests are failing. can you recheck please?

intgr:
> Maybe connect_retries_timeout then seems ok.

Renamed to `connect_retries_timeout`

> but the tests are failing. can you recheck please?

The failures don't seem to be related to my changes. Let's see if it gets resolved with a rebase.

intgr:
@auvipy Can you approve the workflows so we can see if the test issue is fixed?

auvipy (approved):

PR #1611, opened 2022-10-16
Allow azurestoragequeues transport to be used with Azurite emulator in docker-compose

### LINKED ISSUE #1168: Request: Accept full range of azure storage queue connection parameters
I'd like to test the azure storage queue transport locally using the Azure Storage emulator, Azurite [[link](https://github.com/Azure/Azurite)], which allows you to host a version of the blob and queue storage service locally using a docker image running on localhost accessible via http.

Kombu should be able to connect to it; however, the connection url it accepts doesn't make room for any other connection parameters than `account_name` and `access_key`, limiting you to e.g. the default endpoint (`core.windows.net`) and transport scheme (`https`).

Here is the full range accepted by `QueueService`:

> QueueService(account_name=None, account_key=None, sas_token=None, is_emulated=False, protocol='https', endpoint_suffix='core.windows.net', request_session=None, connection_string=None, socket_timeout=None)

Kombu just looks at the `account_name` and `account_key` based and uses the defaults for the rest:

https://kombu.readthedocs.io/en/latest/reference/kombu.transport.azurestoragequeues.html

At a minimum, by supporting `connection_string`, you could use the `QueueService` itself to parse everything for you. After a brief look into the source, I wasn't able to quickly determine where this change would need to be made - maybe in  `azurestoragequeue.py` or its base class`virtual.Channel`?

I'd be happy to make a PR if I knew where to look.

davidbossanyi:
When integration testing the `azurestoragequeues` transport locally using [Azurite](https://github.com/Azure/Azurite) within a docker-compose stack, the queue url becomes `http://<service>:10001/devstoreaccount1` rather than `http://localhost:10001/devstoreaccount1` where `<service>` is the name of the azurite service in docker-compose. As described in [this issue on the Azure python SDK](https://github.com/Azure/azure-sdk-for-python/issues/19202), this results in a failure to connect to the queue service. This PR applies the suggested workaround in [this comment](https://github.com/Azure/azure-sdk-for-python/issues/19202#issuecomment-1158057765) to enable the kombu transport to work with Azurite in local docker-compose stacks for integration testing.

This PR should close #1168

### DIFF
--- a/kombu/transport/azurestoragequeues.py
+++ b/kombu/transport/azurestoragequeues.py
@@ -179,7 +179,7 @@ class Transport(virtual.Transport):
     can_parse_url = True
 
     @staticmethod
-    def parse_uri(uri: str) -> tuple[Union[str, dict], str]:
+    def parse_uri(uri: str) -> tuple[str | dict, str]:
         # URL like:
         #  azurestoragequeues://STORAGE_ACCOUNT_ACCESS_KEY@STORAGE_ACCOUNT_URL
         #  azurestoragequeues://SAS_TOKEN@STORAGE_ACCOUNT_URL
--- a/kombu/transport/azurestoragequeues.py
+++ b/kombu/transport/azurestoragequeues.py
@@ -35,6 +35,7 @@ from __future__ import annotations
 
 import string
 from queue import Empty
+from typing import Union
 
 from azure.core.exceptions import ResourceExistsError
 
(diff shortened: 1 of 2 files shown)


davidbossanyi:
Seems like the pre-commit CI automatically changes union types from e.g. `Union[dict, str]` to the python 3.10 version `dict | str` which won't work for python < 3.10. Not sure what your preferred way to represent union types is? Let me know and I will alter as required.

auvipy:
import annotation from future let us use python 3.10 annotations in older version :)

auvipy (approved):

PR #1612, opened 2022-10-16
feature: Adding type annotations to Connections

ponponon:
```python
from kombu import Connection
from kombu.pools import connections

connection = Connection('redis://localhost:6379')

with connections[connection].acquire(block=True) as conn:
    print('Got connection: {0!r}'.format(connection.as_uri()))
```

Code editors such as vscode, pycharm, etc. cannot provide intelligent type inference for this `connections[connection].acquire(block=True)` code, so I would like to add typing hint to achieve this (intelligent type inference)

> I would like to add more typing hint to this project, hopefully the author will pass and let me participate in it


auvipy (changes requested):
kombu/pools.py:7:1: F401 'typing.Optional' imported but unused

kombu/pools.py
```diff
@@ -104,9 +105,18 @@ def register_group(group):
 class Connections(PoolGroup):
     """Collection of connection pools."""
 
-    def create(self, connection, limit):
+    def create(self, connection: Connection, limit: int | None):
         return connection.Pool(limit=limit)
 
+    def __getitem__(self, connection: Connection) -> ConnectionPool:
```
auvipy, line 111: I would like to know, what this dunder methods are going to achieve?

ponponon: I have made it very clear

Code editors such as vscode, pycharm, etc. cannot provide intelligent type inference for this connections[connection].acquire(block=True) code, so I would like to add typing hint to achieve this (intelligent type inference)

ponponon: When your mouse hovers over a conn, vscode will tell you that it is the type of this conn is Any, which is very stupid.

When hovering over a conn, I wish vscode would tell me what the explicit type is, instead of a stupid and damned Any

ponponon: And I added these typing hint which will solve this problem very well

ponponon: No type hint 👇
图片

With type annotations 👇
图片


Modern program development is inseparable from the intelligent hints provided by the IDE. With these type annotations, the IDE can provide us with type inference, allowing us to program more efficiently and enjoyably

auvipy (changes requested):
kombu/pools.py:7:1: F401 'typing.Optional' imported but unused
kombu/pools.py:82:80: E501 line too long (80 > 79 characters)
kombu/connection.py:11:1: F401 'typing.Optional' imported but unused
kombu/connection.py:11:1: F401 'typing.Union' imported but unused
kombu/connection.py:1042:80: E501 line too long (93 > 79 characters)
ERROR: InvocationError for command /home/runner/work/kombu/kombu/.tox/flake8/bin/flake8 -j2 kombu t (exited with code 1)

kombu/connection.py
```diff
@@ -8,7 +8,7 @@
 from contextlib import contextmanager
 from itertools import count, cycle
 from operator import itemgetter
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Optional, Union
```
auvipy, line 11:
```suggestion
from typing import TYPE_CHECKING, Any
```

kombu/pools.py
```diff
@@ -4,8 +4,9 @@
 
 import os
 from itertools import chain
+from typing import Optional
```
auvipy, line 7:
```suggestion
```

auvipy (changes requested):

kombu/connection.py
```diff
@@ -1039,6 +1039,9 @@ def prepare(self, resource):
         resource._debug('acquired')
         return resource
 
+    def acquire(self, block: bool = False, timeout: int | float | None = None) -> Connection:
```
auvipy, line 1042: is this method newly introduced?

Kludex (commented):
It will be cool to have an entry on the `pyproject.toml` associated with the annotation. 👀

kombu/connection.py
ponponon: Not new, ConnectionPool inherits from Resource 

Resource has an acquire method, but since Resource is an abstract class (abstracting the connection as a resource), this `-> Connection` type annotation should not be added to the acquire method of Resource, so I added an acquire method to ConnectionPool and added the  I added an acquire method to the ConnectionPool and added the `-> Connection` type annotation to the ConnectionPool to make it semantic

`kombu/connection.py`

```python
class ConnectionPool(Resource):
    """Pool of connections."""
```

`kombu/resource.py`

```python
class Resource:
    """Pool of resources."""

    def acquire(self, block=False, timeout=None):
        """Acquire resource.

        Arguments:
            block (bool): If the limit is exceeded,
                then block until there is an available item.
            timeout (float): Timeout to wait
                if ``block`` is true.  Default is :const:`None` (forever).

        Raises:
            LimitExceeded: if block is false and the limit has been exceeded.
```

thedrow (changes requested):
Is there a better way than re-defining existing methods just to provide typing information for them?
I'm not satisfied with the current approach.

PR #1621, opened 2022-11-23
SQS: avoid excessive GetQueueURL calls by using cached queue url

### LINKED ISSUE #1618: SQS transport calls get_queue_url *every* time before receive_message - should cache?
When using the SQS transport (in my case with celery) I see a *lot* of `GetQueueUrl` AWS API calls (~900 per hour) which seems unnecessary given the queue URL essentially never changes.

As far as I can see this is because:
 - `kombu.transport.SQS.Channel._get_bulk_async` -> `_get_async` -> calls `_get_from_sqs`...
 - ...and `_get_from_sqs` does `queue_url = connection.get_queue_url(queue)` (if you you haven't specified `self.predefined_queues`)

Could `_get_from_sqs` check `self._queue_cache[queue]` instead to save re-fetching the queue URL every time?

There seems to be precedent for this e.g:
 - both `Channel._get_bulk` and `Channel._size` call `_new_queue`...
 - ...which checks `_queue_cache` regardless of whether `self.predefined_queues` exists or not

If the lack of caching is deliberate in this specific case, why is it different from the other cases where `_queue_cache` is used?

I realise that this could be avoided by putting the URLs in `predefined_queues`, but in our environment there are a lot of instances of the application and this extra faff when the discovery from `queue_name_prefix` combined with the queue name works excellently.

sparrowt:
The first commit fixes #1618.

The second clarifies some comments & variable names which I found confusing while trying to figure out what was going on.

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -238,7 +238,7 @@ class Channel(virtual.Channel):
     _predefined_queue_async_clients = {}  # A client for each predefined queue
     _sqs = None
     _predefined_queue_clients = {}  # A client for each predefined queue
-    _queue_cache = {}
+    _queue_cache = {}  # SQS queue name => SQS queue URL
     _noack_queues = set()
     QoS = QoS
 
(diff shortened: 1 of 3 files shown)


auvipy (changes requested):
this will also need proper test coverage

sparrowt:
Hi @auvipy, thanks for looking at my PR - aside from coverage are you happy with the caching change in principle?

I haven't added any new methods or logical code paths, only made a small modification to one existing path - are these core methods not already covered? Is there a way for me to access the coverage results which https://github.com/celery/kombu/actions/runs/3533431074/jobs/6067061888 seems to be recording? Thanks in advance.

auvipy:
to be honest, the actual change is hard to review, due to lots of renaming. can you please add tests? I am in favour of caching if it doesn't break anything, rather improve excessive calls to GetQURL

thedrow:
This would be easier to review if it included just the changes required to avoid excessive GetQueueURL calls.
The rest of the changes are valuable but make the change harder to review.

@sparrowt Any chance you'll refactor this PR and break it down into several PRs?

auvipy:
@sparrowt it would be great if you can revisit this

sparrowt:
Hi folks, sorry for the delayed response - I deliberately made this PR with 2 separate commits, the first of which is a minimal fix for the problem, then the second which renames things which appeared to be confusingly named (whilst I was trying to figure out how everything worked).

Would you still prefer the 2nd commit to be removed to a separate PR potentially, or are you happy to review it commit by commit?

Regarding testing, I assume `_get_async` as a crucial code path is already covered by tests, so hopefully the continued passing of these tests should be sufficient? As mentioned above I don't have access to the coverage details which the CI seemed to be producing.

kombu/transport/SQS.py
```diff
@@ -569,11 +569,14 @@ def _get_bulk_async(self, queue,
...
     def _get_async(self, queue, count=1, callback=None):
-        q = self._new_queue(queue)
+        q_url = self._new_queue(queue)
         qname = self.canonical_queue_name(queue)
         return self._get_from_sqs(
-            qname, count=count, connection=self.asynsqs(queue=qname),
-            callback=transform(self._on_messages_ready, callback, q, queue),
+            queue_name=qname, queue_url=q_url, count=count,
+            connection=self.asynsqs(queue=qname),
+            callback=transform(
```
auvipy, line 581: this art will need appropriate test coverage

kombu/transport/SQS.py
```diff
@@ -583,24 +586,14 @@ def _on_messages_ready(self, queue, qname, messages):
...
 
-    def _get_from_sqs(self, queue,
-                      count=1, connection=None, callback=None):
+    def _get_from_sqs(self, queue_name, queue_url,
+                      connection, count=1, callback=None):
         """Retrieve and handle messages from SQS.
 
         Uses long polling and returns :class:`~vine.promises.promise`.
         """
-        connection = connection if connection is not None else queue.connection
```
auvipy, line 592: and removal of this too

auvipy (changes requested):
beside appropriate regression tests, please re-base as well

kombu/transport/SQS.py
sparrowt: Added in the latest commit

kombu/transport/SQS.py
sparrowt: This is also covered by the unit test in the latest commit

sparrowt:
I've added coverage for the existing core code path which was missing it (and also rebased my PR), how's that? :)

auvipy:
I've restarted the CI

auvipy (approved):

auvipy:
thank you

PR #1632, opened 2022-12-18
add type annotations to kombu/transport/azurestoragequeues

jasonwbarnett:
This isn't complete, but at least it's a start. I'm open to any feedback.

### DIFF
--- a/kombu/transport/azurestoragequeues.py
+++ b/kombu/transport/azurestoragequeues.py
@@ -54,6 +54,7 @@ from __future__ import annotations
 
 import string
 from queue import Empty
+from typing import Any, Optional
 
 from azure.core.exceptions import ResourceExistsError
 
@@ -85,11 +86,11 @@ CHARS_REPLACE_TABLE = {
 class Channel(virtual.Channel):
     """Azure Storage Queues channel."""
 
-    domain_format = 'kombu%(vhost)s'
-    _queue_service = None
-    _queue_name_cache = {}
-    no_ack = True
-    _noack_queues = set()
+    domain_format: str = 'kombu%(vhost)s'
+    _queue_service: Optional[QueueServiceClient] = None
+    _queue_name_cache: dict[Any, Any] = {}
+    no_ack: bool = True
+    _noack_queues: set[Any] = set()
 
     def __init__(self, *args, **kwargs):
         if QueueServiceClient is None:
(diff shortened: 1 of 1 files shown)


kombu/transport/azurestoragequeues.py
```diff
@@ -59,11 +60,11 @@
 class Channel(virtual.Channel):
     """Azure Storage Queues channel."""
 
-    domain_format = 'kombu%(vhost)s'
-    _queue_service = None
-    _queue_name_cache = {}
-    no_ack = True
-    _noack_queues = set()
+    domain_format: str = 'kombu%(vhost)s'
+    _queue_service: QueueServiceClient | None = None
```
thedrow, line 64: Shouldn't _queue_service be [Optional](https://docs.python.org/3/library/typing.html#typing.Optional)?

kombu/transport/azurestoragequeues.py
```diff
@@ -173,9 +174,9 @@ class Transport(virtual.Transport):
 
     Channel = Channel
 
-    polling_interval = 1
-    default_port = None
-    can_parse_url = True
+    polling_interval: int = 1
+    default_port: int | None = None
```
thedrow, line 178: The type here is Optional too

thedrow (changes requested):
I have one thing I need you to change and then we can merge this PR

kombu/transport/azurestoragequeues.py
jasonwbarnett: I agree, but the pre-commit.ci automatically corrected it and switched it from `Optional[QueueServiceClient]` to this. I just assumed this was an alternative way to declare the same thing, but outside of this project I would always use the `Optional[QueueServiceClient]` notation.

What should I do?

jasonwbarnett: I made the updates, and I assume pre-commit.ci will add a commit correcting it, but we'll see.

jasonwbarnett: Yeah, it corrected it so I just re-pushed the original content that you commented on. Should I try to update the pre-commit.ci configuration or merge as is?

jasonwbarnett:
> I have one thing I need you to change and then we can merge this PR

@thedrow I reviewed the inline comments and responded with why I didn't so that. Please let me know how you would like me to proceed.

jasonwbarnett:
I ended up finding the root cause of this. It turns out `pyupgrade` is responsible for making this change, see: https://github.com/asottile/pyupgrade#pep-604-typing-rewrites

jasonwbarnett:
I just rebased this PR off of main.

kombu/transport/azurestoragequeues.py
thedrow: If you can update the pre-commit config that would be great but I wouldn't block merging this PR without that config change.

thedrow:
Yeah let's configure pyupdate to keep `Optional`.

jasonwbarnett:
@thedrow @Nusnus This is ready to be merged.

Nusnus:
LGTM but I'll let @thedrow have the final word here

auvipy (approved):

PR #1640, opened 2023-01-02
azure service bus: add type annotations and use cached property

jasonwbarnett:
Title is self explanatory.

I converted some instance methods from `@property` to `@cached_property` because it's less verbose and I assumed that was the actual intent.

There is one underlying change too which I can't tell if it was a bug or by design, I'd like someone else to review to make sure I'm not changing the intended behavior.

### DIFF
--- a/kombu/transport/azureservicebus.py
+++ b/kombu/transport/azureservicebus.py
@@ -57,7 +57,7 @@ from __future__ import annotations
 
 import string
 from queue import Empty
-from typing import Any
+from typing import Any, Dict, Set
 
 import azure.core.exceptions
 import azure.servicebus.exceptions
@@ -87,8 +87,8 @@ class SendReceive:
     def __init__(self,
                  receiver: ServiceBusReceiver | None = None,
                  sender: ServiceBusSender | None = None):
-        self.receiver = receiver  # type: ServiceBusReceiver
-        self.sender = sender  # type: ServiceBusSender
+        self.receiver: ServiceBusReceiver = receiver
+        self.sender: ServiceBusSender = sender
 
     def close(self) -> None:
         if self.receiver:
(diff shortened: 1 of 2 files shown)


kombu/transport/azureservicebus.py
```diff
@@ -229,7 +227,7 @@ def _delete(self, queue: str, *args, **kwargs) -> None:
         """Delete queue by name."""
         queue = self.entity_name(self.queue_name_prefix + queue)
 
-        self._queue_mgmt_service.delete_queue(queue)
+        self.queue_mgmt_service.delete_queue(queue)
```
jasonwbarnett, line 230: This is the change that I called out in the PR description. I'm not sure if this was accessing the variable instead of using the method intentionally or was an accident.

auvipy:
the additional changes here is causing service request error in the tests

jasonwbarnett:
> the additional changes here is causing service request error in the tests 

Yeah, I see that. Thanks for calling out. That's why this is marked as a draft. I am planning to refactor the tests sometime this week.

jasonwbarnett:
@auvipy this is ready for review 👀

thedrow (approved):

thedrow:
Can you please rebase so that the tests would run?
We had a misconfiguration in our CI workflow because we renamed the master branch to main.

jasonwbarnett:
> Can you please rebase so that the tests would run? We had a misconfiguration in our CI workflow because we renamed the master branch to main.

@thedrow this MR is already rebased off of `main`.

jasonwbarnett:
@thedrow I overlooked what your intent was. I just amended the commit and re-pushed.

auvipy (approved):

PR #1659, opened 2023-03-01
refactor: Refactor utils/json

last-partizan:


kombu/utils/json.py
```diff
@@ -3,88 +3,66 @@
...
-    from django.utils.functional import Promise as DjangoPromise
-except ImportError:  # pragma: no cover
-    class DjangoPromise:
-        """Dummy object."""
+textual_types = ()
 
+try:
+    from django.utils.functional import Promise
 
-class _DecodeError(Exception):
```
last-partizan, line 18: I removed this _DecodeError, because it was not raised anywhere.

It was leftover for compatibility with `simplejson` library which is long gone.

t/unit/utils/test_json.py
```diff
@@ -29,19 +28,18 @@ class test_JSONEncoder:
...
-            'datetime': now,
-            'tz': now_utc,
-            'time': now.time().isoformat(),
-            'date':  date(now.year, now.month, now.day),
+            'time': now.time(),
         }
 
+        serialized = loads(dumps(original))
+
+        assert serialized == original
```
last-partizan, line 41: I improved this test, now it exactly matches original (i added `time` type support).

t/unit/utils/test_json.py
```diff
@@ -53,8 +51,10 @@ def test_binary(self, message):
         }
 
     def test_Decimal(self):
-        d = Decimal('3314132.13363235235324234123213213214134')
-        assert loads(dumps({'d': d})), {'d': str(d)}
+        original = {'d': Decimal('3314132.13363235235324234123213213214134')}
+        serialized = loads(dumps(original))
+
+        assert serialized == original
```
last-partizan, line 57: Decimal is also preserved as Decimal, there is no need to convert it to string.

t/unit/utils/test_json.py
```diff
@@ -70,7 +70,7 @@ def test_UUID(self):
         for constructor in constructors:
             id = constructor()
             loaded_value = loads(dumps({'u': id}))
-            assert loaded_value, {'u': id}
+            assert loaded_value == {'u': id}
```
last-partizan, line 73: This was error. It should be compared, not "assert first", second argument is detailed assertion message.

t/unit/utils/test_json.py
```diff
@@ -103,9 +103,3 @@ def test_loads_bytes(self):
         assert loads(
             str_to_bytes(dumps({'x': 'z'})),
             decode_bytes=True) == {'x': 'z'}
-
-    def test_loads_DecodeError(self):
-        _loads = Mock(name='_loads')
-        _loads.side_effect = _DecodeError(
-            MagicMock(), MagicMock(), MagicMock())
-        assert loads(dumps({'x': 'z'}), _loads=_loads) == {'x': 'z'}
```
last-partizan, line 111: This test is removed, because we no longer use _DecodeError or simplejson.

last-partizan:
@auvipy @Nusnus please take a look.

Nusnus:
> @auvipy @Nusnus please take a look.

Cool - I started the CI, let's see how it goes

kombu/utils/json.py
```diff
@@ -3,88 +3,66 @@
...
-                        "__base64__": True,
-                    }
-            return super().default(o)
-
-
-_default_encoder = JSONEncoder
-
-
-def dumps(s, _dumps=json.dumps, cls=None, default_kwargs=None, **kwargs):
+        # Bytes is slightly trickier, so we cannot pu them directly
```
last-partizan, line 35:
```suggestion
        # Bytes is slightly trickier, so we cannot put them directly
```

last-partizan:
@Nusnus Start it again, i was using walrus operator which doesn't work for old python. I hope typing is working fine.

t/unit/utils/test_json.py
auvipy: good catch

auvipy:
we should use syntax that works in python 3.7

Nusnus:
> @Nusnus Start it again, i was using walrus operator which doesn't work for old python. I hope typing is working fine.

Done

auvipy:
I will look into this again tomorrow morning

kombu/utils/json.py
```diff
@@ -96,14 +78,51 @@ def loads(s, _loads=json.loads, decode_bytes=True, object_hook=object_hook):
...
+
+_encoders: dict[type, tuple[str, EncoderT]] = {}
+_decoders: dict[str, DecoderT] = {
+    "bytes": lambda o: o.encode("utf-8"),
+    "base64": lambda o: base64.b64decode(o.encode("utf-8")),
+}
+
+# NOTE: datetime should be registered before date,
+# because datetime is also instance of date.
+register_type(datetime, "datetime", datetime.isoformat, datetime.fromisoformat)
```
auvipy, line 114: are we registering them 5 times?

last-partizan: Why do you think so?

We register each type once, they're all registered when module is imported.

last-partizan: If i add `print()` to register_type, it looks like this:

```
pytest --disable-warnings t/unit/utils/test_json.py -vs
...
register_type datetime
register_type date
register_type time
register_type decimal
register_type uuid

 t/unit/utils/test_json.py::test_JSONEncoder.test_datetime ✓
 ...
```

PR #1674, opened 2023-03-20
Update azureservicebus.txt cpython only

auvipy:


auvipy (commented):
CI tests are green now

PR #1682, opened 2023-04-01
test azureservicebus pure python beta

auvipy:


auvipy (commented):
this version support pure python and pypy both and feature complete as er microsoft team

PR #1690, opened 2023-04-08
fix: don't crash on `properties`.`body_encoding`: `utf-8`

pankgeorg:
when a message has an `properties`.`body_encoding` of `utf-8` (e.g. the default on https://github.com/node-celery-ts/node-celery-ts), celery workers crash with an error `AttributeError: 'NoneType' object has no attribute 'decode'`. `utf-8` is functionally equivalent to not providing an encoding, so this should be good.

### DIFF
--- a/kombu/transport/virtual/base.py
+++ b/kombu/transport/virtual/base.py
@@ -802,12 +802,12 @@ class Channel(AbstractChannel, base.StdChannel):
         self.exchange_types = None
 
     def encode_body(self, body, encoding=None):
-        if encoding:
+        if encoding and encoding.lower() != 'utf-8':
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
can you please add unit tests for the changes?

pankgeorg:
Yes I would love to! can you help be a bit figuring out where the related tests are?

auvipy:
https://github.com/celery/kombu/tree/main/t/unit/transport

auvipy (commented):
this should be the primary place for adding tests 

https://github.com/celery/kombu/blob/main/t/unit/transport/virtual/test_base.py

auvipy (approved):

PR #1691, opened 2023-04-10
fix: handle keyerror when message is not found in qos and perform basic_ack

IsmaelJS:
When no_ack=True, the received message is not appended into qos, and thus KeyError is raised on each received message.
This fix aims to handle the KeyError raised and perform the basic_ack instead.

### DIFF
--- a/kombu/transport/azureservicebus.py
+++ b/kombu/transport/azureservicebus.py
@@ -273,23 +273,24 @@ class Channel(virtual.Channel):
         return msg
 
     def basic_ack(self, delivery_tag: str, multiple: bool = False) -> None:
-        delivery_info = self.qos.get(delivery_tag).delivery_info
-
-        if delivery_info['exchange'] in self._noack_queues:
-            return super().basic_ack(delivery_tag)
-
-        queue = delivery_info['azure_queue_name']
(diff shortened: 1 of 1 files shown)


kombu/transport/azureservicebus.py
```diff
@@ -273,23 +273,24 @@ def _get(
...
-            return super().basic_ack(delivery_tag)
-
-        queue = delivery_info['azure_queue_name']
-        # recv_mode is PEEK_LOCK when ack'ing messages
-        queue_obj = self._get_asb_receiver(queue)
-
         try:
-            queue_obj.receiver.complete_message(delivery_info['azure_message'])
-        except azure.servicebus.exceptions.MessageAlreadySettled:
+            delivery_info = self.qos.get(delivery_tag).delivery_info
```
auvipy, line 277: thanks for the PR! we need to verify this change with some unit tests

auvipy (changes requested):

kombu/transport/azureservicebus.py
IsmaelJS: done!

auvipy (approved):

PR #1694, opened 2023-04-13
fix mongodb transport obsolete calls

### LINKED ISSUE #1689: Broken MongoDB transport with `pymongo>=4`
As stated in the [4.x breaking changes of `pymongo`](
https://github.com/mongodb/mongo-python-driver/blob/acc6605ea119ca62b4d9dbff07b62bbc7c86c3d8/doc/changelog.rst#breaking-changes-in-40), multiple collection methods have been removed, which cause some errors depending on which `celery` command is applied.

An example is `celery status` (others are also causing similar errors), which ends up raising the following:
```python
TypeError: 'Collection' object is not callable. If you meant to call the 'remove' method on a 'Collection' object it is failing because no such method exists.
``` 
that originates from the following call:
https://github.com/celery/kombu/blob/2de7f9f038dd62e097e490cb3fa609067c1c3c36/kombu/transport/mongodb.py#L259-L260

This transport implementation should be updated with the new corresponding collection methods (e.g.: `delete_one` instead of `remove`). Most of the method mappings are introduced in `pymongo==3.0` where the removed methods in 4.x were originally deprecated. The transport should probably perform some kind of check with the installed `pymongo` version for backward support as well.

fmigneault:
## changes

- replace obsolete collection `remove` by `delete_one`
- add test mock `spec` to check valid methods corresponding to the real objects
  (`AttributeError` raised if real class does not have this attribute/method)

## tests

Validated with latest celery using `celery inspect` calls that were failing before this fix.

## references

- fixes #1689

### DIFF
--- a/kombu/transport/mongodb.py
+++ b/kombu/transport/mongodb.py
@@ -65,6 +65,7 @@ class BroadcastCursor:
 
     def __init__(self, cursor):
         self._cursor = cursor
+        self._offset = 0
         self.purge(rewind=False)
 
     def get_size(self):
@@ -77,7 +78,7 @@ class BroadcastCursor:
         if rewind:
             self._cursor.rewind()
 
-        # Fast forward the cursor past old events
+        # Fast-forward the cursor past old events
         self._offset = self._cursor.collection.count_documents({})
         self._cursor = self._cursor.skip(self._offset)
 
@@ -221,7 +222,7 @@ class Channel(virtual.Channel):
         if queue in self._fanout_queues:
             self._get_broadcast_cursor(queue).purge()
         else:
-            self.messages.remove({'queue': queue})
+            self.messages.delete_many({'queue': queue})
 
         return size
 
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
you used delete one and delete many, on theory they seem to be OK. but I'm not sure about the CI failures. can you please recheck?

fmigneault:
The CI failures seems unrelated to the changes (failing dependency install?). 
I've successfully executed the test suite locally with pytest for everything related to mongodb.

auvipy:
restarted the builld

fmigneault:
Seems related to the following: 
https://community.codecov.com/t/codecov-yanked-from-pypi-all-versions/4259

auvipy:
uggh have to find out what to clean from the CI

auvipy:
but pytest-cov should not be affected

fmigneault:
`pytest-cov` works, it is only the coverage uploader that fails. 
I use the following approach in multiple repositories: https://github.com/crim-ca/weaver/blob/master/.github/workflows/tests.yml#L134-L141

auvipy (changes requested):
can you please pull from main branch?

auvipy (approved):

auvipy:
thanks a lot

PR #1700, opened 2023-04-18
Update tox.ini for lint

auvipy:


sparrowt:
Looks like the problem now is `pydocstyle==1.1.1` is pretty old (2016) and the deprecated way it does `from collections import Set` finally stopped working in python 3.10

`pydocstyle==6.2.0` lists 'support for python 3.10' in the [release notes](https://www.pydocstyle.org/en/stable/release_notes.html#january-2nd-2023) although [the PR](https://github.com/PyCQA/pydocstyle/pull/554) suggests this is more about running pydocstyle's own CI on python 3.10 so if there are issues jumping straight to pydocstyle 6.x then an intermediate version might work fine.

auvipy:
yeah we should align it with celery version

sparrowt:
Also looks like `mypy` is left behind trying to run on 3.8 e.g. that was already failing on this run https://github.com/celery/kombu/actions/runs/4398408364/jobs/7702301475 before pr 1663 was merged

PR #1708, opened 2023-05-11
chore(python): remove Python 3.7 support

stegayet:
Improvement found while I was working on https://github.com/celery/kombu/pull/1705.

## Description

As Python 3.7 will reach end-of-life in June 2023, we can get rid of Python 3.7 support. 
Python 3.8.1 instead of 3.8.0 is required as a requirement for `flake8>=6.0.0`, cf. https://github.com/celery/kombu/pull/1709.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,7 +98,7 @@ setup(
     },
     platforms=['any'],
     license='BSD-3-Clause',
-    python_requires=">=3.7",
+    python_requires=">=3.8.1",
     install_requires=reqs('default.txt'),
     tests_require=reqs('test.txt'),
     extras_require={
@@ -125,7 +125,6 @@ setup(
         'Programming Language :: Python',
         'Programming Language :: Python :: 3 :: Only',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
         'Programming Language :: Python :: 3.10',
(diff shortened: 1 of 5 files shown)


kombu/serialization.py
```diff
@@ -28,8 +27,6 @@ def _decode(t, coding):
 
 pickle_load = pickle.load
 
-#: We have to use protocol 4 until we drop support for Python 3.6 and 3.7.
-pickle_protocol = int(os.environ.get('PICKLE_PROTOCOL', 4))
```
auvipy, line 32: can we not remove this part just for next release?

kombu/serialization.py
```diff
@@ -333,7 +330,7 @@ def register_pickle():
     you to python clients.
     """
     def pickle_dumps(obj, dumper=pickle.dumps):
-        return dumper(obj, protocol=pickle_protocol)
```
auvipy, line 336: and this

auvipy (changes requested):

setup.py
```diff
@@ -98,7 +98,7 @@ def readme():
     },
     platforms=['any'],
     license='BSD-3-Clause',
-    python_requires=">=3.7",
+    python_requires=">=3.8.1",
```
auvipy, line 101:
```suggestion
    python_requires=">=3.8",
```

auvipy (changes requested):

auvipy (approved):

PR #1709, opened 2023-05-11
chore(pkgutils): upgrade flake8 to 6.0.0

stegayet:
Improvements found while I was working on https://github.com/celery/kombu/pull/1705.

This PR upgrades `flake8` to 6.0.0 in `pkgutils.txt` as we are already using 6.0.0 in pre-commit.

### DIFF
--- a/requirements/pkgutils.txt
+++ b/requirements/pkgutils.txt
@@ -1,6 +1,6 @@
 setuptools>=47.0.0
 wheel>=0.29.0
-flake8==5.0.4
+flake8==6.0.0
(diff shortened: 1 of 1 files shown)


auvipy (commented):
isn't it python 3.8+ only?

stegayet:
Your are right, after checking again this is Python 3.8.1 only, so this has to be merged after https://github.com/celery/kombu/pull/1708, I'll also modify the previous PR to require 3.8.1 and not only 3.8.

PR #1710, opened 2023-05-11
chore(requirements): improve requirements management

stegayet:
Improvements found while I was working on https://github.com/celery/kombu/pull/1705.

This PR:
- gets rid of Python 2 specific requirements as `kombu` does not support this version anymore (unused across the repository).
- unifies Linux and Windows test dependencies by adding platform markers in `test-ci.txt` and getting rid of `test-ci-windows.txt`.

Diff between `test-ci.txt` and `test-ci-windows.txt`:
```
>> diff requirements/test-ci.txt requirements/test-ci-windows.txt        main ✭
1a2
> codecov
3d3
< -r extras/mongodb.txt
8d7
< -r extras/sqs.txt
9a9
> -r extras/librabbitmq.txt
```


auvipy (changes requested):
looks better but I will cross check before merge

requirements/test-ci.txt
```diff
@@ -1,11 +1,15 @@
 pytest-cov
+codecov; sys_platform == 'win32'
```
auvipy, line 2: should this be win32 or !win32?

auvipy (changes requested):

requirements/test-ci.txt
stegayet: It should be `win32` as this is windows specific as per current `test-ci-windows.txt`

PR #1712, opened 2023-05-11
chore(structure): restructure project layout

stegayet:
Towards https://github.com/celery/kombu/pull/1705.

This PR restructures project layout as expected by poetry. Only needed change is changing from `t/` to `tests/` to follow standard convention.

With this change, we can safely remove `packages`  argument to `setup` call.


auvipy (commented):
why https://github.com/celery/kombu/pull/1723 don't work stand alone? can you share more detail please

stegayet:
As per [your CI run](https://github.com/celery/kombu/actions/runs/4976322796/jobs/8904278991?pr=1723), it says:
```
  3.9-linux-unit: FAIL code error: Multiple top-level packages discovered in a flat-layout: ['t', 'kombu'].
``` 
First commit in this PR switches from `t` to `tests` and fixes this issue IMO :smile:

Issue #1699, opened 2023-04-18
CI workflow failing: lint (3.10) is trying to use python3.9


sparrowt:
### Problem
Recently the 'CI' workflow has been failing because for some reason the `lint (3.10)` step is trying to find `python3.9`
```
Run tox -v -e flake8 -- -v
flake8: find interpreter for spec PythonSpec(major=3, minor=9)
...
flake8: skipped because could not find python interpreter with spec(s): python3.9
  flake8: SKIP (0.61 seconds)
  evaluation failed :( (0.84 seconds)
Error: Process completed with exit code 255.
```
This seems to be happening on all PRs e.g. [like this](https://github.com/celery/kombu/actions/runs/4728800759/jobs/8390814738)

### Cause
https://github.com/celery/kombu/pull/1663 added python 3.10 to CI and set it as the `python-version` under `lint` in `ci.yaml` but didn't change the `basepython` for `flake8` and friends in tox.ini to match.

Looking back to the change for python 3.9 (4f77faab77bafce288d2baa143e66d4ec870d2eb) we do see that flake8 & friends were moved down a line so I think the same needs to happen here (though it feels odd that we have to specify the same thing in 2 places).

auvipy:
thanks a lot for the investigation

sparrowt:
Np! I was going to open a PR today but I see you already have ❤ #1700

stegayet:
This has been fixed in https://github.com/celery/kombu/pull/1718, thanks a lot for your initial investigation @sparrowt :smile:

auvipy:
thanks both of you

PR #1718, opened 2023-05-14
chore(ci): fix lint job

stegayet:
Follow-up on https://github.com/celery/kombu/issues/1699.

This PR:
- moves all lint steps to Python 3.10
- upgrades `pydocstyle` to 6.3.0 to have Python 3.10 support
- excludes some rules for `pydocstyle` to skip failures for now (let me know if you want to fix them in this PR)
- fixes undetected `mypy` error


stegayet:
I’ve run last checks, this is ready for review 👍

setup.cfg
```diff
@@ -48,7 +48,7 @@ files =
 
 
 [pep257]
-ignore = D102,D104,D203,D105,D213
+ignore = D102,D107,D104,D203,D105,D205,D212,D213,D401,D412,D406,D407,D411,D413,D417
```
auvipy, line 51: are the failures too much to fix?

stegayet: I have to dig it into these issues but it seems to be only formatting issues, I can have a look.

stegayet: I fixed most of the rules, the other ones require some doc rewrite that can maybe handled later ?

auvipy (changes requested):
great job

auvipy (approved):

PR #1734, opened 2023-05-26
prevent event loop polling from stopping on active redis connections 

pawl:
Fixes: https://github.com/celery/celery/discussions/7276

Caused by: https://github.com/celery/kombu/pull/1476

Currently, the Redis transport will always remove the last function added to the event loop (`on_tick`) regardless of which connection is disconnected. If you restart Redis enough (and repeatedly cause connection errors), eventually `_on_connection_disconnect` will remove a function from the event loop for the only active connection and the worker will get stuck.

I've been seeing these issues with workers getting stuck after I migrated from the RabbitMQ broker to Redis on Celery 5.2.6 and Kombu 5.2.3.

This PR changes it to track which event loop function belongs to which connection and remove the correct function from `on_tick`.


pawl:
@auvipy @thedrow would either of you guys be able to review this? And do you think this is the right approach? 

I was able to reproduce this issue locally using this example repo: https://github.com/pb-dod/celery_pyamqp_memory_leak/tree/test_redis_disconnect

I was also able to verify this fix works there too.

I think this fix is somewhat critical, because it's likely making running Celery on the Redis broker unreliable for everyone who upgraded to at least v5.2.3.

kombu/transport/redis.py
```diff
@@ -1288,6 +1288,9 @@ def __init__(self, *args, **kwargs):
         # All channels share the same poller.
         self.cycle = MultiChannelPoller()
 
+        # tracks event loop (on_tick) entries by connection
+        self.on_poll_start_by_connection = {}
```
auvipy, line 1292: are we going to undo https://github.com/celery/kombu/pull/1476 by this changes or adjusting that? can you add integration tests as well?

auvipy (changes requested):

kombu/transport/redis.py
pb-dod: > are we going to undo https://github.com/celery/kombu/pull/1476 by this changes or adjusting that?

This is an adjustment since it will still remove functions from `on_tick` after disconnection occurs to prevent memory leaks, but now we'll be making sure it removes the correct function. I can do some of the same tests I did in https://github.com/celery/kombu/pull/1476 to make sure the leak didn't come back.

> can you add integration tests as well?

I'll work on adding that integration test.

auvipy: OK looking forward to that

pawl:
@auvipy I tested this change in production today, and I'm still seeing the same behavior where workers stop responding the worker heartbeats after about an hour (the red lines are a deployment start/finish):
unnamed

I have INFO level logging turned on, and I see no indication in the logs about why the workers get stuck (the last log message is often a successful task run). Workers on different queues with very different workloads get stuck too.

I'm not seeing the issue with workers getting stuck when I use RabbitMQ as a broker.

I think this PR fixes an issue around workers getting stuck after several Redis disconnections, but now I'm not sure it fixes the primary cause of https://github.com/celery/celery/discussions/7276 This also explains why I was getting this issue without seeing Redis connection errors in my logs.

pawl:
@auvipy I added the integration test: https://github.com/celery/kombu/pull/1734/commits/34e366d733ba8010f574a39a93a47d46fbe7bbd2

It looks like it's revealing a problem. Good call on adding it!

I assumed the two `connection` objects here were the same:
https://github.com/celery/kombu/blob/2df5be26ea6fa4b7979aab7e110613ea1efaf266/kombu/transport/redis.py#L1307
https://github.com/celery/kombu/blob/2df5be26ea6fa4b7979aab7e110613ea1efaf266/kombu/transport/redis.py#L1300

However:
* `register_with_event_loop`'s `connection` is a `Transport`
* `_on_disconnect`'s `connection` is a Redis `Connection`.

The way I have it currently won't work because I'm adding `Transport`s as keys to `on_poll_start_by_connection` and those will never match with Redis connection objects.

I'll need to re-think the solution for this.

What I tested in production today effectively disabled the fix from https://github.com/celery/kombu/pull/1476 and workers were still getting stuck.

auvipy (changes requested):
can we close this in favor of https://github.com/celery/kombu/pull/2007?

auvipy:
> can we close this in favor of #2007?

also should we try to extract relevant integrations tests from here to add to the test suite?

PR #1735, opened 2023-05-29
Support redis >= 4.5.2

danigm:
The redis-py 4.5.2 changes the UnixDomainSocketConnection class so now it inherits from AbstractConnection:
https://github.com/redis/redis-py/releases/tag/v4.5.2

This patch makes sure that the health_check_interval parameter is checked for the __init__ method of the main class and also the bases, so it doesn't fail with the newer version of redis-py.


auvipy (changes requested):
thanks for the patch. the related integration test also need some adjustment. also please check if the unit tests need some adjustment

auvipy (approved):

auvipy:
thanks for the contribution! if you can also check if celery needs any code update as well?

PR #1747, opened 2023-06-07
Update PyCurl version in SQS

auvipy:


kumekay:
@auvipy Is this only going to be released at the end of the month or can I expect it sooner? 
(Without it poetry solver has troubles and I have to pin kombu to 5.2.4)

kumekay:
Great, thank you!

PR #1629, opened 2022-12-15
Allowing `Connection.ensure()` to retry on specific exceptions given by policy

Nusnus:
To be used with [task_publish_retry_policy](https://github.com/celery/celery/blob/7b585138af8318d62b8fe7086df7e85d110ac786/celery/app/defaults.py#L270)

PR #1755, opened 2023-06-18
Reverted unwanted constraint introduced in #1629 with max_retries

Nusnus:
#1629 introduced a constraint forcing setting `max_retries` when it was originally designed to be able to be `None` for infinite retries.

This PR relaxes back the constraint as it was before while keeping the change from #1629


auvipy (approved):
OK thanks for the partial revert. I hope to get chance to review all big PR O:)

PR #1759, opened 2023-06-22
[fix #1726] Use boto3 for SQS async requests

### LINKED ISSUE #1726: [Bug] AWS Connection.py Using Private Botocore `get_response` Which Expects JSON But is Being Passed XML
Hello Kombu folks!

### tl;dr
Kombu is using an internal/private method, [get_response](https://github.com/boto/botocore/blob/develop/botocore/response.py#L180-L201), from Botocore which changed behaviour recently due to other Botocore updates to the wire protocol used for SQS, which breaks Kombu.
The change has been temporarily reverted, but Kombu cannot depend on a stable behaviour of this method going forward as the changes will soon be applied again.

### Full Context/Explanation
The implementation of the AWS asynchronous connection is currently using the non-public method `get_response` from `botocore`, here is an example:
https://github.com/celery/kombu/blob/3e098dc94ed2a389276ccf3606a0ded3da157d72/kombu/asynchronous/aws/connection.py#L232-L236

`get_response` is vendored in here:
https://github.com/celery/kombu/blob/3e098dc94ed2a389276ccf3606a0ded3da157d72/kombu/asynchronous/aws/ext.py#L1-L9

Botocore recently switched to JSON for the wire protocol it uses to communicate with SQS [here](https://github.com/boto/botocore/commit/3e1f77d8c8eae589cdfb4b9ec09c816570f19904#diff-53c9ed27fbb58a2b8274cc8203f6dbbf167840984a94be3b55df322762f01b5bL6-R15) (Note: Commit url doesn't expand nicely, you'll need to follow the link).

Kombu builds XML AWS Query API requests manually (instead of using something like the boto3 SDK client) to communicate with SQS:
https://github.com/celery/kombu/blob/3e098dc94ed2a389276ccf3606a0ded3da157d72/kombu/asynchronous/aws/connection.py#L185-L203

The resulting XML responses are then passed to the aforementioned `get_response`, which now expects JSON due to the change in Botocore also linked above, for parsing. **In this situation, `get_response` returns a response object with an empty body, which Kombu reads to be a legitimate empty result. This means that a request to get messages, for example, will always return empty. In this way it fails silently.** 

Botocore has reverted the change [here](https://github.com/boto/botocore/commit/47676c3310601dc2b1a1132799e9cc975d386b4a). But this is only a temporary fix, as JSON is the preferred wire protocol.

### Possible Paths Forward
1. Kombu could move to using the public boto3 SDK to interact with SQS (create queues, read messages, etc) instead of hand crafting AWS Query API requests.
   **pros**: Seamless support from a public interface which will handle the wire protocol migration smoothly
   **cons**: This is a large refactoring, and presumably, there is some historical reason that we're unaware of for why Kombu is hand crafting requests since it is a complex piece of code and there was likely a reason for the effort to build it this way.
1. Re-institute the XML parsing that was previously in place before the migration to `get_response` (which was done in [this PR here](https://github.com/celery/kombu/pull/693/files#diff-a33f9d4b5ed32656d1d451c1aed5a1cc2b73969ed6af85df50c2efed6de73cfdL262-R234)).
   **pros**: Smaller scope for the change which could likely be completed sooner. Code worked in the past
   **cons**: More complex handcrafted request/response handling existing in Kombu rather than simply using boto3 SDK
1. Something else entirely? Please let us know what you think!

Thanks in advance for your time :pray:   
 
CC:
- MWAA: @kytha, @rafidka, @dhegberg, @yumike2006
- Airflow: @onikolas, @vincbeck
- Boto: @nateprewitt

rafidka:
# Overview

Context on this can be found in #1726 , but briefly, AWS SQS is planning on a change for making JSON the default communication protocol instead of XML for their boto3 library. Currently, `kombu` manually craft the AWS request, which results in an XML response when it hits SQS, yet uses boto3 to parse the response. So, when the latter changes to JSON, it will break Celery when used with SQS (and, consequently, break Airflow when used with Celery and SQS.)

This PR fixes the issue by using boto3 all the way, instead of just for parsing responses. Thus, when AWS SQS eventually changes the communication protocol for its boto3 client from XML to JSON, the requests will ask for JSON and the response will be parsed as JSON, thus avoiding the problem altogether.

# Testing

To reproduce the issue, I created this simple Celery application:

## worker.py

```python
from celery import Celery

app = Celery('worker', broker='sqs://')

app.config_from_object
app.conf.update(
    broker_transport_options={
        'region': 'us-west-2',
        'visibility_timeout': 300,
        'polling_interval': 1,
        'predefined_queues': {
            'celery': {
                'url': 'https://sqs.us-west-2.amazonaws.com/<account ID>/celery-test',
            }
        }
    },
)


@app.task(acks_late=False)
def add(x, y):
    print(x + y)
    return x + y
```

## scheduler.py

```python
import time

from worker import add

while True:
    print("Scheduling task...")
    add.delay(4, 4)
    time.sleep(3)
```

Then I specifically installed botocore version `1.29.127`, which switched to JSON formatting (before it was rolled back in the next version) and confirmed the issue is happening. After my fix in this PR, the issue disappeared.


auvipy (changes requested):
can you please check the build failures?

rafidka:
> can you please check the build failures?

Yeah, did notice the lint broke. I will fix it. Meanwhile, I would still appreciate feedback on the code. Thanks.

auvipy:
it is not lint broke :) its a test failure =================================== FAILURES ===================================
_________________________ test_Channel.test_get_async __________________________

self = <t.unit.transport.test_SQS.test_Channel object at 0x7f7f958ff610>

    @pytest.mark.usefixtures('hub')
    def test_get_async(self):
        """Basic coverage of async code typically used via:
        basic_consume > _loop1 > _schedule_queue > _get_bulk_async"""
        # Prepare
        for i in range(3):
            message = 'message: %s' % i
            self.producer.publish(message)
    
        # SQS.Channel.asynsqs constructs AsyncSQSConnection using self.sqs
        # which is already a mock thanks to `setup` above, we just need to
        # mock the async-specific methods (as test_AsyncSQSConnection does)
        async_sqs_conn = self.channel.asynsqs(self.queue_name)
        async_sqs_conn.get_list = Mock(name='X.get_list')
    
        # Call key method
        self.channel._get_bulk_async(self.queue_name)
    
>       assert async_sqs_conn.get_list.call_count == 1
E       AssertionError: assert 0 == 1
E        +  where 0 = <Mock name='X.get_list' id='140185630587344'>.call_count
E        +    where <Mock name='X.get_list' id='140185630587344'> = <kombu.asynchronous.aws.sqs.connection.AsyncSQSConnection object at 0x7f7f82b4a090>.get_list

rafidka:
> it is not lint broke :) its a test failure

Ah, ok, sorry I missed that. I ran tests locally, but I had to scope it to some of the files only as I started getting some Azure related errors. @auvipy , what is the easiest way to run all tests locally?

rafidka:
@auvipy , I fixed the build issues. Could you please take a look?

auvipy:
yes, thanks! I am going to review it

auvipy (commented):
the codes look good. can you also check in the docs to see if anything need adjustment? thanks for your effort here

rafidka:
> the codes look good. can you also check in the docs to see if anything need adjustment? thanks for your effort here

Sure, will do.

dhegberg (approved):

Kalhan22 (approved):

rafidka:
> the codes look good. can you also check in the docs to see if anything need adjustment? thanks for your effort here

Hi @auvipy . I checked the documentation under the doc/ folder and I didn't see anything related to the methods I removed so we should be good. I also rebased my branch and fixed a couple of issues the linter was complaining about. Finally, 3 members from my team also reviewed the change and they think the change looks good. Please let me know if there is anything else that I can help with.

One important thing I would like to mention is regarding the methods I removed. I confirmed that those methods are not being used within Celery, but I cannot confirm they are not referenced outside. However, since those methods are not documented, and they are kind of internal methods, they are not supposed to be used by others, and if that's the case, then the using code should be updated accordingly, much like how we are updating kombu to avoid using botocore's internal stuff. Please let me your thoughts on this.

auvipy:
I agree with your view points, I am going to review it and merge when feel confident. thanks for taking on this

auvipy (approved):

auvipy:
Thanks a lot Rafid!

auvipy:
we got a report on performance regression here https://github.com/celery/kombu/issues/1783#issuecomment-1722747499. would be thankful to have this verified

auvipy:
we might need to revert the change https://github.com/celery/kombu/pull/1799

rafidka:
@auvipy , I don't think this is a good idea. Reverting the change is a ticking bomb that will result in kombu being unusable completely when botocore is updated. Performance degradation seems the lesser of the two evils, assuming it is indeed caused by my PR. What is the timeline for the new release? I can try to find sometime to debug the issue to see if I can root cause the performance degradation issue.

auvipy:
> @auvipy , I don't think this is a good idea. Reverting the change is a ticking bomb that will result in kombu being unusable completely when botocore is updated. Performance degradation seems the lesser of the two evils, assuming it is indeed caused by my PR. What is the timeline for the new release? I can try to find sometime to debug the issue to see if I can root cause the performance degradation issue.

reverting is the last resort. you can take your time to re investigate the issue in coming weeks.

rafidka:
> > @auvipy , I don't think this is a good idea. Reverting the change is a ticking bomb that will result in kombu being unusable completely when botocore is updated. Performance degradation seems the lesser of the two evils, assuming it is indeed caused by my PR. What is the timeline for the new release? I can try to find sometime to debug the issue to see if I can root cause the performance degradation issue.
> 
> reverting is the last resort. you can take your time to re investigate the issue in coming weeks.

@auvipy , what is the time frame we are talking about? I have an important deadline on October 18. If that's too far, I will try to find someone else in my team to take a look at this.

auvipy:
That is a little bit too late. We want to cut a release this month, for bug fix stuff. But since it is a big change we can stop reverting for now and wait till 30 Oct, 2023 to find a Solution.

rafidka:
We think that there indeed is an issue with Pull Request, and my colleague and I are actively working on root causing and fixing this issue. We will keep you updated.

rafidka:
@auvipy , I found the root cause behind the performance degradation. Please see my comment [here](https://github.com/celery/kombu/issues/1783#issuecomment-1747739904).

Issue #222, opened 2013-04-04
SQS broker url doesn't include queue_name_prefix


j0hnsmith:
I'm using a AWS IAM to access the SQS queue, the broker url is `sqs://{user_name}@queue.amazonaws.com/{account_id}/{queue_name}`, the `BROKER_TRANSPORT_OPTIONS['queue_name_prefix']` isn't included in the queue name.

I looked in the `fields` param https://github.com/celery/kombu/blob/master/kombu/connection.py#L560 and it the queue name is in virtual host, but with my account id as well (eg 12345/queue_name) so there's no easy way to insert it (I'm not sure if the queue is supposed to be in virtual host or not).

josephclim:
@j0hnsmith were you able to resolve this issue? I just had an application that behaved like this in production but cannot seem to replicate the issue on local. I'm also using SQS and AWS IAM, but using Celery 3.1.18 and Flask deployed on Elastic Beanstalk. Also I don't set the `BROKER_URL` but leave the broker as `sqs://`

Lifang13:
I am having issues to use sqs. http://docs.celeryproject.org/en/latest/getting-started/brokers/sqs.html 
this doc did not say what is host name after @

auvipy:
@Lifang13 what is the current status of this with kombu 4.6.x?

auvipy:
so is it actually working?

shoaib-mohd:
Absolutely Yes

auvipy:
thank you! but should we document that as example? if so can you please contribute?

shoaib-mohd:
Yes, You can document it as an example.
[auvipy](https://github.com/auvipy) please tell me, where should I post it as a contribution?

auvipy:
you should be able to find relevant sections in userguide and reference if needed https://github.com/celery/kombu/tree/main/docs

shoaib-mohd:
Hello @auvipy 
I have created a pull request by updating the docs/userguide/connections.rst file.

PR #1779, opened 2023-08-31
Kombu & celery with SQS #222

shoaib-mohd:
+ celery with sqs #222

SQS broker url doesn't include queue_name_prefix #222
ANSWER:-
I have used the below code for queue_name_prefix...
It is in Flask App running on production..

    from celery import Celery
    def make_celery(app):
        celery = Celery(
            app.import_name,
            broker="sqs://",
            broker_transport_options={
                "queue_name_prefix": "{SERVICE_ENV}-{SERVICE_NAME}-"
            },
        )
        task_base = celery.Task
    
        class ContextTask(task_base):
            abstract = True
    
            def __call__(self, *args, **kwargs):
                with app.app_context():
                    return task_base.__call__(self, *args, **kwargs)
    
        celery.Task = ContextTask
    
        return celery

### DIFF
--- a/docs/userguide/connections.rst
+++ b/docs/userguide/connections.rst
@@ -67,7 +67,7 @@ resources:
 .. _connection-urls:
 
 Celery with SQS
-=============
+=================
 SQS broker url doesn't include queue_name_prefix by default.
 So we can use the following code snippet to make it work in celery.
 .. code-block:: python
--- a/docs/userguide/connections.rst
+++ b/docs/userguide/connections.rst
@@ -67,7 +67,7 @@ resources:
 .. _connection-urls:
 
 Celery with SQS
-=================
+===============
 SQS broker url doesn't include queue_name_prefix by default.
 So we can use the following code snippet to make it work in celery.
 .. code-block:: python
--- a/docs/userguide/connections.rst
+++ b/docs/userguide/connections.rst
@@ -67,7 +67,7 @@ resources:
 .. _connection-urls:
 
 Celery with SQS
-====
+=============
 
 .. code-block:: python
 
--- a/docs/userguide/connections.rst
+++ b/docs/userguide/connections.rst
@@ -68,7 +68,7 @@ resources:
 
 Celery with SQS
 =============
-
+SQS broker url doesn't include queue_name_prefix by default. So we can use the following code snippet to make it work.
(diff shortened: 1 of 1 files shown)


docs/userguide/connections.rst
```diff
@@ -66,6 +66,33 @@ resources:
 
 .. _connection-urls:
 
+Celery with SQS
+====
```
auvipy, line 70:
```suggestion
=============
```

auvipy (changes requested):

docs/userguide/connections.rst
```diff
@@ -66,6 +66,33 @@ resources:
 
 .. _connection-urls:
 
+Celery with SQS
+=============
+
```
auvipy, line 71:
```suggestion
SQS broker url doesn't include queue_name_prefix by default. So we can use the following code snippet to make it work.
```

auvipy (changes requested):
or should we use the docs in celery instead?

shoaib-mohd:
You should use it in both.

docs/userguide/connections.rst
```diff
@@ -66,6 +66,33 @@ resources:
 
 .. _connection-urls:
 
+Celery with SQS
+=============
```
auvipy, line 70:
```suggestion
=================
```

docs/userguide/connections.rst
```diff
@@ -66,6 +66,33 @@ resources:
...
+SQS broker url doesn't include queue_name_prefix by default. So we can use the following code snippet to make it work.
```
auvipy, line 71:
```suggestion
SQS broker url doesn't include queue_name_prefix by default.
So we can use the following code snippet to make it work in celery.
```

auvipy (changes requested):

docs/userguide/connections.rst
```diff
@@ -66,6 +66,34 @@ resources:
 
 .. _connection-urls:
 
+Celery with SQS
+=================
```
auvipy, line 70:
```suggestion
===============
```

auvipy (changes requested):

auvipy (approved):

auvipy:
I think we can hold of for now, but in future we might try

PR #1788, opened 2023-09-12
Add azure service bus auto lock renewal

KonstantinLapkovsky:


auvipy (changes requested):
this seems to be  a new feature. can you please add additional tests to verify this won't regress?

auvipy:
this will need tests coverage as well

PR #1641, opened 2023-01-04
azure service bus: add managed identity support

jasonwbarnett:
This is a work in progress. I still need to add a bunch of unit tests.

edit 1: I spent a bunch of time trying to get the unit tests to work and failed in my efforts. I have been running this code in production since this PR was originally opened (Feb 2023).

Issue #1345, opened 2021-05-26
Authentication with Azure AD against Azure Service Bus


Centzilius:
Hi,
could you please consider implementing an alternative authentication method for Azure Service Bus that is not SAS. 
I'd like to use Azure Managed Identites to authenticate to reduce keeping secrets on my machine. 

A way to implement this would probably look similar to this: 
https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/servicebus/azure-servicebus/samples/sync_samples/client_identity_authentication.py (Replacing EnvironmentCredential with either DefaultAzureCredential or ManagedIdentityCredential from the python package azure.identity). 

The problem with this would probably be, how to identify that you want to authenticate with something else than a SAS but that could maybe be handled by providing specific dummy usernames in the connection string?

What do you think?

auvipy:
no. contributions are welcome

marnikow:
Hi

PR #1641 solved this issue I think.

I don't know if I should open a new issue, but I noticed an inconsistency in the mentioned PR. The new documentation states to use one of these formats:

- azureservicebus://DefaultAzureIdentity@SERVICE_BUSNAMESPACE
- azureservicebus://ManagedIdentityCredential@SERVICE_BUSNAMESPACE

While the code checks for one of these formats: 
- azureservicebus://DefaultAzureCredential@SERVICE_BUSNAMESPACE
- azureservicebus://ManagedIdentityCredential@SERVICE_BUSNAMESPACE

(DefaultAzureIdentity vs DefaultAzureCredential)

marnikow:
It seems there is also another error in the code. When using DefaultAzureCredential azureservicebus.py `as_uri` function tries iterate over DefaultAzureCredential.

`[2023-09-21 15:52:33,161: CRITICAL/MainProcess] Unrecoverable error: TypeError("argument of type 'DefaultAzureCredential' is not iterable")
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/celery/worker/worker.py", line 202, in start
    self.blueprint.start(self)
  File "/usr/local/lib/python3.8/dist-packages/celery/bootsteps.py", line 112, in start
    self.on_start()
  File "/usr/local/lib/python3.8/dist-packages/celery/apps/worker.py", line 135, in on_start
    self.emit_banner()
  File "/usr/local/lib/python3.8/dist-packages/celery/apps/worker.py", line 169, in emit_banner
[...]

PR #1801, opened 2023-10-02
fix azure servicebus using managed identity support

marnikow:
When using managed identities with azure service bus the connection fails as mentioned in #1345 

This PR is a continuation of #1641


auvipy (changes requested):
the tests are looking good. can you also check the lint issues please?

auvipy (approved):

Issue #1800, opened 2023-09-27
kombu redis requeue concurrency bug


jiangxianfu:
problem：

When I have a queue of 3 consumers consuming messages, sometimes I don't get the complete data correctly.

for debug :
i modify file ,add print log  [py3/lib/python3.10/site-packages/kombu/transport/redis.py ]
```
 417     def restore_by_tag(self, tag, client=None, leftmost=False):
 418 
 419         def restore_transaction(pipe):
 420             p = pipe.hget(self.unacked_key, tag)
 421             pipe.multi()
 422             self._remove_from_indices(tag, pipe)
 423             if p:
 424                 M, EX, RK = loads(bytes_to_str(p))  # json is unicode
 425                 self.channel._do_restore_message(M, EX, RK, pipe, leftmost)
 426                 **print('restore tag')**
 427             else:
 428                 **print('un restore tag')**
 429 
 430         with self.channel.conn_or_acquire(client) as client:
 431             client.transaction(restore_transaction, self.unacked_key)

```

Reproduction step:

1. create a queue and enqueue 3 items data

(py3) [jiangxf@xxx ~]$ python test_kombu.py --init
Namespace(init=True, requeue=False, ack=False)
ok 0
ok 1
ok 2

2. simulate two consumer to get item from redis queue, but only requeue data

(py3) [jiangxf@xxx ~]$ python test_kombu.py --requeue
Namespace(init=False, requeue=True, ack=False)
{'test-1': 1695786971.69251}
restore tag
finished-requeue
{'test-2': 1695786971.6956441}
restore tag
finished-requeue
{'formtest-0': 1695785710.66742}
restore tag
finished-requeue
{'formtest-1': 1695785711.2942934}

py3) [jiangxf@xxx ~]$ python test_kombu.py --requeue
Namespace(init=False, requeue=True, ack=False)
{'test-1': 1695787583.5387304}
restore tag
finished-requeue
{'test-0': 1695787583.5360556}
restore tag
finished-requeue
{'test-2': 1695787583.5412734}
restore tag
finished-requeue
{'test-2': 1695787583.5412734}
restore tag
finished-requeue
{'test-0': 1695787583.5360556}
**un restore tag** ---the data is missing
finished-requeue

3. i simulate a normal consumer to get item from redis queue

(py3) [jiangxf@xxx ~]$ python test_kombu.py --ack
Namespace(init=False, requeue=False, ack=True)
{'test-1': 1695787583.5387304}
finished-ack

[...]

PR #1805, opened 2023-10-07
fix: redis requeue concurrency bug  #1800

jiangxianfu:
this bug is ack concurrency caused by del unack key.

for debug i print full stack :

delete a58b8b31-c8c8-4c1e-88d9-f857bbf375a2 unacked_index
  File "/home/jiangxf/py3/lib/python3.10/site-packages/kombu/transport/virtual/base.py", line 639, in _callback
    return callback(message)
  File "/home/jiangxf/py3/lib/python3.10/site-packages/kombu/messaging.py", line 656, in _receive_callback
    return on_m(message) if on_m else self.receive(decoded, message)
  File "/home/jiangxf/py3/lib/python3.10/site-packages/kombu/messaging.py", line 622, in receive
    [callback(body, message) for callback in callbacks]
  File "/home/jiangxf/py3/lib/python3.10/site-packages/kombu/messaging.py", line 622, in <listcomp>
    [callback(body, message) for callback in callbacks]
  File "/home/jiangxf/test_kombu.py", line 22, in message_process_requeue
    message.requeue()
  File "/home/jiangxf/py3/lib/python3.10/site-packages/kombu/message.py", line 187, in requeue
    self.channel.basic_reject(self.delivery_tag, requeue=True)
  File "/home/jiangxf/py3/lib/python3.10/site-packages/kombu/transport/virtual/base.py", line 680, in basic_reject
    self.qos.reject(delivery_tag, requeue=requeue)
  File "/home/jiangxf/py3/lib/python3.10/site-packages/kombu/transport/redis.py", line 385, in reject
    self.ack(delivery_tag)
  File "/home/jiangxf/py3/lib/python3.10/site-packages/kombu/transport/redis.py", line 379, in ack
    self._remove_from_indices(delivery_tag).execute()
  File "/home/jiangxf/py3/lib/python3.10/site-packages/kombu/transport/redis.py", line 397, in _remove_from_indices
    traceback.print_stack(limit=10)

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -382,7 +382,9 @@ class QoS(virtual.QoS):
     def reject(self, delivery_tag, requeue=False):
         if requeue:
             self.restore_by_tag(delivery_tag, leftmost=True)
-        self.ack(delivery_tag)
+        else:
+            self._remove_from_indices(delivery_tag).execute()
+        super().ack(delivery_tag)
 
     @contextmanager
     def pipe_or_acquire(self, pipe=None, client=None):
--- a/t/unit/transport/test_redis.py
+++ b/t/unit/transport/test_redis.py
@@ -1521,9 +1521,15 @@ class test_MultiChannelPoller:
     def test_qos_reject(self):
         p, channel = self.create_get()
         qos = redis.QoS(channel)
-        qos.ack = Mock(name='Qos.ack')
+        qos._remove_from_indices = Mock(name='_remove_from_indices')
         qos.reject(1234)
-        qos.ack.assert_called_with(1234)
+
+    def test_qos_requeue(self):
+        p, channel = self.create_get()
+        qos = redis.QoS(channel)
+        qos.restore_by_tag = Mock(name='restore_by_tag')
+        qos.reject(1234, True)
+        qos.restore_by_tag.assert_called_with(1234, leftmost=True)
 
     def test_get_brpop_qos_allow(self):
         p, channel = self.create_get(queues=['a_queue'])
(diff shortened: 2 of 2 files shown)


kombu/transport/redis.py
```diff
@@ -382,7 +382,9 @@ def ack(self, delivery_tag):
     def reject(self, delivery_tag, requeue=False):
         if requeue:
             self.restore_by_tag(delivery_tag, leftmost=True)
-        self.ack(delivery_tag)
+        else:
```
auvipy, line 385: it would be great if you could add unit test for this change and if possible, integration test which will give us a lot of confidence

auvipy, line 385: it would be great if you could add unit test for this change and if possible, integration test which will give us a lot of confidence

auvipy (changes requested):

kombu/transport/redis.py
jiangxianfu: ok

auvipy:
tests seems to be passing

kbiedrzycki:
I feel like we're getting impacted by the similar issue, but not exactly sure. @auvipy is there a plan to move this forward? What we're exactly observing is a case where there is a message that's being worked on (`acks_late=True`) and a cold shutdown event gets issued. Now, celery from time to time does not restore the message while it is mid-processing. Wondering what might cause that state, but maybe this is partially affected by this bug?

auvipy (changes requested):
can you please pull from main branch to test with python 3.12? we got new python on the ci

auvipy (approved):

Issue #1783, opened 2023-09-04
Slow SQS task processing due to `wait_time_in_seconds` regression v5.3.2


dtorres-sf:
We recently update from Kombu 5.3.1 to 5.3.2 and saw a very dramatically decrease in the number of tasks that can be processed. Here is an example for a single worker processing a full queue - both graphs in number of messages deleted (processed) a minute.

V5.3.1 (about 500 messages a minute):

![image](https://github.com/celery/kombu/assets/30841716/45378103-ade4-458f-beaa-53a0e7510a47)

Here is the same exact worker and messages being processed on v5.3.2 (about 5 messages a minute):

![image](https://github.com/celery/kombu/assets/30841716/4ff80ce6-b44b-4d8b-94c4-660b1659631c)

We traced this down to just the upgrade of Kombu. When diving into detail we found that the throughput limitation on 5.3.2 was solely due to one setting `wait_time_in_seconds`. When we looked into the logs we found that after a message was processed (which only took 100ms) the worker sat idle for 9.9 seconds, before processing the next task, even if the queue is completely full. Sure enough the `wait_time_in_seconds` was set to 10 seconds. On 5.3.1 Kombu correctly realizes there is more on the queue and doesn't wait the `wait_time_in_seconds` to process the next message.

We are using Amazon SQS, celery[sqs] v5.2.7, here are the broker settings:
```
CELERY_BROKER_TRANSPORT_OPTIONS = {
    "region": "us-west-2",
    "visibility_timeout": 14400,
    "polling_interval": 1,
    'wait_time_seconds': 10
}
```

Let me know if there is anything else I can provide.

vifelisberto:
We're experiencing the same issue, our worker was extremely slow after updating and we weren't figuring out what it was.

Until I checked this past issue https://github.com/celery/celery/issues/4995 where I commented on the `kombu` version, I believe the problem has happened again

auvipy:
were you able to track down the commit causing this? can you also try celery v5.3.4 with the latest kombu release, please? and let us know

dtorres-sf:
@auvipy I have tried using celery 5.3.4. The celery upgrade did not fix the issue. I confirmed the issue still exists on 5.3.4.

I have not been able to narrow down the exact change that caused the issue but the suspect PR is this one:

https://github.com/celery/kombu/pull/1759/files

This is purely based off of code review, since a lot of the SQS code path changed as part of this PR.

samueltrevena:
[...]

PR #1807, opened 2023-10-09
Use the correct protocol for SQS requests

rafidka:
TL;DR - The use of boto3 in #1759 resulted in relying on blocking (synchronous) HTTP requests, which caused the performance issue reported in #1783.

`kombu` previously used to craft AWS requests manually as explained in detail in #1726, which resulted in an outage when botocore temporarily changed the default protocol to JSON (before rolling back due to the impact on celery and airflow.) To fix the issue, I submitted #1759, which changes `kombu` to use `boto3` instead of manually crafting AWS requests. This way when boto3 changes the default protocol, kombu won't be impacted.

While working on #1759, I did extensive debugging to understand the multi-threading nature of kombu. What I discovered is that there isn't an actual multi-threading in the true sense of the word, but an event loop that runs on the same thread and process and orchestrate the communication with SQS. As such, it didn't appear to me that there is anything to worry about my change, and the testing I did didn't discover any issue. However, it turns out that while kombu's event loop doesn't have actual multi-threading, its [reliance on pycurl](https://github.com/celery/kombu/blob/main/kombu/asynchronous/http/curl.py#L48) (and thus libcurl) meant that the requests to AWS were being done asynchronously. On the other hand, boto3 requests are always done synchronously, i.e. they are blocking requests.

The above meant that my change introduced blocking on the event loop of kombu. This is fine in most of the cases, since the requests to SQS are pretty fast. However, in the case of using long-polling, a call to SQS's ReceiveMessage can last up to 20 seconds (depending on the user configuration).

To solve this problem, I rolled back my earlier changes and, instead, to address the issue reported in #1726, I now changed the `AsyncSQSConnection` class such that it crafts either a `query` or a `json` request depending on the protocol used by the SQS client. Thus, when botocore changes the default protocol of SQS to JSON, kombu won't be impacted, since it crafts its own request and, after my change, it uses a hard-coded protocol based on the crafted requests.

This solution shouldn't be the final solution, and it is more of a workaround that does the job for now. The final solution should be to completely rely on boto3 for any communication with AWS, and ensuring that all requests are async in nature (non-blocking.) This, however, is a fundamental change that requires a lot of testing, in particular performance testing.

### DIFF
--- a/kombu/asynchronous/aws/connection.py
+++ b/kombu/asynchronous/aws/connection.py
@@ -193,7 +193,7 @@ class AsyncAWSQueryConnection(AsyncConnection):
         param_payload = {'data': params}
         if verb.lower() == 'get':
             # query-based opts
-            signing_type = 'presignurl'
+            signing_type = 'presign-url'
             param_payload = {'params': params}
 
         request = AWSRequest(method=verb, url=path, **param_payload)
--- a/kombu/asynchronous/aws/sqs/connection.py
+++ b/kombu/asynchronous/aws/sqs/connection.py
@@ -4,9 +4,8 @@ from __future__ import annotations
 
 import json
 
-from vine import transform
-
 from botocore.serialize import Serializer
+from vine import transform
 
 from kombu.asynchronous.aws.connection import AsyncAWSQueryConnection
 from kombu.asynchronous.aws.ext import AWSRequest
--- a/kombu/asynchronous/aws/sqs/connection.py
+++ b/kombu/asynchronous/aws/sqs/connection.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import json
-
 from botocore.serialize import Serializer
 from vine import transform
 
--- a/kombu/asynchronous/aws/sqs/connection.py
+++ b/kombu/asynchronous/aws/sqs/connection.py
@@ -2,9 +2,14 @@
 
 from __future__ import annotations
 
+import json
+
 from vine import transform
 
+from botocore.serialize import Serializer
+
 from kombu.asynchronous.aws.connection import AsyncAWSQueryConnection
+from kombu.asynchronous.aws.ext import AWSRequest
 
 from .ext import boto3
 from .message import AsyncMessage
@@ -24,6 +29,84 @@ class AsyncSQSConnection(AsyncAWSQueryConnection):
             region_name=region, debug=debug,
             **kwargs
         )
+
+    def _create_query_request(self, operation, params, queue_url, method):
+        params = params.copy()
+        if operation:
+            params['Action'] = operation
+
+        # defaults for non-get
+        param_payload = {'data': params}
+        if method.lower() == 'get':
+            # query-based opts
+            param_payload = {'params': params}
+
+        return AWSRequest(method=method, url=queue_url, **param_payload)
+
+    def _create_json_request(self, operation, params, queue_url):
+        params = params.copy()
+        params['QueueUrl'] = queue_url
+
+        service_model = self.sqs_connection.meta.service_model
+        operation_model = service_model.operation_model(operation)
+
+        url = self.sqs_connection._endpoint.host
+
+        headers = {}
+        # Content-Type
+        json_version = operation_model.metadata['jsonVersion']
+        content_type = f'application/x-amz-json-{json_version}'
+        headers['Content-Type'] = content_type
+
+        # X-Amz-Target
+        target = '{}.{}'.format(
+            operation_model.metadata['targetPrefix'],
+            operation_model.name,
+        )
+        headers['X-Amz-Target'] = target
+
+        param_payload = {
+            'data': json.dumps(params),
+            'headers': headers
+        }
+
+        method = operation_model.http.get('method', Serializer.DEFAULT_METHOD)
+        return AWSRequest(
+            method=method,
+            url=url,
+            **param_payload
+        )
+
+    def make_request(self, operation_name, params, queue_url, verb, callback=None):  # noqa
+        """
+        Overide make_request to support different protocols.
+
+        botocore is soon going to change the default protocol of communicating
+        with SQS backend from 'query' to 'json', so we need a special
+        implementation of make_request for SQS. More information on this can
+        be found in: https://github.com/celery/kombu/pull/1807.
+        """
+        signer = self.sqs_connection._request_signer
+
+        service_model = self.sqs_connection.meta.service_model
+        protocol = service_model.protocol
+
+        if protocol == 'query':
+            request = self._create_query_request(
+                operation_name, params, queue_url, verb)
+        elif protocol == 'json':
+            request = self._create_json_request(
+                operation_name, params, queue_url)
+        else:
+            raise Exception(f'Unsupported protocol: {protocol}.')
+
+        signing_type = 'presign-url' if request.method.lower() == 'get' \
+            else 'standard'
+
+        signer.sign(operation_name, request, signing_type=signing_type)
+        prepared_request = request.prepare()
+
+        return self._mexe(prepared_request, callback=callback)
 
     def create_queue(self, queue_name,
                      visibility_timeout=None, callback=None):
--- a/kombu/asynchronous/aws/sqs/connection.py
+++ b/kombu/asynchronous/aws/sqs/connection.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+
 from botocore.serialize import Serializer
 from vine import transform
 
--- a/t/unit/asynchronous/aws/sqs/test_connection.py
+++ b/t/unit/asynchronous/aws/sqs/test_connection.py
@@ -1,11 +1,9 @@
 from __future__ import annotations
 
 import json
-
 from unittest.mock import MagicMock, Mock
 
-from kombu.asynchronous.aws.ext import AWSRequest
-from kombu.asynchronous.aws.ext import boto3
+from kombu.asynchronous.aws.ext import AWSRequest, boto3
 from kombu.asynchronous.aws.sqs.connection import AsyncSQSConnection
 from kombu.asynchronous.aws.sqs.message import AsyncMessage
 from kombu.asynchronous.aws.sqs.queue import AsyncQueue
@@ -13,7 +11,6 @@ from kombu.utils.uuid import uuid
 from t.mocks import PromiseMock
 
 from ..case import AWSCase
-
 
 SQS_URL = 'https://sqs.us-west-2.amazonaws.com/'
 
--- a/t/unit/asynchronous/aws/sqs/test_connection.py
+++ b/t/unit/asynchronous/aws/sqs/test_connection.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
+import json
+
 from unittest.mock import MagicMock, Mock
 
+from kombu.asynchronous.aws.ext import AWSRequest
 from kombu.asynchronous.aws.ext import boto3
 from kombu.asynchronous.aws.sqs.connection import AsyncSQSConnection
 from kombu.asynchronous.aws.sqs.message import AsyncMessage
@@ -10,6 +13,9 @@ from kombu.utils.uuid import uuid
 from t.mocks import PromiseMock
 
 from ..case import AWSCase
+
+
+SQS_URL = 'https://sqs.us-west-2.amazonaws.com/'
 
 
 class test_AsyncSQSConnection(AWSCase):
@@ -30,6 +36,141 @@ class test_AsyncSQSConnection(AWSCase):
         sqs_client.get_queue_url = MagicMock(return_value={
             'QueueUrl': 'http://aws.com'
         })
+
+    def MockRequest(self):
+        return AWSRequest(
+            method='POST',
+            url='https://aws.com',
+        )
+
+    def MockOperationModel(self, operation_name, method):
+        mock = MagicMock()
+        mock.configure_mock(
+            http=MagicMock(
+                get=MagicMock(
+                    return_value=method,
+                )
+            ),
+            name=operation_name,
+            metadata={
+                'jsonVersion': '1.0',
+                'targetPrefix': 'sqs',
+            }
+        )
+        return mock
+
+    def MockServiceModel(self, operation_name, method):
+        service_model = MagicMock()
+        service_model.protocol = 'json',
+        service_model.operation_model = MagicMock(
+            return_value=self.MockOperationModel(operation_name, method)
+        )
+        return service_model
+
+    def assert_requests_equal(self, req1, req2):
+        assert req1.url == req2.url
+        assert req1.method == req2.method
+        assert req1.data == req2.data
+        assert req1.params == req2.params
+        assert dict(req1.headers) == dict(req2.headers)
+
+    def test_create_query_request(self):
+        operation_name = 'ReceiveMessage',
+        params = {
+            'MaxNumberOfMessages': 10,
+            'AttributeName.1': 'ApproximateReceiveCount',
+            'WaitTimeSeconds': 20
+        }
+        queue_url = f'{SQS_URL}/123456789012/celery-test'
+        verb = 'POST'
+        req = self.x._create_query_request(operation_name, params, queue_url,
+                                           verb)
+        self.assert_requests_equal(req, AWSRequest(
+            url=queue_url,
+            method=verb,
+            data={
+                'Action': (operation_name),
+                **params
+            },
+            headers={},
+        ))
+
+    def test_create_json_request(self):
+        operation_name = 'ReceiveMessage'
+        method = 'POST'
+        params = {
+            'MaxNumberOfMessages': 10,
+            'AttributeName.1': 'ApproximateReceiveCount',
+            'WaitTimeSeconds': 20
+        }
+        queue_url = f'{SQS_URL}/123456789012/celery-test'
+
+        self.x.sqs_connection = Mock()
+        self.x.sqs_connection._request_signer = Mock()
+        self.x.sqs_connection._endpoint.host = SQS_URL
+        self.x.sqs_connection.meta.service_model = Mock()
+        self.x.sqs_connection.meta.service_model.protocol = 'json',
+        self.x.sqs_connection.meta.service_model.operation_model = MagicMock(
+            return_value=self.MockOperationModel(operation_name, method)
+        )
+
+        req = self.x._create_json_request(operation_name, params, queue_url)
+        self.assert_requests_equal(req, AWSRequest(
+            url=SQS_URL,
+            method=method,
+            data=json.dumps({
+                **params,
+                "QueueUrl": queue_url
+            }),
+            headers={
+                'Content-Type': 'application/x-amz-json-1.0',
+                'X-Amz-Target': f'sqs.{operation_name}'
+            },
+        ))
+
+    def test_make_request__with_query_protocol(self):
+        # Do the necessary mocking.
+        self.x.sqs_connection = Mock()
+        self.x.sqs_connection._request_signer = Mock()
+        self.x.sqs_connection.meta.service_model.protocol = 'query'
+        self.x._create_query_request = Mock(return_value=self.MockRequest())
+
+        # Execute the make_request called and confirm we are creating a
+        # query request.
+        operation = 'ReceiveMessage',
+        params = {
+            'MaxNumberOfMessages': 10,
+            'AttributeName.1': 'ApproximateReceiveCount',
+            'WaitTimeSeconds': 20
+        }
+        queue_url = f'{SQS_URL}/123456789012/celery-test'
+        verb = 'POST'
+        self.x.make_request(operation, params, queue_url, verb)
+        self.x._create_query_request.assert_called_with(
+            operation, params, queue_url, verb
+        )
+
+    def test_make_request__with_json_protocol(self):
+        # Do the necessary mocking.
+        self.x.sqs_connection = Mock()
+        self.x.sqs_connection._request_signer = Mock()
+        self.x.sqs_connection.meta.service_model.protocol = 'json'
+        self.x._create_json_request = Mock(return_value=self.MockRequest())
+
+        # Execute the make_request called and confirm we are creating a
+        # query request.
+        operation = 'ReceiveMessage',
+        params = {
+            'MaxNumberOfMessages': 10,
+            'AttributeName.1': 'ApproximateReceiveCount',
+            'WaitTimeSeconds': 20
+        }
+        queue_url = f'{SQS_URL}/123456789012/celery-test'
+        verb = 'POST'
+        self.x.make_request(operation, params, queue_url, verb)
+        self.x._create_json_request.assert_called_with(
+            operation, params, queue_url
+        )
 
     def test_create_queue(self):
         self.x.create_queue('foo', callback=self.callback)


auvipy:
I think we better revert the change first then we do other stuff. that is the safest thing for short term

auvipy (changes requested):
you can rebase this now for further review and making the the diff smaller for better review. we have to be careful to merge this, so let us take some time. thanks for your continuous effort

auvipy:
well this will take some time to merge this PR. so you can work on other issues as well :)

rafidka:
I managed to fix the issue and confirmed that my updated code is working with both query and JSON protocol. However, I am still marking it as WIP as I don't want it to get merged before getting approvals from the following:

- 2 other senior members from my team (Amazon MWAA)
- 1 member from the SQS team
- 1 member from the botocore team

rafidka:
@auvipy , it is a bit of a pain to get all the tests to run successfully, and I keep getting some weird formatting issue. Is there a way to run the tests locally before updating the PR?

Nusnus:
> @auvipy , it is a bit of a pain to get all the tests to run successfully, and I keep getting some weird formatting issue. Is there a way to run the tests locally before updating the PR?

Here you go @rafidka 

### Unit
```
tox -e 3.11-unit
```
#### Also possible:
* Make sure there's no `.python-version` file / no pyenv environment is used 
```
make docker-unit-tests
```
### Integration
```
tox -e 3.11-integration
```
#### Also possible:
```
docker run -d -p 6379:6379 --name redis redis:latest
docker run -d -p 5672:5672 -p 15672:15672 --name rabbitmq rabbitmq:management
docker start rabbitmq redis
tox -e 3.11-integration-rabbitmq_redis
docker stop rabbitmq redis
```
Using `rabbitmq:management` for debugging @ http://localhost:15672/#/
### Lint (pre-commit)
```
tox -e lint
```

kombu/asynchronous/aws/connection.py
```diff
@@ -229,9 +229,11 @@ def get_status(self, operation, params, path='/', parent=None, verb='GET', callb
 
     def _on_list_ready(self, parent, markers, operation, response):
         service_model = self.sqs_connection.meta.service_model
+        protocol = service_model.protocol
```
eddy-aws, line 232: Could this be a class level static variable?

kombu/asynchronous/aws/ext.py
```diff
@@ -18,9 +18,50 @@ class BotoCoreError(Exception):
...
+def get_response(operation_model, http_response, protocol):
+    """Parses the HTTP response of an AWS operation.
+
+    This is an almost identical copy of botocore's get_response:
+
+    https://github.com/boto/botocore/blob/1.29.126/botocore/response.py#L180-L201
+
+    with the exception that instead of hard-coding the protocol to 'query',
+    we expect the caller to specify the protocol. This helps us deal with both
+    jquery and json protocols. For more context on this, the reader is refered
```
eddy-aws, line 37: what is jquery?

kombu/asynchronous/aws/connection.py
rafidka: No, since the `self.sqs_connection` itself is an instance variable rather than a static variable, We could make the `protocol` an instance variable as well, but there isn't much advantage, and I would rather not since we can easily access the protocol from the service model.

kombu/asynchronous/aws/ext.py
```diff
@@ -18,9 +18,50 @@ class BotoCoreError(Exception):
...
 
 __all__ = (
     'exceptions', 'AWSRequest', 'get_response'
 )
+
+
+def get_response(operation_model, http_response, protocol):
+    """Parses the HTTP response of an AWS operation.
+
+    This is an almost identical copy of botocore's get_response:
```
dhegberg, line 31: Can the botocore implementation be aliased and operation_model.metadata['protocol'] be set to the protocol parameter?

It would be safer if the botocore implementation was used inside this function to prevent drift.

kombu/asynchronous/aws/sqs/connection.py
```diff
@@ -25,6 +29,75 @@ def __init__(self, sqs_connection, debug=0, region=None, **kwargs):
...
+        if operation:
+            params['Action'] = operation
+
+        # defaults for non-get
+        param_payload = {'data': params}
+        if verb.lower() == 'get':
+            # query-based opts
+            param_payload = {'params': params}
+
+        return AWSRequest(method=verb, url=queue_url, **param_payload)
```
dhegberg, line 43: Can we keep `verb` consistent with `method` used elsewhere?

kombu/asynchronous/aws/sqs/connection.py
```diff
@@ -25,6 +29,75 @@ def __init__(self, sqs_connection, debug=0, region=None, **kwargs):
...
+
+        if protocol == 'query':
+            request = self._create_query_request(
+                operation, params, queue_url, verb)
+        elif protocol == 'json':
+            request = self._create_json_request(operation, params, queue_url)
+        else:
+            raise Exception(f'Unsupported protocol: {protocol}.')
+
+        signing_type = 'presignurl' if request.method.lower() == 'get' \
```
dhegberg, line 93: Why use `presignurl` on `get`?  It's not clear from https://github.com/boto/botocore/blob/develop/botocore/signers.py when pre-signed urls should be used.  Can you provide some documentation?

kombu/asynchronous/aws/sqs/connection.py
rafidka: @dhegberg , I tried to stay consistent with the original implementation to reduce the amount of changes. Having said that, yeah, it is confusing, so I will just rename everything to method.

kombu/asynchronous/aws/sqs/connection.py
rafidka: It also confused me, and the only explanation I have is that the documentation [here](https://github.com/boto/botocore/blob/develop/botocore/signers.py#L132) is inaccurate and they meant to instead say:

> This should be used when pre-signing a GET* request"

Having said that, I believe this code (which I copied from the [original implementation of `make_request`](https://github.com/celery/kombu/blob/main/kombu/asynchronous/aws/connection.py#L194-L196) in the `kombu.aws.connection.AsyncAWSQueryConnection` class) is actually incorrect, since the signer name is `presign-url` rather than `presignurl`, but I was planning on keeping it as is until I get feedback from the botocore team.

kombu/asynchronous/aws/ext.py
rafidka: Actually, we don't need this method anymore, and I will just remove it. Initially, I was trying to have `get_response` always parse the response as `query`. This, however, didn't work, and I decided to craft the request to match the format the operation is using, and hence we don't need our version of `get_response` and I will accordingly rename the PR.

rafidka:
Latest update:

- I got feedback from a senior member of MWAA team and a senior member of SQS team.
- I am still waiting for feedback from 2 others, plus feedback from the kombu team.
- I am currently working on performance testing, and will generate some reports and graphs to reduce the likelihood of introducing another performance degradation.
- Until then, I am still marking the PR as WIP to avoid accidentally merging it before all the items above are finished.

auvipy (changes requested):
after all process is complete, we will need the test coverage as well.

kombu/asynchronous/aws/sqs/connection.py
```diff
@@ -25,6 +29,76 @@ def __init__(self, sqs_connection, debug=0, region=None, **kwargs):
...
+        json_version = operation_model.metadata['jsonVersion']
+        content_type = f'application/x-amz-json-{json_version}'
+        headers['Content-Type'] = content_type
+
+        # X-Amz-Target
+        target = '{}.{}'.format(
+            operation_model.metadata['targetPrefix'],
+            operation_model.name,
+        )
+        headers['X-Amz-Target'] = target
```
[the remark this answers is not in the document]
rafidka: Which document are you referring to?

What I did was to check what botocore is doing in their [JSONSerialiazer](https://github.com/boto/botocore/blob/develop/botocore/serialize.py#L347-L371), and I don't see those fields in there.

kombu/asynchronous/aws/sqs/connection.py
```diff
@@ -25,6 +29,76 @@ def __init__(self, sqs_connection, debug=0, region=None, **kwargs):
...
+            url=url,
+            **param_payload
+        )
+
+    def make_request(self, operation_name, params, queue_url, verb, callback=None):  # noqa
+        signer = self.sqs_connection._request_signer
+
+        service_model = self.sqs_connection.meta.service_model
+        protocol = service_model.protocol
+
```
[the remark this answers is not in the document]
rafidka: Totally. I will do that.

rafidka:
> after all process is complete, we will need the test coverage as well.

@auvipy Can you explain what kind of coverage you are looking for?

kombu/asynchronous/aws/sqs/connection.py
```diff
@@ -25,6 +29,76 @@ def __init__(self, sqs_connection, debug=0, region=None, **kwargs):
...
+        url = self.sqs_connection._endpoint.host
+
+        headers = {}
+        # Content-Type
+        json_version = operation_model.metadata['jsonVersion']
+        content_type = f'application/x-amz-json-{json_version}'
+        headers['Content-Type'] = content_type
+
+        # X-Amz-Target
+        target = '{}.{}'.format(
```
auvipy, line 61: can we try f string here? or there are some version restrictions?

auvipy (changes requested):
additional unit test coverage for the new code. and if possible Integration tests

kombu/asynchronous/aws/sqs/connection.py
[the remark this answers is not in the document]
rafidka: Added.

kombu/asynchronous/aws/sqs/connection.py
rafidka: To be honest, I think this is cleaner, since the placeholders would be long, making it hard to read:

```
f'{operation_model.metadata['targetPrefix']}.{operation_model.name}'
```

especially since that it would make the line longer than 80 chars, meaning that we will need to split the line, or otherwise define variables to make it shorter.

Let me know if you feel strongly about it.

rafidka:
While at this, does Celery/Kombu have automated performance tests? In my opinion, this is important. Had Kombu had some performance tests, we wouldn't have missed the performance degradation issue following my previous PR.

rafidka:
@auvipy , the build is failing with the following error for Python 3.12:

```
=========================== short test summary info ============================
ERROR t/integration/test_kafka.py::test_KafkaBasicFunctionality::test_connect
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
======================== 18 warnings, 1 error in 1.01s =========================
3.12-linux-integration-py-kafka: exit 1 (1.43 seconds) /home/runner/work/kombu/kombu> pytest -xv -E kafka t/integration -v pid=6605
3.12-linux-integration-py-kafka: docker> remove '186dc5efa430' (from 'zookeeper')
3.12-linux-integration-py-kafka: docker> remove 'dcbc798a422b' (from 'kafka')
.pkg: _exit> python /opt/hostedtoolcache/Python/3.12.0/x64/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: exit None (0.00 seconds) /home/runner/work/kombu/kombu> python /opt/hostedtoolcache/Python/3.12.0/x64/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__ pid=5315
  3.12-linux-integration-py-kafka: FAIL code 1 (62.26=setup[60.83]+cmd[1.43] seconds)
  evaluation failed :( (62.42 seconds)
```

This doesn't seem related to my changes, so I presume Python 3.12 support is still having issues?

Please advise.

auvipy:
You are right. I fixed most of the Python 3.12 test failures but the kafka driver seems not to support python 3.12 as of yet. You can ignore this for now.

kombu/asynchronous/aws/sqs/connection.py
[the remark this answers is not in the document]
o-nikolas: I'll find the docs, I think they were in a slack thread that I will share with you.

rafidka:
@auvipy , just wanted to follow up on this. Is there anything else missing in this PR, or we are good to merge?

auvipy:
I will consider this for only in kombu 5.4.0. as we reverted the original PR. this kind of big refactor should be in a newer feature release only from now on.

rafidka:
> I will consider this for only in kombu 5.4.0. as we reverted the original PR. this kind of big refactor should be in a newer feature release only from now on.

Thanks, Asif. Is there any ETA yet for v5.4.0 release?

ziyanli-amazon:
Do we have an update on this?

rafidka:
@auvipy FYI - https://aws.amazon.com/blogs/aws/new-for-amazon-sqs-update-the-aws-sdk-to-reduce-latency/.

auvipy:
Will check Saturday morning.

auvipy:
got a report that new version of botocore is breaing SQS https://github.com/celery/celery/discussions/8627

rafidka:
> got a report that new version of botocore is breaing SQS [celery/celery#8627](https://github.com/celery/celery/discussions/8627)

Yes, that's due to what is mentioned in the link I shared above:  https://aws.amazon.com/blogs/aws/new-for-amazon-sqs-update-the-aws-sdk-to-reduce-latency/. Customers will need to avoid upgrading their botocore at this point. The other solution is for this PR to get merged, and a new version of kombu is released.

auvipy:
that is on roadmap as we should not introduce breaking change in a patch release. thanks for confirming

rafidka:
> that is on roadmap as we should not introduce breaking change in a patch release

The change in this PR is not breaking. It is backward compatible.

auvipy:
but the overall new feature of SQS is.

rafidka:
> but the overall new feature of SQS is.

I am not sure I follow. Are you concerned about the underlying protocol changing for customers? Or something else?

auvipy:
> Are you concerned about the underlying protocol changing for customers?

yes, we are getting lot of reports regarding that. btw can you plz rebase it again?

auvipy:
I am considering this and few other PR to include in a bug fix release tomorrow or day after tomorrow

rafidka:
> I am considering this and few other PR to include in a bug fix release tomorrow or day after tomorrow. should we pin the boto versions with this patch as well as a safety measure?

No, I don’t think we should pin the botocore versions. The whole point of this PR is so we don’t have to pin the version, and make sure kombu works with any botocore version. If we plan to pin the version of botocore, then we should probably not merge this PR.

In my humble opinion, what we need is to for the open source community to do as much testing as possible for this PR, and then merge. I did a lot of testing internally, but my test will probably not be as thorough as someone from the Celery/Kombu team, since my exposure to Celery/Kombu is relatively limited.

rafidka:
> should we pin the boto versions with this patch as well as a safety measure?

As mentioned above, I don't recommend pinning the boto version, if by pinning you mean a specific version. However, I do recommend limiting the range of supported botocore versions, since Celery is no more compatible with the new botocore.

nickdjones:
I know this isn't a bulletproof test as it's not using real SQS but I've managed to replicate this in our local Docker Compose stack using this image https://hub.docker.com/r/softwaremill/elasticmq-native in place for SQS. We have a Celery container and a separate container for Beat so it is easy to see when one is sending tasks to the other. For all tests the Celery version is 5.3.1 because Poetry complained about this PR being versioned as kombu 5.3.2 which is lower than the requirements for 5.3.5 but no testing has showed main Celery to be the issue so I don't think that matters:

boto3 1.28.80 + botocore 1.31.80 + kombu 5.3.3: Events processed successfully
boto3 1.28.84 + botocore 1.31.84 + kombu 5.3.3: Celery beat sending events, Celery not picking them up
boto3 1.28.84 + botocore 1.31.84 + kombu pinned to this PR branch: Events processed successfully

Not sure how much this helps everyone as it isn't actually using SQS, it is using the latest botocore though and has been incredibly useful when we were debugging what version we should pin our environment to.

kombu/asynchronous/aws/sqs/connection.py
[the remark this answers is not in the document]
auvipy: > I'll find the docs, I think they were in a slack thread that I will share with you.

would you mind doing a last round of review?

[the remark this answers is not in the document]
rafidka: > > I'll find the docs, I think they were in a slack thread that I will share with you.
> 
> would you mind doing a last round of review?

@auvipy , we discussed this actually internally and there is nothing to worry about here. I simply followed what the JSONSerializer class from botocore [is doing](https://github.com/boto/botocore/blob/develop/botocore/serialize.py#L347-L371).

kombu/asynchronous/aws/sqs/connection.py
```diff
@@ -2,9 +2,13 @@
 
 from __future__ import annotations
 
+import json
```
auvipy, line 5:
```suggestion
import json
```

auvipy (changes requested):

kombu/asynchronous/aws/sqs/connection.py
```diff
@@ -2,9 +2,13 @@
...
+
```
auvipy, line 6:
```suggestion
```

auvipy (changes requested):

auvipy (approved):

rafidka:
@auvipy , thanks for merging this PR. I am just curious to learn about performance tests that Celery currently does as part of the release process.

rafidka:
I am asking because I felt with both my PRs (this one and https://github.com/celery/kombu/pull/1759) I didn't get the impression that the Celery team did some performance testing, and it was mostly my testing, which makes me slightly worried as I cannot possibly cover the different cases of Celery, and my testing is mostly focused on Airflow/MWAA.

auvipy:
@rafidka thanks for raising the concern/question. @Nusnus is planning/doing something similar here https://github.com/celery/pytest-celery which is work in progress. the previous stress test tools are out off date now so we can't do much right now. but hopefully that is going to change soon

rafidka:
@auvipy , that's great to hear. @Nusnus , is there any document for what is planned to be covered in that package? I can probably see if I can allocate some time to help with SQS-related test coverage.

Nusnus:
I have been under this internal struggle for a long time now between progressing with the project and communicating my plans. So far, I have chosen 100% focus on progress to get pytest-celery v1.0.0 out as soon as I possibly can.

My MVP strategy goes like this:
- Develop MVP infra for developing new Celery Smoke Tests
- Develop smoke tests  <-- WE ARE HERE 
- Missing an infra feature for _this_ test?
- Implement/refactor/fix infra feature
- Repeat until the first batch of smoke tests are implemented.

This way, I lead the MVP feature list based on actual tests I am developing in parallel, which helps me define the project scope on the fly TDD style. I have a `smoke_tests` branch in Celery that I am working on for that.

> - Repeat until the first batch of smoke tests are implemented.

This means I have a list of tests I plan to implement that once done and passing, will allow me to binge-doc the entire project end-to-end and start moving into beta/RC stages etc.

@rafidka - This could be an excellent time to collaborate on possibly adding SQS support to `pytest-celery`. The current feature list only provides automatic built-in support to RabbitMQ (broker), Redis (broker/backend), and manual built-in support (due to the instability of the feature) to Memcached (backend). If we can implement a stable SQS broker that could be huge, but we'll have to be patient as the API design is still at alpha stage and is due to possible changes still.

Timeline-wise, I hope to start _binge-docing_ in about a month, hopefully less.
The gaps in my plan are based on possible bugs I might find during the smoke tests (some I'll fix, some I'll mark).

The end game is having Celery v5.4 level up in terms of QA for production environments.

Nusnus:
> @rafidka - This could be an excellent time to collaborate on possibly adding SQS support to `pytest-celery`. The current feature list only provides automatic built-in support to RabbitMQ (broker), Redis (broker/backend), and manual built-in support (due to the instability of the feature) to Memcached (backend). If we can implement a stable SQS broker that could be huge, but we'll have to be patient as the API design is still at alpha stage and is due to possible changes still.

If you're up for it, I can contact you when the time comes (let's say Dec/Jan) to collaborate together on it, could be fun and interesting!

Nusnus:
P.S I am currently working on about 5 active branches in both Celery and Pytest Celery, for the infra, the smoke tests, fixes for bugs I found (waiting for merge for v5.4), and more... It's a git rollercoaster, I have to admit haha 😄
But at least I am having fun hahah 😜

auvipy:
> P.S I am currently working on about 5 active branches in both Celery and Pytest Celery, for the infra, the smoke tests, fixes for bugs I found (waiting for merge for v5.4), and more... It's a git rollercoaster, I have to admit haha 😄 But at least I am having fun hahah 😜

Just don't overwhelm yourself with 5.4 stuffs now. you will reach the goals gradually and eventually.  And don't forget to take breaks to avoid possible burnouts. Celery need you for long time! so beside celery, don't forget to take care of yourselves too.

rafidka:
> > @rafidka - This could be an excellent time to collaborate on possibly adding SQS support to `pytest-celery`. The current feature list only provides automatic built-in support to RabbitMQ (broker), Redis (broker/backend), and manual built-in support (due to the instability of the feature) to Memcached (backend). If we can implement a stable SQS broker that could be huge, but we'll have to be patient as the API design is still at alpha stage and is due to possible changes still.
> 
> If you're up for it, I can contact you when the time comes (let's say Dec/Jan) to collaborate together on it, could be fun and interesting!

Thank you, @Nusnus . I discussed with my manager and we will try to fit some time for this into our roadmap, though unlike for December, but good chance next year. I think what will help is when you have a document where you jot down your ideas of what you will be working on. Then I could also add some ideas for SQS-related stuff in Celery (we can only promise help on this front unfortunately).

Nusnus:
> > > @rafidka - This could be an excellent time to collaborate on possibly adding SQS support to `pytest-celery`. The current feature list only provides automatic built-in support to RabbitMQ (broker), Redis (broker/backend), and manual built-in support (due to the instability of the feature) to Memcached (backend). If we can implement a stable SQS broker that could be huge, but we'll have to be patient as the API design is still at alpha stage and is due to possible changes still.
> > 
> > 
> > If you're up for it, I can contact you when the time comes (let's say Dec/Jan) to collaborate together on it, could be fun and interesting!
> 
> Thank you, @Nusnus . I discussed with my manager and we will try to fit some time for this into our roadmap, though unlike for December, but good chance next year. I think what will help is when you have a document where you jot down your ideas of what you will be working on. Then I could also add some ideas for SQS-related stuff in Celery (we can only promise help on this front unfortunately).

Sounds great !
Thank you!

P.S
@auvipy 🫶

Nusnus:
> > > @rafidka - This could be an excellent time to collaborate on possibly adding SQS support to `pytest-celery`. The current feature list only provides automatic built-in support to RabbitMQ (broker), Redis (broker/backend), and manual built-in support (due to the instability of the feature) to Memcached (backend). If we can implement a stable SQS broker that could be huge, but we'll have to be patient as the API design is still at alpha stage and is due to possible changes still.
> > 
> > 
> > If you're up for it, I can contact you when the time comes (let's say Dec/Jan) to collaborate together on it, could be fun and interesting!
> 
> Thank you, @Nusnus . I discussed with my manager and we will try to fit some time for this into our roadmap, though unlike for December, but good chance next year. I think what will help is when you have a document where you jot down your ideas of what you will be working on. Then I could also add some ideas for SQS-related stuff in Celery (we can only promise help on this front unfortunately).

@rafidka 
I’ve just released the [first beta version](https://github.com/celery/pytest-celery/releases/tag/v1.0.0b1) of the new plugin.
Can be seen in action in the new celery [smoke tests](https://github.com/celery/celery/tree/main/t/smoke) in Celery.

v1.0.0 is pretty much close feature-wise. Bug fixing & Documentation are my highest priorities right now alongside supporting the newly merged test suites, in preparation for Celery v5.4 release.
If your roadmap permits allocating time, we can discuss further support for SQS.

P.S
This is a silent release so there are no official announcements.

PR #1811, opened 2023-10-12
Create a lock on cached_property if not present

### LINKED ISSUE #1804: Assigning to cached_property is broken on Python 3.12
This code worked fine before:

```python
$ pip freeze | egrep -i 'celery|kombu'
celery==5.3.4
kombu==5.3.2

$ python
Python 3.12.0 (main, Oct  5 2023, 13:07:56) [GCC 13.2.1 20230826] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from celery import Celery
>>> from celery import Celery
>>> c = Celery()
>>> class MyTask(c.Task): pass
...
>>> c.Task = MyTask
  File "/home/adrian/dev/indico/py3/env-312/lib/python3.12/site-packages/kombu/utils/objects.py", line 37, in __set__
    with self.lock:
         ^^^^^^^^^
AttributeError: 'cached_property' object has no attribute 'lock'
```

It likely  broke because of this change in 3.12:

> The undocumented locking behavior of [cached_property()](https://docs.python.org/3/library/functools.html#functools.cached_property) is removed, because it locked across all instances of the class, leading to high lock contention. This means that a cached property getter function could now run more than once for a single instance, if two threads race. For most simple cached properties (e.g. those that are idempotent and simply calculate a value based on other attributes of the instance) this will be fine. If synchronization is needed, implement locking within the cached property getter function or around multi-threaded access points.

tari:
This fixes #1804 (fixing breakage caused by use of undocumented implementation details of functools.cached_property) by ensuring a lock is always present on cached_property attributes, which is required to safely support setting and deleting cached values in addition to computing them on demand.

### DIFF
--- a/kombu/utils/objects.py
+++ b/kombu/utils/objects.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from threading import RLock
+
 __all__ = ('cached_property',)
 
 try:
@@ -25,10 +27,17 @@ class cached_property(_cached_property):
             # This is a backport so we set this ourselves.
             self.attrname = self.func.__name__
 
+        if not hasattr(self, 'lock'):
+            # Prior to Python 3.12, functools.cached_property has an
+            # undocumented lock which is required for thread-safe __set__
+            # and __delete__. Create one if it isn't already present.
+            self.lock = RLock()
+
     def __get__(self, instance, owner=None):
         # TODO: Remove this after we drop support for Python<3.8
         #  or fix the signature in the cached_property package
-        return super().__get__(instance, owner)
+        with self.lock:
+            return super().__get__(instance, owner)
 
     def __set__(self, instance, value):
         if instance is None:
--- a/t/unit/utils/test_objects.py
+++ b/t/unit/utils/test_objects.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from unittest import mock
+
 from kombu.utils.objects import cached_property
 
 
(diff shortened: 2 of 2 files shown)


auvipy (commented):
thanks for the patch. do you think the change should be covered with some unit tests at least?

auvipy (changes requested):
We are hit by a test failure with python 3.12 here https://github.com/celery/kombu/pull/1812#issuecomment-1759039111, was thinking if this change going to fix that? or we need to update the test as well?

kombu/utils/objects.py
```diff
@@ -25,10 +27,17 @@ def __init__(self, fget=None, fset=None, fdel=None):
...
+            # undocumented lock which is required for thread-safe __set__
+            # and __delete__. Create one if it isn't already present.
+            self.lock = RLock()
+
     def __get__(self, instance, owner=None):
         # TODO: Remove this after we drop support for Python<3.8
         #  or fix the signature in the cached_property package
-        return super().__get__(instance, owner)
+        with self.lock:
+            return super().__get__(instance, owner)
```
ThiefMaster, line 40: Doesn't this now prevent concurrent reads of the cached value? I think it's only necessary to lock during writes, isn't it?

tari: Reads need to synchronize with writes, otherwise a write could be lost if a read found no value present but the write occurred before the read computed and stored the new value. If the presence check and computation were here rather in the superclass only locking after the initial read might be valid, but what I've written here matches the standard library behavior before Python 3.12.

ThiefMaster: ok, makes sense, especially if stdlib did the same

tari:
> thanks for the patch. do you think the change should be covered with some unit tests at least?

Yes, I should be able to add a test to verify that locks are held as needed.

> We are hit by a test failure with python 3.12 here [#1812 (comment)](https://github.com/celery/kombu/pull/1812#issuecomment-1759039111), was thinking if this change going to fix that? or we need to update the test as well?

That looks like a different issue, though I didn't look at the test very closely.

tari:
> thanks for the patch. do you think the change should be covered with some unit tests at least?

Added a test that verifies the lock is acquired for all of get, set, and delete operations. I could also test that the lock is held at appropriate times, but opted not to do that since it would require doing a threading dance in the test to verify; just ensuring that the operations acquire the lock seems fine.

auvipy (approved):

Issue #1493, opened 2022-02-20
support self signed certificates


vogt31337:
Hi,

I'm using pyamqp maybe this is also valid / interesting for other transport layers.
Since version 4.6.8 kombu doesn't accept self signed certificates (at least for amqps connections), which is problematic at least for me...
If I understood issues #1149 and #1151 correctly there is some way to configure the ssl behaviour.
Maybe you could update the documentation to explain how to enable self signed certificates.

BR

vogt31337:
Since about a month has passed, I'd like to ask if there was something done?
@thedrow Maybe you could provide some quick tips?

BR

davidjrice:
Hey @vogt31337 ran into this issue trying to get setup with SSL myself. Posting for posterity in case any else does too  :)

These `celery` configuration options work for me and allowed `kombu` to talk with my SSL configured Redis instance

```python
import ssl
from app import config
from celery import Celery


app = Celery(
    __name__,
    broker=config.REDIS_URI,
    broker_use_ssl={
        "ssl_cert_reqs": ssl.CERT_NONE,
        "ssl_keyfile": config.REDIS_SSL_KEY,
        "ssl_certfile": config.REDIS_SSL_CERT,
        "ssl_ca_certs": config.REDIS_SSL_CA_CERT,
    },
)
```

* You should setup the following configuration variables, pointing to the full path of your config (recommend) but relative paths are also supported
* `REDIS_SSL_KEY`: e.g. "redis.key"
* `REDIS_SSL_CERT`: e.g. "redis.crt"
* `REDIS_SSL_CA_CERT`: e.g. "ca.crt"

PR #1826, opened 2023-11-28
fix(docs): add Redis to the list of transports where SSL is supported

davidjrice:
Hey,

* Ran into this when trying to setup `celery` configuration to correctly support SSL using a Redis backend. It *is* supported but this bit of documentation had me second guessing things
* As-per [my comment](https://github.com/celery/kombu/issues/1493#issuecomment-1829589295) on #1493 with a working configuration, this is working as expected in the current release

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -82,7 +82,7 @@ class Connection:
 
     Note:
     ----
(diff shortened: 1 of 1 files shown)


Nusnus (approved):
The docs CI is fixed in this PR #1827

Issue #1774, opened 2023-07-27
`RuntimeError: Set changed size during iteration` with celery prefork


claudinoac:
In celery and kombu `v5.3.1`, we are constantly being hit by the error below
There are around 4 workers distributed across 3 different servers, listening to the same queues in prefork mode, with multiple processes (4 to 16 processes each).
Most of the workers are daemonized using celeryd init-script. Some of them are manually spawned through CLI for debug monitoring.
All of them are connected to the same redis broker
Some workers simply throw the error below and stop working, randomly.
We've been forced to restart those every few hours.
This was happening before `v5.3`, but from https://github.com/celery/celery/issues/7162 we thought that issue would be fixed after that update, but that unfortunately didn't happen.

[2023-07-27 01:28:05,206: CRITICAL/MainProcess] Unrecoverable error: RuntimeError('Set changed size during iteration')
Traceback (most recent call last):
  File "/opt/Conversus/venv/lib/python3.8/site-packages/celery/worker/worker.py", line 202, in start
    self.blueprint.start(self)
  File "/opt/Conversus/venv/lib/python3.8/site-packages/celery/bootsteps.py", line 116, in start
    step.start(parent)
  File "/opt/Conversus/venv/lib/python3.8/site-packages/celery/bootsteps.py", line 365, in start
    return self.obj.start()
  File "/opt/Conversus/venv/lib/python3.8/site-packages/celery/worker/consumer/consumer.py", line 336, in start
    blueprint.start(self)
  File "/opt/Conversus/venv/lib/python3.8/site-packages/celery/bootsteps.py", line 116, in start
    step.start(parent)
  File "/opt/Conversus/venv/lib/python3.8/site-packages/celery/worker/consumer/consumer.py", line 726, in start
    c.loop(*c.loop_args())
  File "/opt/Conversus/venv/lib/python3.8/site-packages/celery/worker/loops.py", line 97, in asynloop
    next(loop)
  File "/opt/app/venv/lib/python3.8/site-packages/kombu/asynchronous/hub.py", line 310, in create_loop
    for tick_callback in on_tick:
RuntimeError: Set changed size during iteration

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [x] This has already been asked to the [discussions forum](https://github.com/celery/celery/issues/7162) first (similar, but different issue)
- [x] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
[...]

PR #1830, opened 2023-12-07
fix: freeze set during ticks iter in async hub

claudinoac:
Relates to https://github.com/celery/kombu/issues/1774

With redis broker, when the redis server reaches timeout and celery worker is running a long cpu-intensive task, the transport tick automatically disconnects, and that's expected.

The connection error is thrown by the `redis-py` lib, but before that happens, the redis transport layer is auto-disconnected.
Instead of propagating the `ConnectionError`, the async hub gets its ticks set changed in the middle of the for loop that executes its callbacks, resulting in an error such as `RuntimeError: Set changed size during iteration`.
This happens because one of these callbacks belongs to the transport layer itself.

This solution basically freezes those ticks sets so they keep their consistency until the `ConnectionError` is thrown, which allows the celery worker to automatically recover itself, rather than just teardown due to an unrecoverable error like that one above.

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -134,6 +134,7 @@ def get_redis_error_classes():
             IOError,
             OSError,
             exceptions.ConnectionError,
+            exceptions.BusyLoadingError,
             exceptions.AuthenticationError,
             exceptions.TimeoutError)),
         (virtual.Transport.channel_errors + (
@@ -1229,7 +1230,7 @@ class Channel(virtual.Channel):
                 global_keyprefix=self.global_keyprefix,
             )
 
-        return redis.StrictRedis
+        return redis.Redis
 
     @contextmanager
     def conn_or_acquire(self, client=None):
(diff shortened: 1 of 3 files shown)


kombu/transport/redis.py
```diff
@@ -1229,7 +1230,7 @@ def _get_client(self):
                 global_keyprefix=self.global_keyprefix,
             )
 
-        return redis.StrictRedis
```
Nusnus, line 1232: Can you please elaborate on this change?

claudinoac: The library has renamed `StrictRedis` to `Redis` but kept the old name for backwards compatibility, but someday they'll remove it.

https://github.com/redis/redis-py/blob/master/CHANGES#L580

Nusnus (approved):
Thank you and sorry for the delay!

claudinoac:
@Nusnus Thank you!

Issue #1809, opened 2023-10-11
Failure for sentinel backend when using `global_prefix` configuration: functools.partial object has not attribute from_pool


adam-homeboost:
Given a celery config of:

```
broker_url = "sentinel://some_sential_address:6379"

broker_transport_options = {
    "global_keyprefix": "my_prefix",
    "master_name": "some_name",
}
```

Celery startup fails with:

```
AttributeError: 'functools.partial' object has no attribute 'from_pool'
```

Remove the "global_prefix" setting and this error does not happen.

The "partial" is set in `kombu.transport.redis:Channel._get_client.

Full Stack Trace:

```
[2023-10-11 02:02:32,121: CRITICAL/MainProcess] Unrecoverable error: AttributeError("'functools.partial' object has no attribute 'from_pool'")
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/kombu/transport/virtual/base.py", line 951, in create_channel
    return self._avail_channels.pop()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
IndexError: pop from empty list

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/celery/worker/worker.py", line 202, in start
    self.blueprint.start(self)
  File "/usr/local/lib/python3.11/site-packages/celery/bootsteps.py", line 116, in start
    step.start(parent)
  File "/usr/local/lib/python3.11/site-packages/celery/bootsteps.py", line 365, in start
    return self.obj.start()
           ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/celery/worker/consumer/consumer.py", line 340, in start
    blueprint.start(self)
  File "/usr/local/lib/python3.11/site-packages/celery/bootsteps.py", line 116, in start
    step.start(parent)
  File "/usr/local/lib/python3.11/site-packages/celery/worker/consumer/connection.py", line 21, in start
    c.connection = c.connect()
                   ^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/celery/worker/consumer/consumer.py", line 469, in connect
    conn = self.connection_for_read(heartbeat=self.amqheartbeat)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1838, opened 2024-01-04
Fix crash when using global_keyprefix with a sentinel connection

adam-homeboost:
Fix for https://github.com/celery/kombu/issues/1809.

When global_keyprefix is set, self.Client is no longer a class, it is a closure to create the class with the prefix argument.  This breaks in a place that expects this to be a class and not just any callable.  Fortunately, in this particular case, the place that is breaking doesn't actually need or care about the prefix and so it is safe to use the main redis client class.  It only cares about the pool that comes back.  This fix does not prevent the key prefix from being used later by kombu's own client extension.

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -1429,7 +1429,7 @@ class SentinelChannel(Channel):
 
         return sentinel_inst.master_for(
             master_name,
-            self.Client,
+            redis.StrictRedis,
         ).connection_pool
 
     def _get_pool(self, asynchronous=False):
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -1429,7 +1429,7 @@ class SentinelChannel(Channel):
 
         return sentinel_inst.master_for(
             master_name,
-            redis.StrictRedis,
+            redis.Redis,
         ).connection_pool
 
     def _get_pool(self, asynchronous=False):
--- a/t/unit/transport/test_redis.py
+++ b/t/unit/transport/test_redis.py
@@ -1770,6 +1770,7 @@ class test_RedisSentinel:
                 connection.channel().client.global_keyprefix
                 == 'some_prefix'
             )
+            connection.close()
 
 
 class test_GlobalKeyPrefixMixin:
--- a/t/unit/transport/test_redis.py
+++ b/t/unit/transport/test_redis.py
@@ -1737,15 +1737,18 @@ class test_RedisSentinel:
     def test_can_create_connection_with_global_keyprefix(self):
         from redis.exceptions import ConnectionError
 
-        connection = Connection(
-            'sentinel://localhost:65534/',
-            transport_options={
-                'global_keyprefix': 'some_prefix',
-                'master_name': 'not_important',
-            },
-        )
-        with pytest.raises(ConnectionError):
-            connection.channel()
+        try:
+            connection = Connection(
+                'sentinel://localhost:65534/',
+                transport_options={
+                    'global_keyprefix': 'some_prefix',
+                    'master_name': 'not_important',
+                },
+            )
+            with pytest.raises(ConnectionError):
+                connection.channel()
+        finally:
+            connection.close()
 
     def test_can_create_correct_mixin_with_global_keyprefix(self):
         from kombu.transport.redis import GlobalKeyPrefixMixin
--- a/t/unit/transport/test_redis.py
+++ b/t/unit/transport/test_redis.py
@@ -1734,6 +1734,40 @@ class test_RedisSentinel:
                 assert (params['connection_class'] is
                         SentinelManagedSSLConnection)
 
+    def test_can_create_connection_with_global_keyprefix(self):
+        from redis.exceptions import ConnectionError
+
+        connection = Connection(
+            'sentinel://localhost:65534/',
+            transport_options={
+                'global_keyprefix': 'some_prefix',
+                'master_name': 'not_important',
+            },
+        )
+        with pytest.raises(ConnectionError):
+            connection.channel()
+
+    def test_can_create_correct_mixin_with_global_keyprefix(self):
+        from kombu.transport.redis import GlobalKeyPrefixMixin
+
+        with patch('redis.sentinel.Sentinel'):
+            connection = Connection(
+                'sentinel://localhost:65534/',
+                transport_options={
+                    'global_keyprefix': 'some_prefix',
+                    'master_name': 'not_important',
+                },
+            )
+
+            assert isinstance(
+                connection.channel().client,
+                GlobalKeyPrefixMixin
+            )
+            assert (
+                connection.channel().client.global_keyprefix
+                == 'some_prefix'
+            )
+
 
 class test_GlobalKeyPrefixMixin:
 


kombu/transport/redis.py
```diff
@@ -1429,7 +1429,7 @@ def _sentinel_managed_pool(self, asynchronous=False):
 
         return sentinel_inst.master_for(
             master_name,
-            self.Client,
+            redis.StrictRedis,
```
auvipy, line 1432: should we also adjust some test cases to verify this change?

auvipy (changes requested):

kombu/transport/redis.py
adam-homeboost: Added tests...

50-Course:
Hello @adam-homeboost,thank you for providing this patch. I re-traced the steps and made a comparable build of both patches (this and replicated environment containing the bug).

Here are my analysis:

before - matching your exact debug information:
```sh
[2024-01-08 09:55:01,745: CRITICAL/MainProcess] Unrecoverable error: AttributeError("'functools.partial' object has no attribute 'from_pool'")
  File "/usr/local/lib/python3.11/site-packages/kombu/transport/virtual/base.py", line 951, in create_channel
    return self._avail_channels.pop()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
IndexError: pop from empty list

  File "/usr/local/lib/python3.11/site-packages/celery/worker/worker.py", line 202, in start
    self.blueprint.start(self)
  File "/usr/local/lib/python3.11/site-packages/celery/bootsteps.py", line 116, in start
    step.start(parent)
  File "/usr/local/lib/python3.11/site-packages/celery/bootsteps.py", line 365, in start
    return self.obj.start()
           ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/celery/worker/consumer/consumer.py", line 340, in start
    blueprint.start(self)
  File "/usr/local/lib/python3.11/site-packages/celery/bootsteps.py", line 116, in start
    step.start(parent)
  File "/usr/local/lib/python3.11/site-packages/celery/worker/consumer/connection.py", line 21, in start
    c.connection = c.connect()
                   ^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/celery/worker/consumer/consumer.py", line 469, in connect
    conn = self.connection_for_read(heartbeat=self.amqheartbeat)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/celery/worker/consumer/consumer.py", line 475, in connection_for_read
    return self.ensure_connected(
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/celery/worker/consumer/consumer.py", line 526, in ensure_connected
    conn = conn.ensure_connection(
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kombu/connection.py", line 406, in ensure_connection
    self._ensure_connection(*args, **kwargs)
  File "/usr/local/lib/python3.11/site-packages/kombu/connection.py", line 459, in _ensure_connection
    return retry_over_time(
           ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kombu/utils/functional.py", line 318, in retry_over_time
    return fun(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kombu/connection.py", line 934, in _connection_factory
    self._connection = self._establish_connection()
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kombu/connection.py", line 860, in _establish_connection
    conn = self.transport.establish_connection()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kombu/transport/virtual/base.py", line 975, in establish_connection
    self._avail_channels.append(self.create_channel(self))
                                ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kombu/transport/virtual/base.py", line 953, in create_channel
    channel = self.Channel(connection)
              ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kombu/transport/redis.py", line 741, in __init__
    self.client.ping()
    ^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kombu/utils/objects.py", line 31, in __get__
    return super().__get__(instance, owner)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/functools.py", line 1001, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kombu/transport/redis.py", line 1254, in client
    return self._create_client(asynchronous=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kombu/transport/redis.py", line 1210, in _create_client
    return self.Client(connection_pool=self.async_pool)
                                       ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kombu/transport/redis.py", line 1248, in async_pool
    self._async_pool = self._get_pool(asynchronous=True)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kombu/transport/redis.py", line 1433, in _get_pool
    return self._sentinel_managed_pool(asynchronous)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kombu/transport/redis.py", line 1427, in _sentinel_managed_pool
    return sentinel_inst.master_for(
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/redis/sentinel.py", line 356, in master_for
    return redis_class.from_pool(
           ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'functools.partial' object has no attribute 'from_pool'
```


after:

I am trying to run the tests, but hitting a few dead-end here. Precisely, I can't catch the test with:

```sh
$ python t/unit/transport/test_redis.py::test_can_create_connection_with_global_prefix
```

Other commands I have tried:
```sh
$ python -k <test_name::removed_for_conciseness> -o python_files=t/unit/transport/test_redis.py
```

Screenshots:

> image 


> image

t/unit/transport/test_redis.py
```diff
@@ -1734,6 +1734,40 @@ def test_sentinel_with_ssl(self):
...
+
+        connection = Connection(
+            'sentinel://localhost:65534/',
+            transport_options={
+                'global_keyprefix': 'some_prefix',
+                'master_name': 'not_important',
+            },
+        )
+        with pytest.raises(ConnectionError):
+            connection.channel()
```
50-Course, line 1748: Can we consider one more approach here:

```python
try:
    connection = Connection(
        'sentinel://localhost:65534/',
        transport_options={
            'global_keyprefix': 'some_prefix',
            'master_name': 'not_important',
        },
    )
    with pytest.raises(ConnectionError):
        connection.channel()
finally:
    connection.close()  # Ensure connection is closed even if an exception occurs

```

Should we consider testing potential exceptions that could occur during connection creation??

50-Course (changes requested):
I have been able to reproduce the simulations locally

I have an un-updated image here at https://github.com/50-Course/global-prefix-error-celery

t/unit/transport/test_redis.py
adam-homeboost: @50-Course 

These tests are inside a test class.  Try:

`pytest t/unit/transport/test_redis.py::test_RedisSentinel::test_can_create_connection_with_global_keyprefix`

In terms of adding things to the test, I hesitate because:

1) This is just the testing of the parameterization of the connection, not the connection itself.

2) This specific structure is a direct copy of the other sentinel connection tests in the same test class and follows the same methodology.  Note that this does not include connection to a REAL sentinel as that is tested elsewhere, as are the specific varieties of exceptions that can happen.

3) I believe strongly that unit tests should be as focused on the specific one thing they are testing as possible as this makes the code simpler, easier to understand and easier for future contributors to grasp and contribute to.  Testing specific kinds of connection error outside the scope of this change belong elsewhere (in my opinion) and (hopefully) already exist in the main redis module.

adam-homeboost:
In terms of adding things to the test, I hesitate because:

1) This is just the testing of the parameterization of the connection, not the connection itself.

2) This specific structure is a direct copy of the other sentinel connection tests in the same test class and follows the same methodology.  Note that this does not include connection to a REAL sentinel as that is tested elsewhere, as are the specific varieties of exceptions that can happen.

3) I believe strongly that unit tests should be as focused on the specific one thing they are testing as possible as this makes the code simpler, easier to understand and easier for future contributors to grasp and contribute to.  Testing specific kinds of connection error outside the scope of this change belong elsewhere (in my opinion) and (hopefully) already exist in the main redis module.

kombu/transport/redis.py
Nusnus, line 1432: `StrictRedis` [is deprecated](https://github.com/redis/redis-py/blob/b5d4d29409f83e9a0056c6ed12414bf58e1ea1c0/CHANGES#L583).
Better use Redis instead.

t/unit/transport/test_redis.py
Nusnus: > @50-Course
> 
> These tests are inside a test class. Try:
> 
> `pytest t/unit/transport/test_redis.py::test_RedisSentinel::test_can_create_connection_with_global_keyprefix`

```python
class test_Redis:

    def setup(self):
        self.connection = Connection(transport=Transport)
        self.exchange = Exchange('test_Redis', type='direct')
        self.queue = Queue('test_Redis', self.exchange, 'test_Redis')

    def teardown(self):
        self.connection.close()
```
If I am not mistaken it means it does not apply to your `connection` obj so I agree with a safer approach like @50-Course proposed.

Also, reviewing a sample of the other tests shows most tests have consistent resource cleanup manually, so it would make sense to apply it here as well.

Nusnus (changes requested):
All in all looks good - just needs to resolve the comments on the PR

adam-homeboost:
As requested: changed connection class to non-deprecated, added resource cleanup to exception test.

Nusnus (approved):
LGTM
Waiting for @50-Course @auvipy to approve/comment before merge.

50-Course:
Foremost, thank you for the heads-up on `pytest` issue. Found out it slipped past me, the omission of the `TestClass` - `test_channel` of which the function in question belong to and found it much later. 

Onto our conversation,
> In terms of adding things to the test, I hesitate because:
> 
>     1. This is just the testing of the parameterization of the connection, not the connection itself.

Thank you for clarification, I am well aware this is parameterization of the `connection` construct, precisely addressing the issue of `global_prefix` when creating the connection.

>     2. This specific structure is a direct copy of the other sentinel connection tests in the same test class and follows the same methodology.  Note that this does not include connection to a REAL sentinel as that is tested elsewhere, as are the specific varieties of exceptions that can happen.

Following this up, I have reviewed the few other tests, while in this scenario the test scope aims to catch the error from the `connection` such as the `get_pool_error`. I was suggesting additional tests could as well be considered so as not to leave the connection pool not cleaned up.

>     3. I believe strongly that unit tests should be as focused on the specific one thing they are testing as possible as this makes the code simpler, easier to understand and easier for future contributors to grasp and contribute to.  Testing specific kinds of connection error outside the scope of this change belong elsewhere (in my opinion) and (hopefully) already exist in the main redis module.

Point understood! I am also a firm believer of the single-responsibility principle. And this would have been a good approach, However, given the a `connection` setup, we might as well clean it up properly as found in the snippet shared by @Nusnus above.

50-Course (approved):
LGTM 🚀

auvipy (approved):

Nusnus:
Good work all!
Thank you @adam-homeboost for your contribution ❤️

PR #1892, opened 2024-01-20
Always convert azureservicebus namespace to fully qualified

ainscore:
Bug from the implementation discussed in issue #1345


auvipy:
thanks a lot!

PR #1953, opened 2024-03-01
Expose cancel callback from py-amqp channel.basic_consume

jcosmao:
When broker send Basic.cancel notification, py-amqp default behavior is to raise a ConsumerCancelled exception. It is then treated as an error even if connection and channel are still operationnal and connection is closed.

Basic.cancel may be sent by broker when a queue is deleted or when replicated queue leader change. py-amqp channel.basic_consume allow to define a callback function on this event. It may be useful to register this callback from kombu when consuming from a queue.


auvipy (commented):
can we add some tests to verify this?

Issue #1841, opened 2024-01-06
[BUG]datetime type serialization error 


ponponon:
A bug was found in the new version of kombu that serializes datetime type to a dictionary

<img width="953" alt="图片" src="https://github.com/celery/kombu/assets/38725104/db2b726c-ab43-4297-949e-bae033ba3935">

The cause was that I noticed this strange datetime format in my message queue：`{"created_at": {"__type__": "datetime", "__value__": "2024-01-06T16:14:46.368136"}}`

> Obviously, the format of the serialized message should look like this: `{"created at": "2024-01-06T16:27:16.458258"}`

But after I reduced the kombu version from 5.3.2 to 5.2.4, it didn't appear again

示例代码：

```python
from kombu import Connection, Exchange, Producer, Queue
from datetime import datetime

# 创建连接
connection = Connection('amqp://pon:pon@192.168.31.245:5672//')

# 创建交换机
exchange = Exchange('ye.events', type='topic')

# 创建队列
queue = Queue('my_queue', exchange=exchange, routing_key='take')

# 创建生产者
producer = Producer(connection, exchange=exchange, routing_key='take')

# 准备要发布的消息
message_body = {
    'created_at': datetime.utcnow()
}

# 发布消息
with producer:
    producer.publish(message_body, declare=[queue])

print("Message published successfully.")



```

ponponon:
I found the problem, the problem is with the _as function at https://github.com/celery/kombu/blob/main/kombu/utils/json.py, why introduce this stupid and smartass function?

If you use kombu as the producer and another library as the consumer, this will make it impossible to deserialise this message body

Issue #1821, opened 2023-11-10
Call out breaking change in 5.3.0


mrcljx:
I spent two hours trying to figure out why Kombu suddenly started decoding datetime objects. 

Even if this was a fix of a regression, 5.2.4 was "latest" for 15 months.

Would be really great if this was called out in the changelog.

Another person who ran into this:  https://github.com/celery/kombu/issues/1678

auvipy:
We generally accept breaking changes in 5.2.0 and 5.3.0 versions. the changes were documented but due to some problem in CI release note script it was overridden twice. would you mind share some workaround so that I can re add them to the 5.3.x upgrade consideration?

mrcljx:
I understand the need for this breaking change. It just came as a surprise when upgrading Celery, so a call out was all I needed. 🙂

I'm going forward with this workaround (our test suite passes again):

```
kombu.serialization.unregister("json")
kombu.serialization.register(
    "json",
    json.dumps, # maybe: orjson
    json.loads,
    content_type="application/json",
    content_encoding="utf-8",
)
```

Could be that this workaround is flawed for messages in flight - I could really see whether 5.2 would also emit __type__ for some fields that the above decoder wouldn't be able to handle.

Off-topic: I haven't seen the custom JSON coding behavior defined anywhere. Would be great to have a blurb (around `__type__` and `__value__`) in the docs at least. Especially as `json` is considered portable in contrast to `pickle`.

auvipy:
I highly appreciate your inputs here. I actually forgot. Now I will revisit the release notes again to mention these type of un intended breaking changes and work around to consider before upgrade. thanks

ulgens:
@auvipy Any chance to prioritize this? It looks like this is the only thing missing for 5.6.0.

Issue #1787, opened 2023-09-12
UUID and datetime are deserialization as dict


OrSecondNature:
celery==5.3.4
kombu==5.3.2
python==3.10.13

I'm using celery over rabbitmq and redis.
Trying to send a UUID or datetime to a celery task and get this value inside the celery task:
`{'__type__': 'datetime', '__value__': '2023-09-12T10:45:58.528141'}`

adriancarayol:
Might be related: https://github.com/celery/kombu/pull/1754

Issue #1749, opened 2023-06-08
Backwards incompatible change in 5.3.0 (json serialization for datetime and UUID)


vitorquintanilha:
Hello there,

Version 5.3.0 introduces a backwards-incompatible change which modifies how `datetime` and `UUID` objects are serialized. Before, they were serialized as strings. Now they are serialized as custom dicts.

Source of change: https://github.com/celery/kombu/pull/1575

This has an enormous potential to break existing services using kombu (which happened to me). I don't know if Kombu follows [semantic versioning](https://semver.org/) but, if it does, this change should have waited for a major version bump since it introduces an incompatible change to a public API.

If it doesn't, we could at least put a big warning on the changelog explaining the update and the steps to work around the change if needed.

This would avoid having other developers (like myself) wasting time investigating issues caused by this change.

Thanks in advance!

auvipy:
it was actually a bug before. now it works properly. we usually introduce new API's in 5.2,5.3 releases like django does

cloutierjo:
Beg to add more information, I would say current way look way more like a bug to me than string serialization for a datetime or uuid. I can understand for more complex type, but not single value type.

auvipy:
I'm open to accept more improvement in this regard. you can start with failing unit & integration tests :)

auvipy:
should be fixed now with new release https://github.com/celery/kombu/releases/tag/v5.3.1

JeffreyCA:
Yeah I was affected and surprised by this change as well. My project still uses Celery 5.1.2 so didn't expect it to consume the latest version of kombu.

LincolnPuzey:
> we usually introduce new API's in 5.2,5.3 releases like django does

It is true that Django introduces breaking changes in X.1,X.2 versions.
However these are clearly documented as such in the release notes, e.g.
https://docs.djangoproject.com/en/4.2/releases/4.2/#backwards-incompatible-changes-in-4-2
Did kombu make a similar effort? In my opinion, no, the change was listed in a dot point along with every other change for the release: https://docs.celeryq.dev/projects/kombu/en/stable/changelog.html / https://github.com/celery/kombu/releases

Also, speaking as someone who has done several Django upgrades on medium-large projects, these breaking changes are usually restricted to small, obscure parts of the Django API. Any breaking change to a major part of the Django API typically goes through a separate deprecation process:
https://docs.djangoproject.com/en/dev/internals/release-process/#deprecation-policy
[...]

PR #1955, opened 2024-03-01
add escape hatch for custom JSON serialization

vlindhol:
### Summary

As reported in multiple issues (#1841, #1821, #1787, #1749), it was a surprising breaking change when some Python types became wrapped in a `{ "__type__": ..., "__value__": ...}` envelope in 5.3.0 and subsequently deserialized back into the native types.

Using `kombu.util.json.register_type()` one could at least avoid e.g. `datetime` becoming deserialized into a `datetime` and keep it as a `str` as in <5.3.0, but it's currently impossible to get rid of the envelope itself.

This PR changes `register_type()` to accept `marker=None`, which signals that the JSON encoder should only transform the value using
the `encoder` arg, but should not stick the result in an envelope.

A side-effect of this is that the decoder thus becomes unnecessary to
specify (since there is no `__type__` in the resulting JSON that would
activate it). If `decoder` is omitted, default to a simple identity function.

This way one can replicate the behavior of pre-5.3.0 fairly well by overriding the default-registered types as follows:

```python
def _serialize_datelike(datetime_or_date):
    if not isinstance(datetime_or_date, datetime.datetime):
        datetime_or_date = datetime.datetime(
            datetime_or_date.year,
            datetime_or_date.month,
            datetime_or_date.day,
            0,
            0,
            0,
            0,
        )
    r = datetime_or_date.isoformat()
    if r.endswith("+00:00"):
        r = r[:-6] + "Z"
    return r


register_type(datetime.datetime, None, _serialize_datelike)
register_type(datetime.date, None, _serialize_datelike)
register_type(datetime.time, None, lambda o: o.isoformat())
register_type(uuid.UUID, None, lambda o: str(o))
register_type(Decimal, None, lambda o: str(o))
```

### Tests

Added some unit tests around `register_type()`, which were completely missing before. Also tests the new functionality.


thedrow (approved):
LGTM

Nusnus (commented):
Linting/pre-commit fails

Nusnus (changes requested):
Linting/pre-commit fails

Nusnus (changes requested):
https://github.com/celery/kombu/actions/runs/8262054534/job/22602777695#step:7:1

vlindhol:
> https://github.com/celery/kombu/actions/runs/8262054534/job/22602777695#step:7:1

Sorry, didn't see that one! I thought CI looked green 😅 Manually ran all CI commands in `ci.yaml` now.

Nusnus (approved):

Nusnus:
Thank you!

PR #2002, opened 2024-05-17
enable/fix test_etcd.py (resolves #2001)

### LINKED ISSUE #2001: test_etcd.py skipped and gives error when enabled
Hi, I'm new with kombu and tried to run the tests via tox (python 3.12). I saw and investigated why etcd is skipped. 

The message I saw was: 
SKIPPED [1] t/unit/transport/test_etcd.py:11: could not import 'etcd': No module named 'etcd'

This is because the module etcd is not in de tox requirements (never added?). A requirements file does exists in requirements/extras/etcd.txt. So added it to (what I think is the most logical file) requirements/test-ci.txt, with '-r extras/etcd.txt'

That resulted in test_etcd being executed however with error on self.patch. 
```
self = <t.unit.transport.test_etcd.test_Etcd object at 0x7f17cdb983e0>

    def setup_method(self):
        self.connection = Mock()
        self.connection.client.transport_options = {}
        self.connection.client.port = 2739
>       self.client = self.patch('etcd.Client').return_value
E       AttributeError: 'test_Etcd' object has no attribute 'patch'

t/unit/transport/test_etcd.py:19: AttributeError
```

Looking at the code changes it might (not sure) not have been working since it was ported to pytest: [Port etcd tests to py.test](https://github.com/celery/kombu/commit/f71ea5c803a29706afd195ef570607838dc53519). I assume the old class 'Case' implemented the patch function.

After changing the patch to patching and the connection setup similar to test_consul.py it passed the test.

I'll prepare and propose a pull request (also my first time).

FrankK-1234:
Enabled and fixed test_etcd.py. See [Issue 2001](https://github.com/celery/kombu/issues/2001).

Added etcd requirements and fixed the test in line with test_consul.py

### DIFF
--- a/requirements/test-ci.txt
+++ b/requirements/test-ci.txt
@@ -15,3 +15,4 @@ urllib3>=1.26.16; sys_platform != 'win32'
 -r extras/brotli.txt
 -r extras/zstd.txt
 -r extras/sqlalchemy.txt
+-r extras/etcd.txt
--- a/t/unit/transport/test_etcd.py
+++ b/t/unit/transport/test_etcd.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from array import array
 from queue import Empty
 from unittest.mock import Mock, patch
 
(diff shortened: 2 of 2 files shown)


pavanshanbhag (approved):
changes looks good

auvipy (approved):
good job!

PR #2007, opened 2024-05-26
Fix Redis connections after reconnect - consumer starts consuming the tasks after crash.

awmackowiak:
**TL;DR**
After updating to version 5.2.3, Celery workers stop processing tasks indefinitely after losing connection to the Redis broker.
    This issue does not occur in version 5.2.2. Analysis revealed that the problem is related to the removal of new Redis connections from the `on_tick` method and the behaviour of the `__del__` method, which removes these connections before they are properly registered.
    The solution is to modify the logic for removing connections from on_tick so that only connections that were properly added to the poller are removed, thereby resolving the reconnection issue.

This cannot be made without @ziollek because he figure out the GC problem and proposed a simple solution.
Big thanks to him :)

**Full Description**
_Problem_
After losing connection to the Redis broker, the Celery worker stops consuming tasks indefinitely. This issue didn't occur previously, but it started happening again after version 5.2.3. The discussion is extensive but reveals a few key problems:

- After a few restarts, the Celery worker stops consuming tasks: https://github.com/celery/celery/issues/8030#issuecomment-2067570550
- The problem doesn’t occur in version 5.2.2: https://github.com/celery/celery/issues/8030#issuecomment-1684160184

So, after a deep dive into the topic, we (with Ziolek) have three main questions to answer:

1. What was the change in version 5.2.3?
2. Why does a couple of restarts reveal the problem?
3. Why doesn’t the problem occur on the first reconnection?

_Analysis_
First, we used our VaaS environment to simulate Redis instance restarts. After a few restarts, the Celery worker wouldn’t consume tasks from Redis, so we easily reproduced the error. The problem seems to be a race condition because of non-deterministic behaviour. Sometimes the reconnection cannot be handled after one or two restarts, but sometimes (if you restart it quickly) the reconnection problem occurs even after 10 restarts.
Secondly, we compared the differences between versions 5.2.2 and 5.2.3 and tried to find the main change causing the problem:
https://github.com/celery/kombu/compare/v5.2.2...v5.2.3#diff-203e309d100714904255d0af4500dd295e40e4ffedfdc9295ab3b386718ca87eR1241-R1247.
One of the features added in this version was logic responsible for removing the connection to Redis from the on_tick method if the file descriptors (fds) are added to the set. This was added to fix a memory leak (because earlier on_tick was only adding items and never removing them), but it started causing the reconnection problem.

You may ask why it broken? The simple answer is: because it removes fresh connections to Redis from the connection pool.

How it Works:
[...]

PR #2011, opened 2024-05-29
Limit requests<2.32.0 due to docker-py issue 3256

Nusnus:
Extracted from https://github.com/celery/kombu/pull/2007 by @awmackowiak

Due to: https://github.com/docker/docker-py/issues/3256

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -2,3 +2,6 @@ typing_extensions; python_version<"3.10"
 amqp>=5.1.1,<6.0.0
 vine
 backports.zoneinfo[tzdata]>=0.2.1; python_version < '3.9'
+# due to this bug https://github.com/docker/docker-py/issues/3256
+# we need to hard-code the version of requests
+requests<2.32.0
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,9 @@ envlist =
     apicheck
     pydocstyle
 
-requires = tox-docker>=3.0
+requires =
+    tox-docker<=4.1
+    requests<2.32.0
 
 [gh-actions]
 python =
--- a/tox.ini
+++ b/tox.ini
@@ -63,7 +63,7 @@ dockerenv =
     RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS=-rabbit tcp_listeners [5672]
 
 [docker:rabbitmq]
-image = rabbitmq:3.12
+image = rabbitmq
 ports = 5672:5672/tcp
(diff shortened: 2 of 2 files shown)


Nusnus:
Don’t tell me it breaks due to new rabbitmq image now 🤦‍♂️

Nusnus:
> Don’t tell me it breaks due to new rabbitmq image now 🤦‍♂️

Trying to limit requests and rabbitmq image separately: https://github.com/celery/kombu/pull/2012

If it won’t work, it might need to be done in the same PR.

Nusnus:
> > Don’t tell me it breaks due to new rabbitmq image now 🤦‍♂️
> 
> Trying to limit requests and rabbitmq image separately: #2012
> 
> If it won’t work, it might need to be done in the same PR.

Ok it didn’t work. Trying to do both in a single PR.

Nusnus:
> > > Don’t tell me it breaks due to new rabbitmq image now 🤦‍♂️
> > 
> > 
> > Trying to limit requests and rabbitmq image separately: #2012
> > If it won’t work, it might need to be done in the same PR.
> 
> Ok it didn’t work. Trying to do both in a single PR.

Reverted the rabbitmq limit and added the tox.ini changes from the #2007 PR (forgot to add it in the first place)

awmackowiak:
@Nusnus I also tried to find a silver bullet for those dependency cycle versions... and I also failed. 
I also agree that is a messy problem which generates serious tech dept :(
But I think is a fresh case so maybe quickly those dependent libraries will update requests as well as everybody does.

max-wittig:
I was about to test https://github.com/celery/kombu/releases/tag/v5.4.0rc1, but sadly this prevents it as I don't feel like downgrading requests as 2.32.0 contains security fixes. To my understanding (maybe I'm wrong), requests was only downgraded to make sure that tests still work, right?

Couldn't this be just downgraded for the tests?

Nusnus:
> Couldn't this be just downgraded for the tests?

Yes, and it was already changed accordingly yesterday 🙂
https://github.com/celery/kombu/pull/2041

max-wittig:
@Nusnus Ahh okay! Sorry I've not seen that! So I will test the next pre-release 🙂  Thanks!

Nusnus:
@max-wittig 
> @Nusnus Ahh okay! Sorry I've not seen that! So I will test the next pre-release 🙂 Thanks!

[v5.4.0rc2](https://github.com/celery/kombu/releases/tag/v5.4.0rc2) is out!
The `requests` package is no longer limited to <2.32.0 per https://github.com/celery/kombu/pull/2041.

PR #2024, opened 2024-06-10
ConnectionPool can't be used after .resize(..., reset=True) (resolves #2018)

### LINKED ISSUE #2018: ConnectionPool can't be used after .resize(..., reset=True)
I close pool and then want to use it again. Reinit.

I found the `.resize()` method which has option `reset`. reset=True allow close all available connections and then `.setup()` will be called, but pool will be closed (`._closed=True`) and any `.acquire()` call will raise an error.

https://github.com/celery/kombu/blob/1df567a5e813b9030b858b454e51eb70ea928e6a/kombu/resource.py#L189

FrankK-1234:
Hi, can you please have a look at this pull request. It addresses the issue #2018.

What did I do to tackle the issue. First I added several test case for the resize method (in test_connection.py), as they were not available. Basically the test test_acquire_resize_reset() triggered the given issue.

In short the use of force_close_all() (used also in other cases) caused the pool to be set on _closed = True. In the acquire method this resulted in the RuntimeError('Acquire on closed pool'). 

I've added a parameter **close_pool** in force_close_all with default of True (old behaviour as used by other callers), which allows the pool to be kept open if set to False. In the resize method, reset branch I've added force_close_all(close_pool=False).

PLEASE NOTE:
While testing the resize I've found 2 issues related to resizing which I think are a problem. They are addressed as well.
- Issue 1. The resize calls the setup from within connectionPool and channelPool. This method allocated resource based on the current setting of limit. This works fine for fresh pools however if it is called within the resize then with acquired/dirty resources then it would create to many resources. I fixed this by subtracting the amount of dirty resource from the limit. See connection.py:1069/1095
- Issue 2. The _shrink_down (only called by resize) did not keep in mind the dirty resources and would not correctly shrink if there were dirty resouces. Now it shrink resources and leaves the dirty one as it was before. See resource.py:220

Please have a look at the changes,
Regards,

Frank

### DIFF
--- a/kombu/resource.py
+++ b/kombu/resource.py
@@ -144,15 +144,20 @@ class Resource:
     def collect_resource(self, resource):
         pass
 
-    def force_close_all(self):
+    def force_close_all(self, close_pool=True):
         """Close and remove all resources in the pool (also those in use).
 
         Used to close resources from parent processes after fork
         (e.g. sockets/connections).
+
+        Arguments:
+        ---------
+            close_pool (bool): If True (default) then the pool is marked
+                as closed. In case of False the pool can be reused.
         """
         if self._closed:
             return
-        self._closed = True
+        self._closed = close_pool
         dirty = self._dirty
         resource = self._resource
         while 1:  # - acquired
@@ -188,7 +193,7 @@ class Resource:
         self._limit = limit
         if reset:
             try:
-                self.force_close_all()
+                self.force_close_all(close_pool=False)
             except Exception:
                 pass
         self.setup()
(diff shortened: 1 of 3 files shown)


kombu/resource.py
```diff
@@ -211,7 +216,9 @@ def __exit__(
         resource = self._resource
         # Items to the left are last recently used, so we remove those first.
         with getattr(resource, 'mutex', Noop()):
-            while len(resource.queue) > self.limit:
+            # keep in mind teh dirty resources are not shrinked
```
auvipy, line 219:
```suggestion
            # keep in mind the dirty resources are not shrinking
```

kombu/resource.py
```diff
@@ -144,15 +144,20 @@ def release(self, resource):
     def collect_resource(self, resource):
         pass
 
-    def force_close_all(self):
+    def force_close_all(self, close_pool=True):
```
auvipy, line 147: since it is introducing a new API here I'm thinking of BC/including this in a minor version. not fully sure about it right now

thedrow (approved):

thedrow:
Thank you for the contribution!

spumer:
Good news!

Guys, only one question: why method `setup()` exists? Looks like pool can created new entry by calling `.new()` when limit not reached.

Due `setup()` calling we create "lazy" objects, then we check it "lazy" or not (in acquire()), and evaluate in `prepare()`.
Looks like removing pre-filling pool should simplify Resource class and all related logic.

Issue #2031, opened 2024-06-21
Kombu transport SQS channel using expired sts token to connect to SQS


chenxg283:
Having the following exception from time to time, not easy to reproduce though.  

CRITICAL/MainProcess] Unrecoverable error: Exception('Request HTTP Error  HTTP 403  Forbidden (b\'{"__type":"com.amazon.coral.service#ExpiredTokenException","message":"The security token included in the request is expired"}\')')
Traceback (most recent call last):
  File "/opt/app-root/lib64/python3.9/site-packages/celery/worker/worker.py", line 202, in start
    self.blueprint.start(self)
  File "/opt/app-root/lib64/python3.9/site-packages/celery/bootsteps.py", line 116, in start
    step.start(parent)
  File "/opt/app-root/lib64/python3.9/site-packages/celery/bootsteps.py", line 365, in start
    return self.obj.start()
  File "/opt/app-root/lib64/python3.9/site-packages/celery/worker/consumer/consumer.py", line 340, in start
    blueprint.start(self)
  File "/opt/app-root/lib64/python3.9/site-packages/celery/bootsteps.py", line 116, in start
    step.start(parent)
  File "/opt/app-root/lib64/python3.9/site-packages/celery/worker/consumer/consumer.py", line 746, in start
    c.loop(*c.loop_args())
  File "/opt/app-root/lib64/python3.9/site-packages/celery/worker/loops.py", line 97, in asynloop
    next(loop)
  File "/opt/app-root/lib64/python3.9/site-packages/kombu/asynchronous/hub.py", line 373, in create_loop
    cb(*cbargs)
  File "/opt/app-root/lib64/python3.9/site-packages/kombu/asynchronous/http/curl.py", line 122, in on_readable
    return self._on_event(fd, _pycurl.CSELECT_IN)
  File "/opt/app-root/lib64/python3.9/site-packages/kombu/asynchronous/http/curl.py", line 139, in _on_event
    self._process_pending_requests()
  File "/opt/app-root/lib64/python3.9/site-packages/kombu/asynchronous/http/curl.py", line 145, in _process_pending_requests
    self._process(curl)
  File "/opt/app-root/lib64/python3.9/site-packages/kombu/asynchronous/http/curl.py", line 191, in _process
    request.on_ready(self.Response(
  File "/opt/app-root/lib64/python3.9/site-packages/vine/promises.py", line 168, in __call__
    svpending(*ca, **ck)
[...]

PR #2035, opened 2024-06-23
fix to refresh sts token at 80% of TTL for issue #2031

chenxg283:
Fix for issue #2031 

1. merge the common code to refresh the sts token in one if-block, i.e. either when sts token not present or when the token is going to expire soon;
2. replanced the datetime.utcnow() as it is deprecated;
3. refresh the sts token at 80% of its TTL.

### DIFF
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -128,7 +128,7 @@ import base64
 import socket
 import string
 import uuid
-from datetime import datetime, timezone, timedelta
+from datetime import datetime, timedelta, timezone
 from queue import Empty
 
 from botocore.client import Config
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -128,7 +128,7 @@ import base64
 import socket
 import string
 import uuid
-from datetime import datetime
+from datetime import datetime, timezone, timedelta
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
we will need extensive test coverage to very this changes

Nusnus:
> we will need extensive test coverage to very this changes

I am 100% with @auvipy here. I’ve converted the PR to draft until tests can be added.

Thank you and I hope you understand @chenxg283 🙏

chenxg283:
Sorry, i am new to this.  Am I supposed to write the code for the test coverage?

Nusnus:
> Sorry, i am new to this. 

It’s fine :)

> Am I supposed to write the code for the test coverage?

Yes, you need to add automatic tests (that the CI will run automatically, but you can also run locally using `tox`).
These tests need to validate that the changes the PR introduces are indeed working as expected. They also need to ensure that there are no bugs, to the best reasonable effort possible.

You can see the existing [test suites](https://github.com/celery/kombu/tree/main/t) to check how other tests are implemented.

auvipy (changes requested):
maybe this can be done with https://github.com/celery/kombu/pull/2216?

Issue #1835, opened 2023-12-19
SQS - Failure to restore unacked messages


YahelOpp:
Versions:

- Python 3.10
- Kombu 5.3.4
- Celery 5.3.6
- Boto3 1.33.7

After seeing messages stuck in `NotVisible` state at our production queues, I've set out to verify the warm shutdown mechanism works properly. My test setup:
- Simple task: sleeps for a desired amount of seconds, then retries with a countdown of 10.
- Max retries: 3
- Single worker, threads pool, concurrency 5

What I see happens:
- A task is run
- It raises the retry -> putting a new message in the queue
- The message is consumed by the same worker directly (message `received`)
- During the countdown period, the message is in `NotVisible` state
- During the countdown period, I shutdown the worker with SIGTERM
- The worker successfully starts a `Warm shutdown`
- The worker notes that it is restoring 1 unacked message
- The message **stays** in `NotVisible` state

Expected behavior: the message should move to `Visible` state.

A possible explanation I've found: 
By manually changing the code in `kombu/transport/SQS.py` at the `_put` function in the case `message.get('redelivered')`, adding a duplicate call to `change_message_visibility` - the problem seemed to go away. I've tested this against real live SQS and against Localstack. In both the problem was consistent and solved once I added the duplicate call.

To sum up, the problem might be on boto3's side and not in kombu, but I'm unsure about this.
I will be happy to assist and make a PR / provide a simple testing scenario.

Thanks

YahelOpp:
The patch I've done that makes it work as expected is:
```

# Patch a fix for SQS channel
original_put = Transport.Channel._put

def patched_put(self, queue, message, **kwargs):
    original_put(self, queue, message, **kwargs)
    if message.get("redelivered"):
        q_url = self._new_queue(queue)
        c = self.sqs(queue=self.canonical_queue_name(queue))
        c.change_message_visibility(
            QueueUrl=q_url,
            ReceiptHandle=message['properties']['delivery_tag'],
            VisibilityTimeout=0
        )


Transport.Channel._put = patched_put
```

pbudzon:
I think this is a similar issue to what we're seeing, but we see that during a cold shutdown. Our scenario is:

- run celery with concurrency > 1
- run a long(ish) running task
- send SIGQUIT to celery while task is running
- see `Restoring X unacknowledged message(s)` warning in the log
- the message still shows as in-flight (unavailable, invisible) in the SQS

Relevant settings:
```
[...]

Issue #1819, opened 2023-11-07
SQS messages can be dropped during shutdown


mgorven:
When celery is stopped gracefully, the in-flight request to SQS is not completed. If SQS returns messages for this request it considers them delivered, but Celery doesn't do anything with them (even if `task_reject_on_worker_lost` is set). SQS will redeliver the message when the visibility timeout expires, but this isn't great for a graceful shutdown.

SQS uses `CurlClient` to make requests, which uses `celery.worker.loops.asynloop` to do IO. The event loop is the first thing to stop when `SIGTERM/SIGINT` is received, and so in-flight requests are simply dropped.

auvipy:
@rafidka any insight to share?

rafidka:
@auvipy , I would need to do some debugging on this. I will try to allocate some time for it, but I cannot promise anything this week.

@mgorven , it would help if you have some easy step-by-step reproduction.

mgorven:
I don't have an easy repro of actually losing events, I'll work on that. What I've done is add debug prints in `CurlClient._setup_request()` and `CurlClient._process()` and I can see that a request is started but never finished during shutdown: https://gist.github.com/mgorven/f671b3b9384b2814d86f7e99451a2936

mgorven:
1. Setup celery with an SQS queue which isn't processing other events and with a visibility timeout of a few minutes
2. In another shell have a `celery call` command ready to dispatch a task
3. Add a debug print in `CurlSetup._setup_request()` so you can see when a request is started
4. Run `celery worker`
5. Ctrl-C the worker just after it starts an SQS request
6. Run `celery call`
7. Start the worker again

You'd expect the worker to immediately process the dispatched task, but instead it only does so after the visibility timeout expires.

Nusnus:
Potentially fixed with https://github.com/celery/celery/pull/9213 (although it’s in Celery, not Kombu) @mgorven @rafidka

Nusnus:
> Potentially fixed with [celery/celery#9213](https://github.com/celery/celery/pull/9213) (although it’s in Celery, not Kombu) @mgorven @rafidka

[Celery v5.5.0b3](https://github.com/celery/celery/releases/tag/v5.5.0b3) released.

ishan-1010:
[...]

PR #2049, opened 2024-07-01
SQS: when redelivering a message apply visibility timeout based on polling time

pbudzon:
Fix for: https://github.com/celery/kombu/issues/1835
Partially also: https://github.com/celery/kombu/issues/1819
Possibly related: https://github.com/celery/celery/discussions/8583

**Issue**
when not using short polling (`wait_time_seconds > 0`) and `task_acks_late=True` it is possible for message to become stuck with full visibility timeout when celery is shut down. I was able to replicate that consistently with cold shutdowns. 

_Similar issue happens with warm shutdowns (see related discussion/bugs above). And with `task_acks_late=False` I believe this causes the message to become lost completely._

**To reproduce**
1. Execute a long(ish) running job (long enough for you to perform those steps).
2. Send `SIGQUIT` to celery master process.
3. See `Restoring X unacknowledged message(s)`.
4. See that the message is still in-flight with a full visibility timeout.

Note: this is a race condition, so it doesn't happen every time, but in my tests it happened MOST of the time (like 80%).

**What happens**
I believe the internal details are described in https://github.com/celery/kombu/issues/1819. 
On AWS side, you can see in CloudTrail logs the `ChangeMessageVisibility` call appearing correctly, but right after it there is (at least one, sometimes more, depending on your `concurrency`) a `ReceiveMessage` call which fetches the message again from the queue, causing its visibility timeout to be set back to non-zero (whatever your settings are). 

Screenshot 2024-07-01 at 19 52 45

**This fix**
This change changes the visibility timeout of the redelivered message from 0 (zero) to `wait_time_seconds` value. While this doesn't solve the underlying issue, it prevents the message from being re-fetched by the still-running `ReceiveMessage` call(s). 
For most scenarios this should create a considerable improvement, because:
1. With `task_acks_late=True`, the message will be hidden for `wait_time_seconds` (default 10 seconds), instead of visibility timeout (default 30 minutes). **I have tested that change in this scenario and it is working as expected.**
2. With `task_acks_late=False`, this should prevent the message from being lost completely. **I did not test that scenario, however.**


kombu/transport/SQS.py
```diff
@@ -439,7 +439,7 @@ def _put(self, queue, message, **kwargs):
             c.change_message_visibility(
                 QueueUrl=q_url,
                 ReceiptHandle=message['properties']['delivery_tag'],
-                VisibilityTimeout=0
+                VisibilityTimeout=self.wait_time_seconds
```
auvipy, line 442: can you please update the tests as well?

auvipy (changes requested):

pbudzon:
Issue also reported here: https://github.com/celery/celery/discussions/7283 back in 2022.

pbudzon:
Do you need anything else to merge this?

auvipy (commented):
should we also consider updating related documents for this?

thedrow:
@Nusnus Can you please take a look?

Nusnus:
> @Nusnus Can you please take a look?

I haven’t fully reviewed this PR yet, but I offer an alternative approach at https://github.com/celery/celery/pull/9213.

@pbudzon check it out, it might solve your original problem (still using 10s).

Nusnus:
> @Nusnus I'll see if we can upgrade internally to celery 5.5 and if this new approach solves the problem.
> 
🔥

> The background for this specific issue, as far as I understood, was caused by a race condition inside kombu where ReceiveMessage calls are made by using an async library and happen "in the background" of everything else - and they are not properly (or quickly enough?) terminated when shutdown happens, so part of the code restoring messages to the queue on shutdown can happen while there's an active ReceiveMessage call still "running" which then immediately fetches the restored message (removing it from the queue or hiding it again, depending on `task_acks_late`) and then celery exists so the message is lost (or stays hidden with visibility timeout, again depending on `task_acks_late`).
> 
> That is my understanding based partially on what I saw in the code and on this: #1819
> 
Basically yeah, the worker terminates too fast. It is very confusing during debugging because the debugger keeps the worker process alive, but I am able to replicate it using the smoke tests so I could test automatically.

> So, with the new shutdowns, I imagine this would still be an issue with the hard shutdown and cold shutdown. Others - I cannot say without testing. That would mainly depend on whether the async ReceiveMessage is killed fast enough, but that's controlled by kombu not celery, so I'm not sure if it's even possible to solve this from celery's side (but I don't know all the source code, so again, I don't know for sure).
> 
> What would you say is the best approach to terminating celery when we have time constraints (like in containers services like ECS, etc, where we have 1-2 minutes to exit cleanly) and long-running tasks - so we are fairly sure the tasks won't be able to finish within the time limit anyway, but we want to have absolute certainty they will end up back in the queue? Right now, we're using Cold Shutdowns and (apart from this bug, which we've fixed with this PR applied directly in our environment) it performs as expected returning the messages to the queue correctly.

The approach with the soft shutdown is to allow the user to define a timer before everything collapses. 10-30 seconds should be enough. It means it will give a “limited warm shutdown” of 10-30s (for example) so the tasks will run a bit more. Another `SIGQUIT` will stop the tasks but still wait for the timeout to expire so the race condition should be “avoided”.

The hard shutdown is never used directly. It is the result of multiple SIGQUIT signals (or SIGINT or SIGTERM if REMAP_SIGTERM is configured). But yes, it will do the same as now, forcing termination of the worker process immediately.

Nusnus:
> @Nusnus I'll see if we can upgrade internally to celery 5.5 and if this new approach solves the problem.

[Celery v5.5.0b3](https://github.com/celery/celery/releases/tag/v5.5.0b3) released.

Nusnus (commented):
Hey there, have you tried v5.5 beta?
Do you still need the changes of this PR or does the soft shutdown fix your use-case?

Thank!

pbudzon:
@Nusnus apologies, we did not have a chance to test the 5.5 yet. I'm hoping we'll be able to do that in the coming weeks - I'd like to keep this PR open until we can confirm this issue has been resolved, please.

auvipy:
@spawn-guy it would be great if you can check it out. as you are doing SQS on production and contributing to SQS lately...

spawn-guy:
sounds interesting - i should test it out

auvipy (approved):

PR #2059, opened 2024-07-12
SQS: add support for passing MessageAttributes

aviramha:
Hi!
I looked for a contributing guide and couldn't find one, so apologies if I missed any requirement/need.

SQS supports MessageAttributes, and with this PR, users using celery can easily add message attributes like this:
```python
from celery import Celery, signature

# Initialize the Celery app with SQS as the broker
app = Celery('sqs_producer', broker='sqs://')

# AWS configuration
app.conf.update(
    broker_transport_options={
        'region': 'eu-north-1',
        'polling_interval': 1,
    },
    task_default_queue='ManualTesting1'
)


# Custom signature function
def custom_signature(message_body):
    return signature('sqs_producer.send_message_to_sqs', args=(message_body,))

if __name__ == '__main__':
    message_body = {
        'key1': 'value1',
        'key2': 'value2'
    }
    # Using the custom signature to create a task
    task = custom_signature(message_body)
    # Apply the task asynchronously
    result = task.apply_async(message_attributes = {
    'Attribute1': {'DataType': 'String',
            'StringValue': 'STRING_VALUE'}
    })

```

Notes:
- I don't put the message attributes into the message properties - I feel that's a bit weird to include it there, but would understand if that might be problematic (if someone used this attribute before and now it won't appear)

This is probably useful for many reasons, but in our case it's blocking use of feature we provide over it - https://github.com/metalbear-co/mirrord/issues/2066

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -11,7 +11,7 @@
 :release-date: 11 July, 2024
 :release-by: Tomer Nosrati
 
-The ``requests`` is no longer limited to <2.32.0 per #2041.
+The ``requests`` package is no longer limited to <2.32.0 per #2041.
 Contribution #2007 by @awmackowiak was confirmed to have solved the Redis reconnection bug.
 
 - Bump mypy from 1.10.0 to 1.10.1 (#2039)
--- a/docs/reference/kombu.transport.SQS.rst
+++ b/docs/reference/kombu.transport.SQS.rst
@@ -63,3 +63,11 @@ The above policy:
 +-----------------------------------------+--------------------------------------------+
 | ``6th attempt``                         | 640 seconds                                |
 +-----------------------------------------+--------------------------------------------+
+
+
+Message Attributes
+------------------------
+
+SQS supports sending message attributes along with the message body.
+To use this feature, you can pass a 'message_attributes' as keyword argument
+to `basic_publish` method.
--- a/kombu/transport/SQS.py
+++ b/kombu/transport/SQS.py
@@ -34,6 +34,14 @@ up to 'prefetch_count' messages from queueA and work on them all before
 moving on to queueB.  If queueB is empty, it will wait up until
 'polling_interval' expires before moving back and checking on queueA.
 
+Message Attributes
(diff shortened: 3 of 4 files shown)


auvipy (changes requested):
thanks for the PR! this will need sufficient tests and possibly some documentation as it seems to be a somewhat new feature for SQS

aviramha:
I wasn't sure what's expected in terms of tests and docs - also couldn't build docs - got this error:
```
make lint
flake8 "kombu" "t"
(cd "docs/"; /Library/Developer/CommandLineTools/usr/bin/make apicheck)
Makefile:12: *** commands commence before first target.  Stop.
make: *** [apicheck] Error 2
```

aviramha:
Found how to run tests, should be okay now.

Nusnus (approved):
LGTM

@aviramha can you share some logs showing it works on an actual environment though?

Just to be on the safe-side so we can also see it actually works (apart of the code review which makes sense to me + the unit test which looks fine).

Thanks!

aviramha:
```
mirrord-operator-557c4cf78b-vrjfh mirrord-operator   2024-07-12T13:10:41.952140Z TRACE sqs_splitting::queue_splitter: new message to filter. message attributes: Some({"Attribute1": MessageAttributeValue { string_value: Some("STRING_VALUE"), binary_value: None, string_list_values: None, binary_list_values: None, data_type: "String" }})!
mirrord-operator-557c4cf78b-vrjfh mirrord-operator     at operator/sqs-splitting/src/queue_splitter.rs:357 on ThreadId(2)
mirrord-operator-557c4cf78b-vrjfh mirrord-operator     in sqs_splitting::queue_splitter::filter_messages with queue_url: "https://sqs.ca-central-1.amazonaws.com/xxx/alpha-ggg-hh-ppp", main_output_queue_url: "https://sqs.us-east-1.amazonaws.com/xxx/mirrord-main-gggg-kkkk-queue-alpha-bbb-ppp-zzzz"
````

Issue #2097, opened 2024-08-12
ConsumerMixin no longer send heartbeat if `timeout=None`


vitorquintanilha:
:wave: 

First I want to notice that the change in https://github.com/celery/kombu/issues/1997 is backwards incompatible, and still was release on a **minor version** with **no breaking change warning or mention on the Changelog**. This change broke all our consumers.

Second I'd like to question the logic behind the change. Why should `timeout=None` should cause heartbeats not to be sent? `timeout=None` means that we don't want to have any timeout while waiting for new messages. In this case we STILL want to send hearbeats to the broker, otherwise our connections will get killed.

Would you care to explain why it make sense that `timeout=None` should make the client stop sending hearbeats? What should I do if I don't want a timeout (e.g. I want my consumer to run forever, regardless of how long it takes for new messages to arrive) and I still want heartbeats to be sent so my connection is not broken?

Am I not properly understanding the API here?

Thanks!

Nusnus:
@auvipy @smart-programmer @FrankK-1234 [#2016](https://github.com/celery/kombu/pull/2016)

FrankK-1234:
Hello @vitorquintanilha,

After a closer look at the changes (from @smart-programmer) and the stated implied change . I have to agree it is not backwards compatible, perhaps the misunderstanding came from the name heartbeat_check. Actually I think there would be a problem if the parameter safety_interval is changed as that dictates how fast the heartbeat_check is called (it should be called about every second). So the question arise if this is the right place to implement that call.

My advice is to revert the change including the tests I had added, sorry for just assuming a right implementation. @auvipy could you make the revert if you agree? I think there might be another way to reach the original problem and that is to only call heartbeat_check when conn.heartbeat is set, but that is not up to me to decide.

As a workaround one could call the method with a high timeout (e.g. 1000000000) value which would still trigger the heartbeat check but never (well 31 years without a message) trigger the time-out raise.

With kind regards,

Frank

PR #1998, opened 2024-05-12
enhance: allow users to disable broker heartbeats 

smart-programmer:
broker heartbeats clutter the logs console which prevents users from seeing important logs. this allows users to disable broker heartbeats if timeout isn't provided.

[reference issue](https://github.com/celery/kombu/issues/1997)

PR #2104, opened 2024-09-05
Revert: allow users to disable broker heartbeats by not providing a timeout (#2097, #2016)

FrankK-1234:
Hello,

In issue #2097 it is mentioned that #1997, #1998, #2016 caused heartbeats not to be sent any more if timeout is None or 0. I'm not sure if we want to revert the change from @smart-programmer but this pull request will do so. Perhaps @auvipy can give his opinion. I reverted the change and changed the test case to expect the opposite old behavior which is to send heartbeats even if timeout is None or 0.

Please review and decide if revert is wanted or not.

With kind regards,

Frank

### DIFF
--- a/kombu/mixins.py
+++ b/kombu/mixins.py
@@ -195,11 +195,10 @@ class ConsumerMixin:
                 try:
                     conn.drain_events(timeout=safety_interval)
                 except socket.timeout:
-                    if timeout:
-                        conn.heartbeat_check()
-                        elapsed += safety_interval
-                        if elapsed >= timeout:
-                            raise
+                    conn.heartbeat_check()
+                    elapsed += safety_interval
+                    if timeout and elapsed >= timeout:
+                        raise
                 except OSError:
                     if not self.should_stop:
                         raise
--- a/t/unit/test_mixins.py
+++ b/t/unit/test_mixins.py
@@ -91,7 +91,7 @@ class test_ConsumerMixin:
             next(it)
         c.connection.heartbeat_check.assert_called()
 
-    def test_consume_drain_no_heartbeat_check(self):
+    def test_consume_drain_heartbeat_check_no_timeout(self):
         c, Acons, Bcons = self._context()
         c.should_stop = False
         it = c.consume(no_ack=True, timeout=None)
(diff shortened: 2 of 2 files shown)


auvipy (changes requested):
can you please share more detail explanations please?

FrankK-1234:
Hello @auvipy,

Let me try and recap the changes. The original request #1997 change was to prevent heartbeats when timeout was None or zero. This was build by @smart-programmer and accepted by you after I've added some test (#2016). However the claim in issue #2097 is that this changed the behaviour of the interface (@vitorquintanilha). The heartbeats will not be send anymore if timeout is None or zero. After looking more closely I agree with the changed behaviour. Previously it always called conn.heartbeat_check and now only if timeout is set. For me the function name is perhaps confusing because it is not checking for hearbeats but actually sending them.

So the original code always called conn.heartbeat_check upon a timeout:
```
              except socket.timeout:
                    conn.heartbeat_check()
                    elapsed += safety_interval
                    if timeout and elapsed >= timeout:
                        raise
```

While the new code only called it when the timeout paramter was set and non zero:
'''
                except socket.timeout:
                    if timeout:
                        conn.heartbeat_check()
                        elapsed += safety_interval
                        if elapsed >= timeout:
                            raise
'''

This changed the behaviour for users which rely on heartbeat being send but who do not (want) to specify a timeout and thus broke their code after an upgrade. The workaround I suggested is to use a high timeout value. However I think we should consider reverting the change, which this pull request does. 

I did not remove the test cases but adapted them to reflect the 'wanted' behaviour.

Is this enough explanation to decide what to do (not sure what to add more)? 

Thanks in advance,

Frank

auvipy:
yes make sense!

auvipy (approved):

auvipy:
should we reopen https://github.com/celery/kombu/issues/1997? or this PR fix both?

FrankK-1234:
> should we reopen #1997? or this PR fix both?

This PR does not fix #1997, not sure what to do with the PR the proposed solution is incorrect. The original problem, or not so nice behaviour in logging, remains. Is it worth the hassle?

vitorquintanilha:
Thanks everyone for your help on this topic :)

Issue #2113, opened 2024-09-12
Kombu 5.4.1 not working with Celery[SQS] 5.4.0


Hawkline451:
Celery worker is running but is not picking any messages from the SQS queue.

Kombu 5.4.0 works fine.

Nusnus:
The changes regarding SQS in v5.4.1 were:
* SQS: Fixes #2091 queue_delete() method doesn't actually delete the queue by @ivanprjcts in https://github.com/celery/kombu/pull/2099
* Validate SQS SSL certs by default by @poundifdef in https://github.com/celery/kombu/pull/2094

@Hawkline451 can you please share more info?
Is there any error? Does it sometimes work sometimes doesn’t?

If I had to guess, I’d say it’s the SSL [changes](https://github.com/celery/kombu/pull/2094/files) (`validate_cert` changed from default `false` to default `true`), but the original PR says
> ... I'm running python 3.12.4 and tested with the latest Celery+Kombu from github.

So I am not completely sure.
@ivanprjcts @poundifdef can you please give it a look as well in case I am missing something?

Thank you!

imran8294:
Hi @Nusnus 
Kombu `5.4.1` is not compatible with Celery `5.4.0.` As a result, the messages are not being consumed. Although the Celery workers are initialized, they are unable to retrieve messages from the queue.

With kombu <= `5.4.0` its working as expected.

Nusnus:
> Hi @Nusnus 
> Kombu `5.4.1` is not compatible with Celery `5.4.0.` As a result, the messages are not being consumed. Although the Celery workers are initialized, they are unable to retrieve messages from the queue.
> 
> With kombu <= `5.4.0` its working as expected.  

@thedrow @auvipy 

Fyi.

As this is an SQS only issue I prefer not to yank the last release but revert the PR that caused it (once found) and release a new patch.

auvipy:
Yeah that should work

Hawkline451:
> @Hawkline451 can you please share more info? Is there any error? Does it sometimes work sometimes doesn’t?
> 
> If I had to guess, I’d say it’s the SSL [changes](https://github.com/celery/kombu/pull/2094/files) (`validate_cert` changed from default `false` to default `true`), but the original PR says
> 
> > ... I'm running python 3.12.4 and tested with the latest Celery+Kombu from github.
> 

There is no errors or warnings, we test this running the worker in an EC2 server in debug.

Aditional info
```
boto3==1.35.5
[...]

PR #2114, opened 2024-09-14
Revert "Validate SQS SSL certs by default"

Nusnus:
Reverts celery/kombu#2094 per the conversation in https://github.com/celery/kombu/issues/2113


auvipy (approved):
seems we should have proper tests before merging in the original PR

Issue #279, opened 2013-11-20
Support message sizes > 256KB by using a third-party backend...


diranged:
SQS only supports messages up to 256KB. Given that limitation, its very easy to hit the limit and fail your task submission. Here's a simple example:

> > > group(benchmark_tasks.WaitTask().s(args=[0, data]) for i in xrange(1024)).apply_async()
> > >   Traceback (most recent call last):
> > >     File "<console>", line 1, in <module>
> > >     File "/opt/nextdoor-ve/lib/python2.7/site-packages/celery/canvas.py", line 194, in apply_async
> > >       return self._apply_async(args, kwargs, *_options)
> > >     File "/opt/nextdoor-ve/lib/python2.7/site-packages/celery/app/builtins.py", line 213, in apply_async
> > >       list(tasks), result.serializable(), gid, args), *_options
> > >     File "/opt/nextdoor-ve/lib/python2.7/site-packages/celery/app/task.py", line 474, in apply_async
> > >       *_options)
> > >     File "/opt/nextdoor-ve/lib/python2.7/site-packages/celery/app/amqp.py", line 250, in publish_task
> > >       *_kwargs
> > >   File "/opt/nextdoor-ve/lib/python2.7/site-packages/kombu/messaging.py", line 164, in publish
> > >       routing_key, mandatory, immediate, exchange, declare)
> > >     File "/opt/nextdoor-ve/lib/python2.7/site-packages/kombu/connection.py", line 453, in _ensured
> > >       return fun(_args, *_kwargs)
> > >     File "/opt/nextdoor-ve/lib/python2.7/site-packages/kombu/messaging.py", line 180, in _publish
> > >       mandatory=mandatory, immediate=immediate,
> > >     File "/opt/nextdoor-ve/lib/python2.7/site-packages/kombu/transport/virtual/__init__.py", line 469, in basic_publish
> > >       exchange, routing_key, *_kwargs)
> > >     File "/opt/nextdoor-ve/lib/python2.7/site-packages/kombu/transport/virtual/exchange.py", line 61, in deliver
> > >       _put(queue, message, *_kwargs)
> > >     File "/opt/nextdoor-ve/lib/python2.7/site-packages/kombu/transport/SQS.py", line 232, in _put
> > >       q.write(m)
> > >     File "/opt/nextdoor-ve/lib/python2.7/site-packages/boto/sqs/queue.py", line 220, in write
> > >       delay_seconds)
[...]

PR #2116, opened 2024-09-18
Integrate large payload support for SQS

auvipy:
This adds support for handling large payloads in SQS. The 'sqs_extended_client' is imported and utilized for fetching file from S3 as payload when necessary.

As Kombu asynchronously fetches new messages from the queue, not using the standard boto3 APIs, we have to manually fetch the s3 file, rather than rely on the sqs_extended_client to perform that action

Relates to: celery/kombu#279

### DIFF
--- a/kombu/asynchronous/aws/ext.py
+++ b/kombu/asynchronous/aws/ext.py
@@ -4,12 +4,14 @@ from __future__ import annotations
 
 try:
     import boto3
+    import sqs_extended_client
     from botocore import exceptions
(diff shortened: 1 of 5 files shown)


kgwangehee:
Hello, 

Do you have any roadmap or due date to integrate this feature to support large message which more than 256kb(in SQS Message Broker)

auvipy:
Coverage needs to be improved

auvipy (commented):
this seems to need more test coverage

PR #2129, opened 2024-09-30
Apply errback and callback when retry occurs

thedrow:
Before this change, retry over time did not call the errback or callback if specified in `broker_transport_options`.
This PR fixes this bug by including them in connection options.


thedrow:
@Nusnus Added tests.

Nusnus (approved):
Cool, thanks!

PR #2148, opened 2024-10-13
Update the transport options according to the retry policy

thedrow:
In https://github.com/celery/kombu/blob/2f5882331237147ccd34fef646e7b38c338430bb/kombu/messaging.py#L234
we ask for the default channel and since it is not created and the https://github.com/celery/kombu/blob/2f5882331237147ccd34fef646e7b38c338430bb/kombu/connection.py#L957 call relies on transport options and not the retry policy.

### DIFF
--- a/kombu/messaging.py
+++ b/kombu/messaging.py
@@ -182,6 +182,7 @@ class Producer:
                 declare.append(self.exchange)
 
         if retry:
+            self.connection.transport_options.update(retry_policy)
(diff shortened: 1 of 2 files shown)


Nusnus (approved):
Please add tests.
But LGTM.

Nusnus:
CleanShot 2024-10-13 at 20 10 07

Again? 😄

Nusnus (approved):

Issue #1785, opened 2023-09-09
Multiple celery fork pool workers don't work


lzgabel:
> ENV:
  Broker: Redis
  Celery: 5.2.7
  Python: 3.9.0

👋 Hi All. FYI, since 5.3.2 was released, we encountered multiple celery fork pool workers don't work, so we rolled back to 5.3.1 and everything returned to normal. 
  

Version: 5.3.2

![image](https://github.com/celery/kombu/assets/24605947/3040537c-18a9-4e9f-a624-d52a1cad07ca)

![image](https://github.com/celery/kombu/assets/24605947/3e9a84fa-3552-4b36-9a08-6ab9d0ba2596)

---

Version: 5.3.1
![image](https://github.com/celery/kombu/assets/24605947/d6931100-cc9b-419a-a3e1-ddc69929d460)

![image](https://github.com/celery/kombu/assets/24605947/edf15f23-8df3-4274-85f1-3128afdc2341)

--- 

🤔 We compared the `kombu` version changes, and when we reverted this PR: #1733 in version 5.3.2, all workers worked normally.

cc @auvipy @Nusnus @mfaw @mbierma.

lzgabel:
Hi @mbierma. After our investigation, due to the following code existing in the system:
```python
@celeryd_init.connect
def recover_job(sender=None, conf=None, **kwargs) -> None:
    i = app.control.inspect()
    running_list = []
    for worker_list in i.active().values():
        for task_item in worker_list:
            name = task_item.get('name')
            # do something
            args = task_item.get('args')
            job_id = args[1]
            if job_id:
                running_list.append(job_id)
        logger.info(f'lost job : {running_list}')
```


It will be blocked here [`_brpop_read()`](https://github.com/celery/kombu/pull/1733/commits/3c65006b28687d3924d41036f6bd59835f0b60fc#diff-203e309d100714904255d0af4500dd295e40e4ffedfdc9295ab3b386718ca87eR1084) because of the content fixed by this PR #1733.

mfaw:
hi @lzgabel , I think this is different than my issue. My issue is not fork pool workers but celery workers themselves. I can have 2 celery workers when I'm using rabbitmq as a broker but, when I use kafka as a broker, only one celery worker is working.

thuibr:
[...]

PR #2154, opened 2024-10-14
revert fix: Prevent redis task loss when closing connection while in poll

thuibr:


Nusnus (changes requested):
Hey, thank you for the fix.
Can you please explain what and why is the change please?
Also, please add unit tests.

Lastly, we can mark this as draft for now and ready again after the above is answered & the CI fully passes.

Thanks!

thuibr:
Hi @Nusnus I was trying to debug https://github.com/celery/kombu/issues/1785 but this doesn't do it. I will follow your guidance once I find a fix.

thuibr:
@Nusnus this now reverts #1733 because that causes bugs for #1785.

thuibr:
I also don't really understand the rationale in #1733. Why would completing a `BRPOP` re-enqueue a message in redis?

thuibr:
Alternatively, I could see setting `socket_timeout` to something by default instead of `None`, which also fixes the issue.

```python
app.conf.broker_transport_options = {'socket_timeout': 5}
```

auvipy:
> Alternatively, I could see setting `socket_timeout` to something by default instead of `None`, which also fixes the issue.
> 
> ```python
> app.conf.broker_transport_options = {'socket_timeout': 5}
> ```

can you please elaborate bit more on the conclusion you draw?

thuibr:
> > Alternatively, I could see setting `socket_timeout` to something by default instead of `None`, which also fixes the issue.
> > ```python
> > app.conf.broker_transport_options = {'socket_timeout': 5}
> > ```
> 
> can you please elaborate bit more on the conclusion you draw?

I came to that conclusion empirically, through trial and error. My guess is that the workers are getting stuck somehow and providing a timeout instead of no timeout at all unblocks them. Does that answer your question?

PR #2160, opened 2024-10-17
Fix: restrict google-cloud-pubsub version

haimjether:
- google-cloud-pubsub version in versions larger than 2.20.3 raises the following error:

site-packages/google/protobuf/internal/well_known_types.py", line 443, in FromTimedelta
    raise AttributeError(
AttributeError: Fail to convert to Duration.
Expected a timedelta like object got str: 'str' object has no attribute 'seconds'

  Fix this by restriction of the allowed package versions.


Nusnus (approved):
Awesome, thanks!

PR #2166, opened 2024-10-20
Added a unit test that proves timeout is used when retry policy is specified

thedrow:


Nusnus (approved):
Approved, but the my [comment](https://github.com/celery/kombu/pull/2167#pullrequestreview-2380613552) from the confirm publish PR also applies here.

The test doesn’t check the actual functionality. It only confirms the arguments are passed as expected.
In this case, adding the unit test is great, in opposition to the other PR, so the PR is easily approved. However, please note that a higher-level test is required to confirm that the timeout is actually respected as expected. This unit test is not trustworthy enough.

You will need to test the negative flows to confirm that it works, which means passing the timeout and testing that it behaves as expected, based on the `timeout` value.

PR #2175, opened 2024-10-23
Fix: restrict google protobuf version

haimjether:
- protobuf version in versions larger than 4.25.5 when used by google-cloud-pubsub raises the following error:

site-packages/google/protobuf/internal/well_known_types.py", line 443, in FromTimedelta
    raise AttributeError(
AttributeError: Fail to convert to Duration.
Expected a timedelta like object got str: 'str' object has no attribute 'seconds'

  Fix this by restriction of the allowed package versions
- Added unit test to validate pubsub and protobuf compatibility
- Enabled google-cloud-pubsub package versions bump


Nusnus (approved):
Great work!
Thanks!

PR #2201, opened 2024-12-05
Close connections in case of an exception

MZauchner:
Fixes: https://github.com/celery/celery/issues/9259

### Issue
When an amqp exception occurs while a connection is in use, the connection becomes unusable. Any attempts to use the connection can potentially block indefinitely.  
Currently, such broken connections are placed back into the connection pool, which causes a later `acquire` call on the pool to return a broken connection. 

### Reproducer
See https://github.com/celery/celery/issues/9259
When a celery app is created with the `confirm_publish=True` option, it is possible for a celery task's `apply_async' method to block indefinitely if a broken connection is retrieved from the connection pool.

### Fix
The `ProducerPool::acquire` and `ConnectionPool::acquire` method now return a wrapped `Connection/Producer`, that ensures the connections are properly closed in case of an exception. A reproducer for the original bug was added as an integration test, to ensure that the fix worked. 

### Additional comments
I believe there is a similar issue with the `ChannelPool`, but in this case I think it is not possible to simply close the connection, as other channels acquired from the pool may still be in use.

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -1035,6 +1035,26 @@ class Connection:
 BrokerConnection = Connection
 
 
+class PooledConnection(Connection):
+    """Wraps :class:`kombu.Connection`.
+
+    This wrapper modifies :meth:`kombu.Connection.__exit__` to close the connection
+    in case any exception occurred while the context was active.
+    """
+
+    def __init__(self, pool, **kwargs):
+        self._pool = pool
+        super().__init__(**kwargs)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type is not None and self._pool.limit:
+            self._pool.replace(self)
+        return super().__exit__(exc_type, exc_val, exc_tb)
+
+
 class ConnectionPool(Resource):
     """Pool of connections."""
 
@@ -1046,7 +1066,7 @@ class ConnectionPool(Resource):
         super().__init__(limit=limit)
 
     def new(self):
-        return self.connection.clone()
+        return PooledConnection(self, **dict(self.connection._info(resolve=False)))
 
     def release_resource(self, resource):
         try:
--- a/t/unit/test_connection.py
+++ b/t/unit/test_connection.py
@@ -1045,6 +1045,43 @@ class test_ConnectionPool(ResourceCase):
         with P.acquire_channel() as (conn, channel):
             assert channel is conn.default_channel
 
+    def test_exception_during_connection_use(self):
+        """Tests that connections retrieved from a pool are replaced.
+
+        In case of an exception during usage of an exception, it is required that the
+        connection is 'replaced' (effectively closing the connection) before releasing
+        it back into the pool. This ensures that reconnecting to the broker is required
+        before the next usage.
+        """
+        P = self.create_resource(1)
+
+        # Raising an exception during a network call should cause the cause the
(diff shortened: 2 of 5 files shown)


Nusnus (changes requested):
Can you add some tests to this please?
That confirms the bug fails without the fix on current main.

Thank you! 🙏

MZauchner:
Sure, I'll also have a look at the tests that failed.

Nusnus:
> Sure, I'll also have a look at the tests that failed.

You can run it locally with `tox -e 3.13-unit` (make sure to use a Python 3.13 virtualenv)

Nusnus:
> Sure, I'll also have a look at the tests that failed.

You can run it locally with `tox -e 3.13-unit`.

MZauchner:
@Nusnus Thanks a lot for your tip with the tests! I added an integration test to the py_amqp tests that should fail without the change.
Also reworked it a bit so the public API of the connection and producer pools are not changed.

kombu/connection.py
```diff
@@ -1035,6 +1035,29 @@ def is_evented(self):
...
+
+    def __getattr__(self, attr):
+        return getattr(self._connection, attr)
+
+    def __enter__(self):
+        return self._connection.__enter__()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type is not None and self._pool.limit:
+            self._pool.replace(self._connection)
```
Nusnus, line 1057: The lack of `return` is intentional?

t/unit/test_connection.py
```diff
@@ -1045,6 +1045,31 @@ def test_acquire_channel(self):
         with P.acquire_channel() as (conn, channel):
             assert channel is conn.default_channel
 
+    def test_exception_during_connection_use(self):
```
Nusnus, line 1048: Very cool test!!
Please document what it does.

It will be usable in other use-cases, so it's better this specific test has proper documentation.

Nice job.

P.s
We don't have many good tests for connection loss edge cases (in general), so this instance will be educational 🚀

Nusnus (changes requested):
Left some comments.

Good work 🔥

Nusnus:
> > Sure, I'll also have a look at the tests that failed.
> 
> You can run it locally with `tox -e 3.13-unit`.

See also: `tox -e lint` ;)

kombu/pools.py
MZauchner, line 1: Removed `PooledProducer` completely, as using inheritance probably won't work nicely (I am assuming that `self.Producer` in the connection pool can be a factory method instead of a class.

t/unit/test_connection.py
MZauchner: Thanks for the very quick feedback. I changed the test slightly (mocking replace is not needed).
Instead I am now checking that the connection is closed in case of an exception.

kombu/connection.py
MZauchner: Updated the class to now be a child class of `Connection`. 
The forwarding using `getattr` was not particularly nice anyway and while it passed the unit tests, it caused issues with celery.

auvipy (commented):
looks good

Nusnus (approved):
Approved!
I also checked a patched Kombu vs all of Celery `main`’s CI to make sure we won’t break anything there.
Everything passes.

Good work!

Merge once the CI here goes full green :)

PR #2208, opened 2024-12-20
Provide annotations for `kombu/utils/encoding.py`

Harry-Lees:
Linked Issue: #1511

For this PR, I aimed to annotate the _existing implementation_ of the encoding functions, regardless of what I thought the function's intention is.

There are some functions e.g. `str_to_bytes`, `bytes_to_str`, or `safe_str` which I think may have potentially misleading implementations. I'm not sure whether this is under the scope of this PR, input on this would be appreciated :).

### DIFF
--- a/kombu/utils/encoding.py
+++ b/kombu/utils/encoding.py
@@ -9,15 +9,14 @@ from __future__ import annotations
 
 import sys
 import traceback
-from typing import Any, Union, Literal, overload, TypeVar
-
+from typing import Any, Literal, TypeVar, overload
 
 T = TypeVar("T")
 
 #: safe_str takes encoding from this file by default.
 #: :func:`set_default_encoding_file` can used to set the
 #: default output file.
-default_encoding_file: Union[str, None] = None
+default_encoding_file: str | None = None
 
 
 def set_default_encoding_file(file: str) -> None:
@@ -71,7 +70,7 @@ def from_utf8(s: str, *args: Any, **kwargs: Any) -> str:
     return s
 
 
-def ensure_bytes(s: Union[str, bytes]) -> bytes:
+def ensure_bytes(s: str | bytes) -> bytes:
     """Ensure s is bytes, not str."""
     if not isinstance(s, bytes):
         return str_to_bytes(s)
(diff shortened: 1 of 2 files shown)


kombu/utils/encoding.py
```diff
@@ -9,77 +9,103 @@
...
     if not isinstance(s, (str, bytes)):
         return safe_repr(s, errors)
     return _safe_str(s, errors)
 
 
-def _safe_str(s, errors='replace', file=None):
+def _safe_str(
+    s: object,
+    errors: str = 'replace',
+    file: Any = None
```
Nusnus, line 107: Some places say `Any` and some `object`.
Is there any specific reason for not using the same annotation all over?

Nusnus (changes requested):
See comments

kombu/utils/encoding.py
Harry-Lees: Apologies for the late response.

There are a number of differences between the annotations of `typing.Any` and `object`, but I don't think it _really_ matters in any of the cases I provided. TLDR; I've updated all uses of `object` to `typing.Any`.

For functions like `safe_repr()` (shown below), technically, `__str__()` and `__repr__()` are functions on the `object` class (see [here](https://docs.python.org/3/reference/datamodel.html#object.__str__)), but I can't think of any real-life examples where an object doesn't have a `__repr__()` function except maybe weird metaclasses or C extension modules.

```python
def safe_repr(
    o: Any,
    errors: str = 'replace',
) -> str:
    """Safe form of repr, void of Unicode errors."""
    try:
        return repr(o)
    except Exception:
        return _safe_str(o, errors)
```

I don't know the codebase that well, but I assume the purpose of functions like these are to catch errors raised by the `__repr__()` function rather than to catch an `AttributeError` in the case where an object doesn't have a repr. So does this mean there's value in the type-checker only accepting objects which have a `__repr__()`? Probably not, there's way more that could go wrong at runtime anyway.

auvipy (changes requested):
My thoughts on this type of type only changes. I think we should discourage this type of contributions. Instead we should allow / encourage this type of changes only in real bugfix /feature adding pr

PR #2221, opened 2025-01-08
Allow dots in Azure Service Bus queue names

PieterBlomme:
In the Azure Service Bus transport, dots are currently replaced by dashes.  This should not be the case, since https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftservicebus allows dots in queue names (it's possible that the conventions have changed)

I realise dots inside a queue name are probably not best practice, but in our case the queue names are determined by a third party and out of our control.


auvipy (commented):
is this going to be a breaking change?

PieterBlomme:
Depends on how a "breaking change" is defined I think, but if a breaking change is defined as a change to the public interface, then I don't think so:
- Previously dots were replaced by dashes in the queue name inside the internal implementation
- Having dots in the queue name is already very rare.  It's now allowed by Azure naming rules, but it's still an odd thing to do and probably not best practice.
- Most users are probably unaware of the behaviour.  If you allow Kombu to create new queues as needed, you will notice that the queue name in Azure has dashes instead of dots.  But it has no impact on the client side.  
- If you do not allow Kombu to create queues (in our case we don't have Manage permissions on the service bus) you will need to make sure to create the service bus queue with the actual name (incl dots) instead of the replaced name (with dashes).  Or change the dots to into dashes client side already

So it would only impact users who:
- Pre-create queues in Azure Service Bus
- Have dots in the Kombu queue name, but actually expect dashes in their service bus queue name
I doubt anyone explicitly wants that behaviour.  But if they accidentally rely on it, the easy fix is to change dots to dashes whenever specifying the kombu queue.

Nusnus (dismissed):
Interesting change. Please rename the PR to a meaningful title.
Also, take note we’re in the middle of a (long) release cycle so this change cannot be merged at least until we finish the release.

Thank you for your patience.

PieterBlomme:
I renamed the PR.  No problem if this takes a while to get merged in, we have a workaround on our end

Nusnus:
> I renamed the PR. No problem if this takes a while to get merged in, we have a workaround on our end

Cool. Thank you!

auvipy (approved):

PR #2224, opened 2025-01-10
fix(sqs): don't crash on multiple predefined queues with aws sts session

iBluemind:
Hi,

I discovered a crash that occurs when using predefined queues with AWS STS authentication, as described in [this documentation](https://docs.celeryq.dev/projects/kombu/en/stable/reference/kombu.transport.SQS.html#predefined-queues).

Even though multiple predefined queues are configured like the example below, only one queue is currently usable.
(If you use queue-2 first, queue-1 becomes unavailable afterward.)
```
transport_options = {
    'predefined_queues': {
        'queue-1': {
            'url': 'https://sqs.ap-northeast-2.amazonaws.com/xxx/aaa'
        },
        'queue-2': {
            'url': 'https://sqs.ap-northeast-2.amazonaws.com/xxx/bbb'
        }
    },
    'sts_role_arn': 'arn:aws:iam::<xxx>:role/STSTest'
}
```

The issue arises because the Channel._predefined_queue_clients variable is not updated, which prevents switching to another queue once one has been used.

Thanks


jinsung-kang (approved):
LGTM

devEunjaeYong (approved):
LGTM

PR #2098, opened 2024-08-14
Resolving TypeError, during version unpacking 

couzhei:
This will make sure that the version unpacking doesn't stop connection to some versions of message brokers. For message brokers with normal-style version, say 3.13.6, 
```Pyhon
def version_string_as_tuple(s: str) -> version_info_t:
    """Convert version string to version info tuple."""
    v = _unpack_version(*s.split('.'))
    ...
```
will result in `['3', '13', '6']` list. But for newer releases of rabbitmq, say when you pull docker image using a specified version, in my case `4.0-rc` as discussed in this  [issue](https://github.com/celery/celery/issues/9161#issuecomment-2288083848) related to #2088 it becomes this:
```Python
['4', '0', '0+beta', '2', '6', 'g65c3daf']
```
which causes that `TypeError`.

PR #2225, opened 2025-01-15
Revert "Resolving TypeError, during version unpacking "

auvipy:
Reverts celery/kombu#2098

might need to revisit this


Nusnus (approved):
It seems the original PR was merged without waiting for our proper review, so with additional errors this is indeed prompt for revert.

PR #2226, opened 2025-01-15
SQS: Support Async JSON SQS Protocol & Message Attributes

hfern:
@auvipy :wave: Hi there! There is no ticket for this issue -- I found this bug myself and thought I'd fix it.

---

Adds support for consuming botocore SQS `json` models in the SQS async code. This only affects consumer code.

Several parameters are supplied incorrectly as they have changed in the `query` to `json` botocore change. If you have a newer version of botocore, you will get the `json` service model instead of the old `query` model. I found several places where kombu was generating calls to the new `json` API  (indicated by the x-amz-json-1.0 content-type) but providing parameter names consistent with the `query` protocol.

This PR makes kombu aware of both the `json` and `query` protocol query names and provide them both. When the AWS request is later constructed from the service model, we apply protocol-specific and common parameters. Unfortunately, AWS has not maintained consistency in the parameter names, so I manually reviewed each API call and supplied both.

The only things that broke were compound data types for ReceiveMessage -- those are attribute names. Other things did not break because the code path for post-process message deletion is through the synchronous branch, which uses standard boto and not the async request builder (`kombu.transport.SQS.Channel.basic_ack`). Nevertheless, I fixed all the parameters in the async client.

Logged request after PR (note new usage of `AttributeNames` instead of `AttributeName.1`):

```http
POST https://sqs.us-west-2.amazonaws.com 200
Content-Type: application/x-amz-json-1.0
X-Amz-Target: AmazonSQS.ReceiveMessage
X-Amz-Date: 20250115T190022Z
Authorization: AWS4-HMAC-SHA256 XXX
Content-Length: 177
User-Agent: Mozilla/5.0 (compatible; urllib3)
Accept-Encoding: gzip,deflate

{"MaxNumberOfMessages": 2, "WaitTimeSeconds": 20, "AttributeNames": ["ApproximateReceiveCount"], "QueueUrl": "https://sqs.us-west-2.amazonaws.com/XXX/identity-hunterq"}



200
x-amzn-RequestId: 05a8dffc-0ce3-5cc7-85c0-bffe04e5c2b2
Date: Wed, 15 Jan 2025 19:00:42 GMT
Content-Type: application/x-amz-json-1.0
Content-Length: 2
connection: keep-alive

{}

```

This code supporting both protocols can be cleaned up if Kombu is willing to require `botocore>=1.34.90`, which was released in April 2024 (boto/botocore/pull/3165). It can take quite a while to move forward the boto3/botocore/moto ecosystem, so Kombu may want to give consumers some time before dropping old botocores. Either way, this PR would be required.

---

As I fixed this to get support for AttributeNames back, I also added support for customizing fetched attributes via `transport_options.fetch_message_attributes`. Now you can supply `["All"]` to get better diagnostics from SQS. I added documentation for this.

### DIFF
--- a/kombu/asynchronous/aws/sqs/connection.py
+++ b/kombu/asynchronous/aws/sqs/connection.py
@@ -76,25 +76,29 @@ class AsyncSQSConnection(AsyncAWSQueryConnection):
             **param_payload
         )
 
-    def make_request(self, operation_name, params, queue_url, verb, callback=None):
+    def make_request(self, operation_name, params, queue_url, verb, callback=None, protocol_params=None):
         """Override make_request to support different protocols.
 
-        botocore is soon going to change the default protocol of communicating
+        botocore has changed the default protocol of communicating
         with SQS backend from 'query' to 'json', so we need a special
         implementation of make_request for SQS. More information on this can
         be found in: https://github.com/celery/kombu/pull/1807.
+
+        protocol_params: Optional[dict[str, dict] of per-protocol additional parameters.
+            Supported for the SQS query to json protocol transition.
         """
         signer = self.sqs_connection._request_signer
 
         service_model = self.sqs_connection.meta.service_model
         protocol = service_model.protocol
+        all_params = (params or {}) | protocol_params.get(protocol, {})
 
         if protocol == 'query':
             request = self._create_query_request(
-                operation_name, params, queue_url, verb)
+                operation_name, all_params, queue_url, verb)
         elif protocol == 'json':
             request = self._create_json_request(
-                operation_name, params, queue_url)
+                operation_name, all_params, queue_url)
         else:
             raise Exception(f'Unsupported protocol: {protocol}.')
 
@@ -133,8 +137,12 @@ class AsyncSQSConnection(AsyncAWSQueryConnection):
     def set_queue_attribute(self, queue, attribute, value, callback=None):
         return self.get_status(
             'SetQueueAttribute',
-            {'Attribute.Name': attribute, 'Attribute.Value': value},
+            {},
             queue.id, callback=callback,
+            protocol_params={
+                'json': {'Attributes': {attribute: value}},
+                'query': {'Attribute.Name': attribute, 'Attribute.Value': value},
+            },
         )
 
     def receive_message(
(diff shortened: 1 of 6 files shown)


auvipy (changes requested):
thanks for noticing and handling this! after initial review, my understanding is this PR support both old and new protocols. since this is a substantial change, should this be included in a major version? as it seems mostly compatibility code.

auvipy:
oh yeah right you are.

kombu/asynchronous/aws/sqs/connection.py
```diff
@@ -163,16 +178,21 @@ def delete_message(self, queue, receipt_handle, callback=None):
...
-            prefix = f'DeleteMessageBatchRequestEntry.{i + 1}'
-            params.update({
-                f'{prefix}.Id': m.id,
-                f'{prefix}.ReceiptHandle': m.receipt_handle,
-            })
+        p_params = {
+            'json': {
+                'Entries': [{'Id': m.id, 'ReceiptHandle': m.receipt_handle} for m in messages],
+            },
+            'query': _query_object_encode({
```
auvipy, line 185: may I know why the private method is being used here? is there any better way?

auvipy (approved):
After reviewing the changes, I think we can merge it for v5.5rc release as it keeps both bc and add forward compat as well

auvipy (changes requested):
can you share update on the following please

> The only things that broke were compound data types for ReceiveMessage -- those are attribute names. Other things did not break because the code path for post-process message deletion is through the synchronous branch, which uses standard boto and not the async request builder (kombu.transport.SQS.Channel.basic_ack). Nevertheless, I fixed all the parameters in the async client.

hfern:
@auvipy In hindsight, my message was ambiguous -- sorry!

The "things that broke" I mentioned was what led me to making this PR. They were already broken. This PR does not break anything -- it actually fixes the breakages from earlier.

auvipy (approved):

auvipy:
hi, can you check, if this issue is related to this? https://github.com/celery/kombu/issues/2257

hfern:
@auvipy Sure I will look.

PR #2229, opened 2025-01-18
fix: interpret the ssl_check_hostname as a boolean

### LINKED ISSUE #2210: ssl_check_hostname parameter is not respected for Redis connections
For testing, I use a redis that only accepts TLS connections, with a certificate that does not match the Redis hostname.

My broker is configured is with the following URL: `BROKER_URL = "rediss://:password@localhost:6379/1?ssl_check_hostname=false&ssl_cert_reqs=required&ssl_certfile=./localhost.crt&ssl_keyfile=./localhost.key&ssl_ca_certs=./CA.crt"`.

ssl_check_hostname is left as-is ("false"), and is evaluated to True when a Redis connection is created:
In `kombu/transport/redis.py`, we have 
```python
connparams == {'host': '0.0.0.0', 'port': 55043, 'username': None, 'password': 'interdiode', 'max_connections': 10, 'socket_timeout': None, 'socket_connect_timeout': None, 'socket_keepalive': None, 'socket_keepalive_options': None, 'health_check_interval': 25, 'retry_on_timeout': None, 'ssl_check_hostname': 'false', 'ssl_certfile': 'tools/localhost.crt', 'ssl_keyfile': 'tools/localhost.key', 'ssl_ca_certs': 'tools/CA.crt', 'connection_class': Connection(), 'db': 1}
```

If I patch `kombu/utils/url.py` (line 42) with a boolean interpretation of `ssl_check_hostname`, everything work as expected.

```python
def parse_url(url):
    # type: (str) -> Dict
    """Parse URL into mapping of components."""
    scheme, host, port, user, password, path, query = _parse_url(url)
    if query:
        keys = [key for key in query.keys() if key.startswith('ssl_')]
        for key in keys:
            if key == "ssl_check_hostname":
                query[key] = query[key].lower() != 'false'
            elif key == 'ssl_cert_reqs':
                query[key] = parse_ssl_cert_reqs(query[key])
                if query[key] is None:
                    logger.warning('Defaulting to insecure SSL behaviour.')

            if 'ssl' not in query:
                query['ssl'] = {}

            query['ssl'][key] = query[key]
            del query[key]

    return dict(transport=scheme, hostname=host,
                port=port, userid=user,
                password=password, virtual_host=path, **query)
```

d9pouces:
!close #2210 

ssl_check_hostname is left as-is ("false"), and is evaluated to True when a Redis connection is created.

### DIFF
--- a/kombu/utils/url.py
+++ b/kombu/utils/url.py
@@ -39,7 +39,9 @@ def parse_url(url):
     if query:
         keys = [key for key in query.keys() if key.startswith('ssl_')]
         for key in keys:
-            if key == 'ssl_cert_reqs':
+            if key == "ssl_check_hostname":
+                query[key] = query[key].lower() != 'false'
+            elif key == 'ssl_cert_reqs':
                 query[key] = parse_ssl_cert_reqs(query[key])
                 if query[key] is None:
                     logger.warning('Defaulting to insecure SSL behaviour.')
--- a/requirements/pkgutils.txt
+++ b/requirements/pkgutils.txt
@@ -5,4 +5,4 @@ tox>=4.4.8
 sphinx2rst>=1.0
 bumpversion==0.6.0
 pydocstyle==6.3.0
-mypy==1.14.0
+mypy==1.14.1
--- a/t/integration/test_redis.py
+++ b/t/integration/test_redis.py
@@ -2,10 +2,13 @@ from __future__ import annotations
 
 import os
 import socket
+import ssl
+import traceback
 from time import sleep
 
 import pytest
 import redis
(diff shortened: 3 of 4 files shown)


auvipy (changes requested):
should we also consider adding an integration test for this?

d9pouces:
What would you consider as an integration test?

auvipy:
you can take idea from here https://github.com/celery/kombu/blob/main/t/integration/test_redis.py

Nusnus (dismissed):
Changed to draft until the CI fully passes

auvipy (approved):

Issue #2232, opened 2025-01-27
chore: fully switch to project/packaging configuration to `pyproject.toml`


stegayet:
As of now, `kombu` uses `setup.cfg`, `setup.py`, and `MANIFEST.in` as project/packaging configuration files besides `pyproject.toml`. The idea would be to fully switch to the latter.

Some of those configuration items only needs to be migrated to `pyproject.toml` (`mypy` for instance), for some others, it will require more work ([`flake8` does not support `pyproject.toml`](https://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations) and could be replaced with [`ruff`](https://docs.astral.sh/ruff/)).

Note that this issue does not cover `kombu` package manager/build backend modernization.

Nusnus:
I don’t recall where the other issue is to reference, but there was already a discussion about this topic last year.
Long story short, we want to refine all of the repositories on the same toolchain etc.

Therefore, it is less recommended to upgrade repo by repo. Instead, it is better to start a discussion about the entire toolchain to evaluate (with the community) what stack/config best fits compared to our current toolchain.

You’re welcome to initiate the effort and take the lead @stegayet.
If you’re interested, I’ll describe better what is required and how we should proceed. If not, it will be done as part of our roadmap for 2025, with an undefined ETA.

stegayet:
Hey, thanks for replying, I can indeed start a discussion around the whole toolchain, let me try to have something ready by the end of the week, looks good to you?

Nusnus:
Thank you!
I’ll post here this week some guidelines to focus the effort in the right direction.
You can start by taking a look at `celery`, `kombu`, and `pytest-celery.` Check out the CI (GitHub Actions/Tox), the Makefiles, the doc folder and its own Makefile. You can start comparing in your head how we do stuff just to get a taste of the diff between these repos. 

> I can indeed start a discussion around the whole toolchain

I prefer we maintain the conversation here for now, @stegayet, until we define clear goals and agree on the action items and guidelines.

Lastly, @auvipy and I are using these tools a lot to maintain the org. We will need to be highly involved to ensure any upgrade indeed works for us as well, as the daily clients of the toolchain.

stegayet:
Sorry, I did not have time to start this item, will try to draft a plan by mid March

auvipy:
I want to personally handle this. you can act as reviewer / supervisor

stegayet:
Sure thing 👍

Nusnus:
[...]

PR #2233, opened 2025-01-27
chore(mypy): switch to `pyproject.toml` configuration

stegayet:
Towards https://github.com/celery/kombu/issues/2232.

## Description

This commit switches `mypy` configuration from `setup.cfg` to `pyproject.toml`.

## How this has been tested?

- `pyproject.toml` configuration is correctly read and same as before, cf. `mypy -vvv .` output.
- pre-commit works as before.
- CI jobs are successful.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -142,7 +142,7 @@ commands =
     pydocstyle {toxinidir}/kombu
 
 [testenv:mypy]
-commands = python -m mypy --config-file setup.cfg
+commands = python -m mypy
 
 [testenv:lint]
 allowlist_externals = pre-commit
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -142,7 +142,7 @@ commands =
     pydocstyle {toxinidir}/kombu
 
 [testenv:mypy]
-commands = python -m mypy --config-file setup.cfg
+commands = python -m mypy
```
stegayet, line 145: There is no need to specify `config-file` CLI option as [`mypy` automatically searches for configuration file](https://github.com/celery/kombu/issues/2232).

Nusnus:
https://github.com/celery/kombu/issues/2232#issuecomment-2651956421

PR #2239, opened 2025-01-29
Use `typing.TYPE_CHECKING` instead of `STATICA_HACK`

bswck:
Fun fact, `TYPE_CHECKING` was released in mid 2016 and `STATICA_HACK` was first used here in late 2012.

### DIFF
--- a/kombu/__init__.py
+++ b/kombu/__init__.py
@@ -6,7 +6,7 @@ import os
 import re
 import sys
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


auvipy (commented):
restarted the failing python 3.8 build

auvipy (commented):
seems OK to me but may be we can defer it to 5.5.x release?

bswck:
If you want it, ping me.

PR #2240, opened 2025-01-30
Fix: expiration ttl as timedelta on gcpubsub

j0j1j2:
image

With protobuf version 5.29.3,  ttl value as string is not supported.

### DIFF
--- a/kombu/transport/gcpubsub.py
+++ b/kombu/transport/gcpubsub.py
@@ -49,6 +49,7 @@ from contextlib import suppress
 from os import getpid
 from queue import Empty
 from threading import Lock
+from datetime import timedelta
 from time import monotonic, sleep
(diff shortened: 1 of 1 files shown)


auvipy:
why the PR was closed?

j0j1j2:
I was using poetry and I met this error with protobuf 5.xx. So I created this PR to fix this error. 
but after few minutes later i found that in [`requirements/extras/gcpubsub.txt`](https://github.com/celery/kombu/blob/main/requirements/extras/gcpubsub.txt), the protobuf version is specified to 4.25.5. 
My PR works on protobuf version 5.29.3. but I cannot assure it works on protobuf 4.25.5, So I closed this issue.

auvipy:
we will need to update the version in future versions

PR #2241, opened 2025-02-05
remove Python 3.8 from CI as EOL

auvipy:


ulgens:
@auvipy @Nusnus Is there reason to keep this PR as draft?

PR #2246, opened 2025-02-18
feat(transport): add valkey and valkeys

### LINKED ISSUE #2245: Support for Valkey
[Valkey](https://github.com/valkey-io/valkey) is an open source drop-in replacement for Redis.

Although the scheme `redis://` is compatible, the package `django-valkey` recommends switching to `valkey://`.

If I change the BROKER_URL from `BROKER_URL=redis://127.0.0.1:6379` to `BROKER_URL=valkey://127.0.0.1:6379`, both celery and celery-beat stops working, because they can't resolve `transport`:

```
  File "/***/lib/python3.13/site-packages/kombu/connection.py", line 201, in __init__
    if not get_transport_cls(transport).can_parse_url:
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/***/lib/python3.13/site-packages/kombu/transport/__init__.py", line 91, in get_transport_cls
    _transport_cache[transport] = resolve_transport(transport)
                                  ~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/***/lib/python3.13/site-packages/kombu/transport/__init__.py", line 72, in resolve_transport
    raise KeyError(f'No such transport: {transport}')
KeyError: 'No such transport: valkey'
```

Zerotask:
This PR adds the alternative transport schemes `valkey` and `valkeys`, which is also the recommendation from `django-valkey`.

Fixes #2245 

I didn't update the documentation because I first want to get feedback.
There could be a new page for Valkey or just an addition for the Redis transport here:
https://docs.celeryq.dev/projects/kombu/en/stable/reference/kombu.transport.redis.html

---
Valkey is an open source drop-in replacement for Redis:
> This project was forked from the open source Redis project right before the transition to their new source available licenses.
(https://github.com/valkey-io/valkey)

https://django-valkey.readthedocs.io/en/latest/configure/advanced_configurations.html


Nusnus (changes requested):
See https://github.com/celery/celery/pull/9300

Zerotask:
> See [celery/celery#9300](https://github.com/celery/celery/pull/9300)

Thank you for referencing that. So this is a blocker and it probably needs a new Valkey transport implementation in kombu?

We recently switched from Redis to Valkey and didn't experience any issues with celery. It works as before.

Nusnus:
> > See [celery/celery#9300](https://github.com/celery/celery/pull/9300)
> 
> Thank you for referencing that. So this is a blocker and it probably needs a new Valkey transport implementation in kombu?
> 
> We recently switched from Redis to Valkey and didn't experience any issues with celery. It works as before.

I don't recall all of the details, but the other issue explains where we are right now.

auvipy:
I think we might be able to use same code for two differently named backend named redis / valkey / dragonfly etc

auvipy (approved):
Yes, this makes most sense as we do not need to maintain double code for the same back end

auvipy (commented):
just adding docs and may be should be OK for now

hronecviktor:
Hello, do you have an ETA on releasing this please?

MehrazRumman (changes requested):
Reusing the Redis transport is clearly the right call. Two things block it as written.

**1. `valkeys://` does not actually enable TLS.** The secure-scheme upgrade is a literal string test in `kombu/connection.py:283`:

```python
if transport == 'rediss' and ssl_available and not ssl:
    ...
    ssl = {'ssl_cert_reqs': CERT_NONE}
```

This PR only adds alias entries, so `transport` is `'valkeys'`, the branch never fires, `conninfo.ssl` stays falsy, and `_connparams()` never sets `connection_class = SSLConnection` (`kombu/transport/redis.py:1631-1636`). The result is a cleartext socket with no warning, while the docstring this PR adds advertises `valkeys://` as the secure form. That is worse than the status quo, where the scheme failed loudly as unknown.

The repo already contains the test that proves it — `test_rediss_connection` at `t/unit/transport/test_redis.py:2144` asserts `Connection('rediss://')` yields an `SSLConnection`. The same test body with `valkeys://` fails today.

Fix is one line plus a test:

```python
if transport in ('rediss', 'valkeys') and ssl_available and not ssl:
```

**2. `README.rst` is generated, so those edits will be reverted.** `Makefile:112` rebuilds it from `docs/templates/readme.txt`, which includes `docs/includes/introduction.txt`. That source file is untouched here and still has no Valkey in its keyword line (`:5`), its built-in-support sentence (`:29`), its link targets (`:56`), or its comparison table (`:82`). So `make readme` wipes all four README changes, and because `docs/introduction.rst` includes the same file, the published Getting Started page never mentions Valkey at all. Making the edits in `docs/includes/introduction.txt` and regenerating fixes both, and resolves the README conflict for free.

**Also:** both merge conflicts are the same trivial collision — main inserted a `*pgmq*` row directly after the `*redis*` row in the comparison tables in `README.rst` and `docs/userguide/connections.rst`, which is exactly where this PR inserts `*valkey*`. Keep both rows.

For reference, celery#9300 is a result-backend change and does not block this transport work.

PR #2260, opened 2025-03-15
Revert "SQS: Support Async JSON SQS Protocol & Message Attributes"

auvipy:
to see if it fixes a regression report Reverts celery/kombu#2226

PR #2266, opened 2025-03-19
SQS: Fix Query Protocol Content-Type header

### LINKED ISSUE #2257: UnknownOperationException after upgrading Kombu to 5.5.0
Our docker build process has upgraded kombu from 5.4.2 to 5.5.0, and we noticed that our containers are in a crash loop. We faced the following error:

```
Unrecoverable error: Exception("Request HTTP Error  HTTP 404  Not Found (b'<UnknownOperationException/>\\n')")
```

After pinning down the version to 5.4.2 again, the issue is fixed. We're using SQS, and this is probably the only impacted broker.

hfern:
Fixes #2257, which is from the error introduced by the JSON API patch in #2226.

Related: #2260

The lack of `Content-Type: application/x-www-form-urlencoded` prevented the SQS endpoint from getting the Action parameter from the body. 

I am still investigating exactly how/why that got dropped from my PR, but this is a patch that should unblock kombu consumers.

### DIFF
--- a/kombu/asynchronous/aws/sqs/connection.py
+++ b/kombu/asynchronous/aws/sqs/connection.py
@@ -39,10 +39,13 @@ class AsyncSQSConnection(AsyncAWSQueryConnection):
             params['Action'] = operation
 
         # defaults for non-get
-        param_payload = {'data': params}
+        param_payload = {'data': params, "headers": {}}
         if method.lower() == 'get':
             # query-based opts
             param_payload = {'params': params}
+
+        if method.lower() == 'post':
+            param_payload["headers"]['Content-Type'] = 'application/x-www-form-urlencoded'
 
         return AWSRequest(method=method, url=queue_url, **param_payload)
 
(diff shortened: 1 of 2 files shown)


auvipy:
ah thanks

kombu/asynchronous/aws/sqs/connection.py
```diff
@@ -40,11 +40,15 @@ def _create_query_request(self, operation, params, queue_url, method):
 
         # defaults for non-get
         param_payload = {'data': params}
+        headers = {}
         if method.lower() == 'get':
             # query-based opts
             param_payload = {'params': params}
 
-        return AWSRequest(method=method, url=queue_url, **param_payload)
+        if method.lower() == 'post':
```
auvipy, line 48: good catch

auvipy (commented):
is it possible to improve the test coverage a bit more?

auvipy (approved):

PR #2267, opened 2025-03-19
SQS: More Protocol Tests

hfern:
Adding more tests as requested in #2266.

I extended a few tests to also call `prepare()` and compare the final request attributes before they go over the network. 

This exposed a few issues where botocore does str-to-bytes comparison, which fails the test in pytest. I cleaned that up -- though it's worth noting that botocore is fairly lenient -- you can provide either str, bytes, a dict, or None. But for testing sanity I cleaned it up.

Verified this is working both with old versions of botocore and new versions.


auvipy (approved):

auvipy (approved):
great improvements on existing fix

Nusnus (commented):
Thank you for the attention to details and the great work @hfern !
We appreciate it a lot!

Issue #2258, opened 2025-03-14
Update to Kombu 5.5.0 stopped processing SQS messages


mdrachuk:
Hello. 

Tbh I'm confused about the debug steps for this, because we didn't have any alerts or error logs. Only the actual usage showed that the tasks from the SQS weren't processing anymore.

Downgrade to 5.4.2 solved the issue for now.

Leaving this here for anybody having similar issues to see and maybe comment on details they can find.

mgorven:
I'm seeing the same thing. It's not completely broken, but processing is extremely slow.

Nusnus:
Can you please share a reproduction script or steps to reproduce, I'd like to solve this ASAP.

Thank you.

Nusnus:
Ok, found it.

## Throughput: 0.99 tasks per second with `urllib3`
```console
python stress_test.py
Processed 50 tasks in 50.68 seconds
Throughput: 0.99 tasks per second
```

## Throughput: 55.25 tasks per second with `pycurl`
```console
python stress_test.py
Processed 50 tasks in 0.90 seconds
Throughput: 55.25 tasks per second
```

In PR https://github.com/celery/kombu/pull/2134 we have migrated from `pycurl` to `urllib3`, by @spawn-guy.
Following this change, we’ve also removed `pycurl` from celery in PR https://github.com/celery/celery/pull/9526 by @jmsmkn.


@mgorven 
> Our workers process tasks from four standard SQS queues. Task execution time is fast (~50ms) and we use `task_acks_late=True`. We run with concurrency of eight and `worker_prefetch_multiplier=16`. With Kombu 5.4.2 one instance can process ~100 tasks per second, with 5.5.0 it drops to ~3/s. CPU usage remains low.

By creating a quick `stress_test.py` to reproduce your case, I could easily see the problem using the latest version.
Reverting [Commit 07c8852](https://github.com/celery/kombu/commit/07c8852be56c5e7f9d13bd6c4b57637d205a4234) from kombu and [Commit 9bf0546](https://github.com/celery/celery/commit/9bf05461dc8de9cb88f4279799e90e1dc0688196) from celery was enough to fix the performance, as can be seen in the logs above.

## Celery Worker
```console
export AWS_DEFAULT_REGION=us-east-1
export AWS_ACCESS_KEY_ID=test
export AWS_SECRET_ACCESS_KEY=test
export AWS_ENDPOINT_URL=http://localhost:4566
celery -A myapp worker -l INFO
```
[...]

PR #2269, opened 2025-03-24
Feature: optional pycurl

spawn-guy:
as it was [reported](https://github.com/celery/kombu/issues/2258) `urllib3` might be slow(er) than `pycurl` as an http client. (or even "not working" in 1 unverified report)

this PR brings the curl back. and uses `pycurl` when available, reverting to `urllib3` when not

as the `pycurl` dependency was removed from being required by `sqs` extra module
to use pycurl - users need to explicitly add and install pycurl library on their own.

the last required version in `pip/requirements.txt` format was
```
pycurl>=7.43.0.5; sys_platform != 'win32' and platform_python_implementation=="CPython"
```

### DIFF
--- a/kombu/asynchronous/http/__init__.py
+++ b/kombu/asynchronous/http/__init__.py
@@ -11,6 +11,7 @@ def Client(hub: Hub | None = None, **kwargs: int) -> BaseClient:
     """Create new HTTP client."""
     try:
         import pycurl
+
         from .curl import CurlClient
         return CurlClient(hub, **kwargs)
     except ImportError:
--- a/kombu/asynchronous/http/__init__.py
+++ b/kombu/asynchronous/http/__init__.py
@@ -9,6 +9,13 @@ __all__ = ('Client', 'Headers', 'Response', 'Request', 'get_client')
 
 def Client(hub: Hub | None = None, **kwargs: int) -> BaseClient:
     """Create new HTTP client."""
+    try:
+        import pycurl
+        from .curl import CurlClient
+        return CurlClient(hub, **kwargs)
+    except ImportError:
+        pass
+    # fall back scenario
     from .urllib3_client import Urllib3Client
     return Urllib3Client(hub, **kwargs)
 
--- a/kombu/asynchronous/http/__init__.py
+++ b/kombu/asynchronous/http/__init__.py
@@ -9,14 +9,10 @@ __all__ = ('Client', 'Headers', 'Response', 'Request', 'get_client')
 
 def Client(hub: Hub | None = None, **kwargs: int) -> BaseClient:
     """Create new HTTP client."""
-    try:
-        import pycurl
-
-        from .curl import CurlClient
+    from .curl import CurlClient
+    if CurlClient.Curl is not None:
         return CurlClient(hub, **kwargs)
-    except ImportError:
-        pass
-    # fall back scenario
+
     from .urllib3_client import Urllib3Client
     return Urllib3Client(hub, **kwargs)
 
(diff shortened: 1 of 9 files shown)


Nusnus (commented):
The py-kafka failures are not related

auvipy:
v5.7 or 5.6 for this?

kombu/asynchronous/http/__init__.py
```diff
@@ -9,6 +9,10 @@
 
 def Client(hub: Hub | None = None, **kwargs: int) -> BaseClient:
     """Create new HTTP client."""
```
auvipy, line 11: will this work as expected? we are not using any try except block here, I'm open to better suggestions as well

auvipy (changes requested):

auvipy:
hopefully the kafka builds will pass now

kombu/asynchronous/http/__init__.py
spawn-guy: ima gonna test the code now

spawn-guy: the code seems to work without `pycurl` and no thrown `Exceptions`

auvipy:
kafka is passing now

auvipy:
for a saner approach, should we just revert it first and then you come with your urllib3 patch again cleanly but keeping pycurl as default one?

Nusnus:
@spawn-guy 
> but the reported problem is in the existing code from before my changes

So the slow downs aren't related to the dep changes?
Very interesting

spawn-guy:
> @spawn-guy 
> > but the reported problem is in the existing code from before my changes
> 
> So the slow downs aren't related to the dep changes?
> Very interesting 
> 
> 

I can't yet figure out how to install pycurl on awslinux2023 yet (this is why I picked this migration in the first place) to test both speeds.
But I was posting inconsistent speedtest results for some time now

auvipy:
> @spawn-guy
> 
> > but the reported problem is in the existing code from before my changes
> 
> So the slow downs aren't related to the dep changes? Very interesting

no they are somewhat related. also we had to use sqs internal / private api which was faster, later we moved to saner public api but with performance regression, then again revert and adding new code. and recently moving to new json protocol properly.

kombu/asynchronous/http/__init__.py
auvipy: ok

requirements/pkgutils.txt
```diff
@@ -6,3 +6,4 @@ sphinx2rst>=1.0
 bumpversion==0.6.0
 pydocstyle==6.3.0
 mypy==1.14.1
+types-pycurl>=7.43.0.5; sys_platform != 'win32' and platform_python_implementation=="CPython"
```
auvipy, line 9: what is types-pycurl? did the package name changed?

auvipy (changes requested):

requirements/pkgutils.txt
spawn-guy: Was suggested by mypy. Otherwise mypy ci was failing

auvipy: Weren't we using only pycurl?

spawn-guy: I started to get this error only now, and types-pycurl was suggested. The error was related to mypy build (which I am not very familiar with). Then I saw other types-* are already used from the same repo. And then if pycurl is optional then full pycurl dependency is not required and types-pycurl would be sufficient to pass mypy build. I suppose mypy build does typechecking that is why it errored

spawn-guy:
just tested the code. finally solved my pycurl installation automation on amzn2023

so the code in this PR - WORKS! both when pycurl **is** and **is not** installed.
the reported str-to-bytes error is also fixed.

however,.. i'll post speed tests soon back in the issue ticket https://github.com/celery/kombu/issues/2258

Nusnus (changes requested):
It appears to be working finally, thank you so much @spawn-guy 🙏

I'll review it this week

PR #2274, opened 2025-03-26
Do not pin package tzdata anymore

stollero:
Hello everyone,

I think we can safely upgrade tzdata to newer versions and do not require a version pin, because it is just a timezone database.

This will help other projects upgrading their dependencies without running into dependency conflicts.

Thanks

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,4 +1,4 @@
 amqp>=5.1.1,<6.0.0
 vine==5.1.0
 backports.zoneinfo[tzdata]>=0.2.1; python_version<"3.9"
-tzdata==2025.2; python_version>="3.9"
+tzdata>=2025.2; python_version>="3.9"


auvipy (changes requested):
I'm bit reluctant to allow this right atm. may be we can reach a consensus later. until then lets hold it.

mrcljx:
Would really appreciate if Kombu wasn't dictating dependencies. Best practices for apps and libraries differ.

Using `==2025.2` claims "this library is known to not work with **any other** version" which is quite definitive (same goes for `typing_extensions` btw). Pandas is a good citizen and uses `tzdata>=2022.7` [for example](https://github.com/pandas-dev/pandas/blob/main/pyproject.toml#L33).

mrcljx:
^ Actually, there's an open issue #2277 about the pinning of `typing_extensions`

jamesmyatt:
Pinning exact versions especially for libraries (rather than applications) doesn't work with Python's dependency management. It's a practice that comes from Javascript, which has an entirely different system. See https://iscinumpy.dev/post/bound-version-constraints/.

auvipy:
I will come to this tomorrow

auvipy (approved):
ok

Issue #2273, opened 2025-03-25
Connection Churn in RabbitMQ with Kombu 5.5.0


25schmekles:
Hello,
We're observing an issue we believe was introduced with Kombu 5.5.0 manifesting as connection/channel churn in RabbitMQ. 

## Environment:
Kombu version: 5.5.0 
Broker: RabbitMQ 3.12.7

Observed in a Celery context (but likely stems from Kombu layer)

## Description:
After upgrading to Kombu 5.5.0, RabbitMQ logs and management UI show vastly increased connection/channel create and delete  activity.

Pinning to Kombu 5.4.2 eliminates the excessive churn, suggesting a regression or unintended side effect in 5.5.0.

The recently published 5.5.1 has not yet been tested.

## This behavior leads to:
Unnecessary load on the broker
Excessive logging
Potential side effects for monitoring or scaling tools that track queue/binding activity

I'm happy to include more details as requested, just wanted to get the issue started/logged.

gsmethells:
We are seeing high channel churn since the release of 5.5.0 manifesting as high CPU load.

auvipy:
thanks for the report! any suggestions to reproduce? can you share any screenshots or anything?

25schmekles:
I don't have reproducible steps yet, still working with my app teams to get something more concrete. As soon as I do I'll be sure to post.

The best I can do is show this grafana chart.
When enabled on this broker we can see the connection/channel churn sky rocket. Then they pin to 5.4.2 and repackage/restart their apps and it stops.

I did this with three three separate application teams using celery with kombu 5.5.0 with the same result.

![Image](https://github.com/user-attachments/assets/6fdb3274-6a8b-4fd5-9d31-3d506449f3ef)

rconnaughton:
Here is a datadog graph showing the open & closed connections to rabbitmq before and after reverting to kombu 5.4.2 around midnight:

<img width="1496" alt="Image" src="https://github.com/user-attachments/assets/983aee37-8109-4a65-a0bb-9b09fcdf4414" />

Connections went from the millions to a few thousand.  

It's hard to say what additional information would be helpful in reproducing this; we have a fleet of 30-ish nodes that each host 7 celery services using prefork concurrency (8 workers, no prefetching).  This was during off-peak hours with very little traffic by our standards (we processed something like 60,000 celery tasks during the period of connection churn in the graph).

25schmekles:
[...]

PR #2275, opened 2025-03-26
Attempt to fix retry loop in `maybe_declare()` when broken connection

olii:
This addresses the issue I found after updating Kombu. Commit by commit, I identified 1686c35a063761f10f6391b57b06ee1383ef1259 as a source of problems.

This is just my quick attempt to fix the issue. What I still see as a problem is this convoluted code flow:
* (1) publish is called with retrying,
* (2) publish function tries to declare entities if they have not been declared yet,
* (3) declare is called with retrying too, instead of failing early because of the outer retry logic in publish,
* ---
* Closed channels/connections are passed down the stack as args/kwargs without any possibility to refresh them - logic in `Connection.ensure()` just passes args and kwargs to nested calls even if the object (or entity) itself is given a new connection.

This PR does not contain any tests. I am unable to reproduce it in the isolated environment. It fails only in our tests of our project. Tests must be run in a certain order for this problem to manifest. I believe it must be something with `register_after_fork()`  where methods that close sockets are called. 

I am not aware of edge cases where this patch wouldn't work.  If you do, please let me know, or prepare a systematic fix and we can close this PR.

Here is a similar issue that also describes a large connection churn rate: https://github.com/celery/kombu/issues/2273

### DIFF
--- a/kombu/common.py
+++ b/kombu/common.py
@@ -135,7 +135,9 @@ def _maybe_declare(entity, channel):
 
     _ensure_channel_is_bound(entity, channel)
 
-    if channel is None:
+    if channel is None or channel.connection is None:
+        # If this was called from the `ensure()` method then the channel could have been invalidated
+        # and the correct channel was re-bound to the entity by calling the `entity.revive()` method.
         if not entity.is_bound:
             raise ChannelError(
                 f"channel is None and entity {entity} not bound.")


auvipy (commented):
the changes seems legit and codcov is also showing the changes are already covered. but not sure if that is enough. Also I can see that channel.connection is gracefully handle later in the code base. should we consider some integration tests for this?

olii:
>but not sure if that is enough
 
True, there could be more places where the old channel is forcefully added to the entity. This PR only fixes the immediate issue I had in our code base. 

An integration test would be great, but I can't reproduce it in a minimal example. I still don't know the point the old channel is closed.

Tomasz-Kluczkowski:
I am testing this in our test environment, will report after longer period of time. This is related to my issue reported here:  https://github.com/celery/celery/issues/4895#issuecomment-2754319065

Tomasz-Kluczkowski:
## Test results:
- immediate problem (infinite connection creation loop) is resolved, I was not able to recreate the issue in 4hrs testing round
- other issues appear though and eventually celery workers become affected
  - critical errors on missing file descriptors
  - master process unable to talk to child processes (UP metric) and using sig kill9 
  - 2 out of 5 child processes died this way and never were recovered
- after 4 hrs of testing the celery master process remained intact, it was connecting to rabbitmq but:
  - due to death of 2 celery child processes it was only showing that task was received but still nothing would happen
  - this is though another robustness issue in other part of the stack
  - the immediate issue is resolved thx to the change to kombu.

auvipy (approved):
thanks. this should fix a critical issue

PR #2283, opened 2025-04-04
Move use of typing_extensions into TYPE_CHECKING block

duncanmmacleod:
This PR moves all (one) usage of `typing_extensions` into a `if TYPE_CHECKING` block, so that the `typing-extensions` package is not required at runtime.

This might relate to #2281.

First-time contributor, sorry if I haven't followed the right contributing procedure.


auvipy (approved):
that's a good one!

PR #2284, opened 2025-04-07
Hotfix: pycurl again

spawn-guy:
- remove `urllib3` `http client` (as not ready for production)
- bring back `pycurl` as hard `sqs` dependency
- fix curl client `request.body` `bytes` conversion (as in current code)


auvipy (changes requested):
I want this rebased after the clean revert

spawn-guy:
@auvipy i've been thinking about your request. and i got to a conclusion.
i assume you want me to "create a revert-commit" and then apply the other fixes that would be removed by the "revert commit".

i see no point in doing exactly that. only for the sake of "commit history".
and the thing is - the PR's seem to be ["squash-merged"](https://github.com/celery/kombu/commit/07c8852be56c5e7f9d13bd6c4b57637d205a4234) that means all commits are squashed into 1 (leaving commit messages in the description only). 

so, in the end `.diff` will be exactly the same as in the current PR code. (revert and re-apply, or apply new on top of existing code)

auvipy:
before merging this, we should also consider https://github.com/celery/kombu/pull/2300

spawn-guy:
@auvipy @Nusnus will you be so kind to `rebase` these branches onto `main` instead of `merging` them into?

kombu/asynchronous/http/curl.py
```diff
@@ -252,14 +252,19 @@ def _setup_request(self, curl, request, buffer, headers, _pycurl=pycurl):
             setopt(meth, True)
 
         if request.method in ('POST', 'PUT'):
-            body = request.body.encode('utf-8') if request.body else b''
```
auvipy: you can rebase this so that we can merge this change

auvipy (changes requested):

kombu/asynchronous/http/curl.py
spawn-guy: Id close this one completely and go with the other one with Urllib3.

Shall I?

auvipy: that could be an option, but may be just extracting the pycurl related improvement would make it usable and the urllib3 vr PR easier to review :)

auvipy: extracted https://github.com/celery/kombu/pull/2322

PR #2285, opened 2025-04-07
Typo in pools's documentation

splanquart:
just a typo in documentation about pools.use_global_limit

### DIFF
--- a/docs/userguide/pools.rst
+++ b/docs/userguide/pools.rst
@@ -189,4 +189,4 @@ argument:
 
     from kombu import pools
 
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
hey, thanks for the pr, how can we verify this?

splanquart:
Hello, 
The following code don't work  :

```python
from kombu import pools
pools.use_default_limit
```

in https://github.com/celery/kombu/blob/main/kombu/pools.py you can see that `use_default_limit` don't existe but `use_global_limit` is used like `use_default_limit` in the documentation.

auvipy (approved):

PR #2303, opened 2025-05-28
Update uuid to uuid7

auvipy:


auvipy:
this will need tests update and thorough testing before merge

WilliamDEdwards:
My $0.02: it should still be possible to manually specify UUID4s - to not break the case where pre-generated UUID4s are passed to signatures.

auvipy:
> My $0.02: it should still be possible to manually specify UUID4s - to not break the case where pre-generated UUID4s are passed to signatures.

would you mind elaborate a bit more please? so that I can clearly understand what you wanted to be done here?

auvipy:
OK, thanks a lot!

kombu/utils/uuid.py
```diff
@@ -1,15 +1,34 @@
-"""UUID utilities."""
 from __future__ import annotations
 
-from typing import Callable
-from uuid import UUID, uuid4
+from typing import Literal
+from uuid import uuid4
 
+try:
+    # Python 3.11 or later
```
gvangool:
```suggestion
    # Python 3.14 or later
```

dejlek: I would rather keep that to the latest version supported by PyPy...

gvangool: Sorry, my comment was about that [`uuid.uuid7`](https://docs.python.org/3/library/uuid.html#uuid.uuid7) is only available in Python 3.14.

dejlek: Then I suggest we wait.

PR #2314, opened 2025-06-10
Remove custom `LifoQueue` class conflicting with recent gevent

olii:
This PR improves compatibility of Kombu with gevent. Removes custom `LifoQueue`.
Tradeoff: When shrinking down the pool, the most recently used item is removed first. However, I don't think the pool shrinking is called that much.

In recent gevent the LifoQueue is monkeypatched. In Kombu the type was implemented as deque, while in Python, Gevent, and Eventlet the queue is based on a list.

https://github.com/python/cpython/blob/3.13/Lib/queue.py#L304
https://github.com/Cue/eventlet/blob/master/eventlet/queue.py#L440
https://github.com/gevent/gevent/issues/2114

### DIFF
--- a/kombu/resource.py
+++ b/kombu/resource.py
@@ -3,17 +3,13 @@
 from __future__ import annotations
 
 import os
-from collections import deque
+from contextlib import nullcontext
 from queue import Empty
-from queue import LifoQueue as _LifoQueue
-from typing import TYPE_CHECKING
+from queue import LifoQueue
 
 from . import exceptions
 from .utils.compat import register_after_fork
 from .utils.functional import lazy
-
-if TYPE_CHECKING:
-    from types import TracebackType
 
 
 def _after_fork_cleanup_resource(resource):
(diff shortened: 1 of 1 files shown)


olii:
The changes are already tested in `t/unit/test_connection.py`. Proper testing would require the installation of `gevent`, but I don't think it is worth it in this case, mainly due to monkeypatching the `gevent` requires.

auvipy:
may be we should also update the minimal gevent version for this change?

olii:
Hello, is there anything that needs to be improved in this PR?

For the reference, the custom LifoQueue was added many years ago:
https://github.com/celery/kombu/blob/40ab91a8555721ed3b60ebe252f644f47fc0835b/Changelog.rst#L5233-L5234

auvipy:
would you mind adding some additional tests to check if the changes do not regress?

olii:
Due to the nature of the changes (mostly removes) I can't think of any other test. 
Here are some existing tests that cover the code:
* [`test_acquire_resize_ignore_err_shrink`](https://github.com/celery/kombu/blob/4c9fb4bbab9586a37456b1c0e3df7c717ddd16d5/t/unit/test_connection.py#L870)
* [`test_remove_limit`](https://github.com/celery/kombu/blob/4c9fb4bbab9586a37456b1c0e3df7c717ddd16d5/t/unit/test_pools.py#L222)

PR #2315, opened 2025-06-11
Remove nested query from sqlalchemy _size

ddelange:
before:
```sql
SELECT count(*) AS count_1 
FROM (SELECT kombu_message.id AS kombu_message_id, kombu_message.visible AS kombu_message_visible, kombu_message.timestamp AS kombu_message_timestamp, kombu_message.payload AS kombu_message_payload, kombu_message.version AS kombu_message_version, kombu_message.queue_id AS kombu_message_queue_id 
FROM kombu_message 
WHERE kombu_message.queue_id = $1::INTEGER) AS anon_1
```
after:
```sql
SELECT count(*) AS count_1 
FROM kombu_message 
WHERE kombu_message.queue_id = $1::INTEGER
```

### DIFF
--- a/kombu/transport/sqlalchemy/__init__.py
+++ b/kombu/transport/sqlalchemy/__init__.py
@@ -213,6 +213,7 @@ class Channel(virtual.Channel):
         return (
             self.session.query(self.message_cls)
             .filter(self.message_cls.queue_id == obj.id)
+            .filter(self.message_cls.visible == True)
             .count()
         )
 


auvipy (changes requested):
can we add some tests to verify the change?

kombu/transport/sqlalchemy/__init__.py
```diff
@@ -209,7 +209,12 @@ def _purge(self, queue):
         return count
 
     def _size(self, queue):
-        return self._query_all(queue).count()
+        obj = self._get_or_create(queue)
+        return (
+            self.session.query(func.count(self.message_cls.id))
+            .filter(self.message_cls.queue_id == obj.id)
+            .scalar()
```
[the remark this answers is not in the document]
ddelange: https://github.com/celery/kombu/pull/2315/commits/32836f920f01a1b55bc47d7e8ff1e58f87a7f755

kombu/transport/sqlalchemy/__init__.py
```diff
@@ -209,7 +209,12 @@ def _purge(self, queue):
         return count
 
     def _size(self, queue):
-        return self._query_all(queue).count()
+        obj = self._get_or_create(queue)
+        return (
+            self.session.query(self.message_cls)
+            .filter(self.message_cls.queue_id == obj.id)
+            .count()
+        )
```
ddelange, line 218: should the count maybe exclude visible=false messages by the way? looks like it's only used for `queue_declare_ok_t`?

ddelange:
hi @auvipy :wave:

anything from my side?

kombu/transport/sqlalchemy/__init__.py
Nusnus: Yes please. Invisible msgs should not be counted.

Nusnus (changes requested):
LGTM, just fix the bug so the count won’t include invisible (e.g, consumed) msgs and we’re good for merge.

@auvipy FYI I'm setting this for v5.6. This is a great clean bugfix.

Great PR btw, thank you!
These types of fixes echo loudly; good catch.

kombu/transport/sqlalchemy/__init__.py
ddelange: ready for review

PR #2322, opened 2025-06-21
fix body bytes encoding in curl.py

auvipy:


auvipy:
will fix the lint errors

PR #2329, opened 2025-07-07
Bump redis requirement to <6 to support v5.3

nsoranzo:


nsoranzo:
Thanks @auvipy ! For what is worth, we have been using redis 5.3.0 as a backend for celery in production for https://usegalaxy.org/ (an instance of https://github.com/galaxyproject/galaxy/ ).

PR #2331, opened 2025-07-08
Cycle through host before calling errback

arnaudmorin:
In oslo.messaging error callback (named _recoverable_error_callback), the code is raising an TimeoutError, which is preventing continuation of the retry in some situation.
This quick fix switch the selection of next host (cycle) before calling the error callback. Then, on next ensure_connection, the new host is selected and we expect it to work correctly :)

See: lp-2096926

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -446,11 +446,14 @@ class Connection:
 
         def on_error(exc, intervals, retries, interval=0):
             round = self.completes_cycle(retries)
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
can you add tests for this change proposed?

auvipy:
what about merging the changes of 2 separate pr in a single one?

auvipy (approved):

PR #2339, opened 2025-07-22
Support for SQS fanout via SNS

rlaunch:
This draft PR enhances the existing SNS support by adding fanout support using AWS SNS. Some existing SNS channel methods have been refactored to allow for easier re-use between SNS & SQS services.

As the Channel class was becoming somewhat sprawling, SNS support was added through composition in a subclass and is only initialised upon use of a fanout function.

To avoid breaking changes, 'supports_fanout' has been left as False by default, but can be enabled via transport_options.

TODO:
- Complete documentation
  - Include all required IAM permissions
  - Explain how cleanup works
- Add unit tests for new methods and update existing ones


rlaunch:
Hi @auvipy,

Sure thing, I'll re-raise a new PR

PR #2348, opened 2025-08-10
Add max_prefetch limit to QoS class and related tests

sashu2310:
BUG:
Workers encounter Out Of Memory (OOM) issues when Redis/RabbitMQ queues are flooded with ETA/countdown tasks or failed tasks. The QoS.increment_eventually() method has no configurable limit, allowing prefetch counts to grow unbounded. This results in workers attempting to load thousands of tasks into memory simultaneously, causing memory exhaustion and crashes. When workers crash, tasks return to the queue and the cycle repeats with other workers, creating an infinite loop of OOM crashes across the worker pool.

FIX:

Introduced max_prefetch parameter to QoS.__init__() to limit the maximum prefetch count a worker can hold in memory
Updated increment_eventually() method to enforce the limit, preventing unbounded prefetch count growth
Added comprehensive documentation for the new parameter with usage examples and integration guidance
Implemented thread-safe limit enforcement using existing mutex to ensure concurrent access safety
Maintained full backward compatibility by allow configurable limit as None


auvipy (commented):
the change seems to be related to https://github.com/celery/celery/pull/9853

auvipy (approved):

PR #2367, opened 2025-09-19
Support client_name connection parameter for redis transport

mdalp:
This is the first step to allow passing the client name from the celery configuration.
I'll send a PR when I can to do that.


auvipy (commented):
would like to have more reference for this, can you share?

Nusnus:
> would like to have more reference for this, can you share?

Same 🙏

auvipy (commented):
can we get better example docs where users can check the example to use in their code?

PR #2361, opened 2025-09-15
Test on Python4.14 release candidate 2

cclauss:
Python v3.14 -- October 7th
* https://www.python.org/download/pre-releases
* https://www.python.org/downloads/release/python-3140rc2

PR #2395, opened 2025-10-19
Use compression.zstd (PEP-784)

Rogdham:
Python 3.14 introduces support for Zstandard in the standard library via the [`compression.zstd`](https://docs.python.org/3.14/library/compression.zstd.html) module ([PEP-784](https://peps.python.org/pep-0784/)).

This PR removes the `zstdandard` extra dependency in favor of the native support in Python 3.14+. For older Python versions, the [backport](https://github.com/Rogdham/backports.zstd) is used.

It also has the advantage of supporting free threading.

Related: [Test on Python 3.14 #2361](https://github.com/celery/kombu/pull/2361)

_Full disclosure: I'm the author and maintainer of `backports.zstd`, and the maintainer of `pyzstd` (which code was used as a base for the integration into Python). I also helped with PEP-784 and its integration into CPython._


auvipy (changes requested):
do this need any further changes or improvements?

Rogdham:
As far as I'm concerned it's good to go!

auvipy (approved):

PR #2414, opened 2025-11-24
Simplify requirements given Py3.9+ baseline

jayaddison:
Some nitpicky/pedantic cleanups for the `requirements.txt` files, given that the [baseline installation Python version is Python3.9-or-later](https://github.com/celery/kombu/blob/88225114201a6a0e13f9fb67f49131a250cc7bf7/setup.py#L97).

Follow-up to PR #2241.


jayaddison:
Thank you both, @cclauss @Nusnus

PR #2417, opened 2025-11-25
pre-commit: Type check Python code with ty

cclauss:
https://docs.astral.sh/ty

% `uvx ty check --exclude=kombu/resource.py`
```
All checks passed!
```


cclauss:
Yes.

PR #2423, opened 2025-12-01
fix: credential_provider compatibility issue with redis < 5.3.0

### LINKED ISSUE #2419: credential_provider breaks older redis versions
The new argument credential_provider is breaking connections with redis-py version 4, this change seems a breaking change so perhaps it should have been released in a major version release.

The issue was caused for me while using redis-py 4 and celery 5, possible fixes include pining kombu version to 5.5.4 or upgrading redis-py to a newer version.

alaminopu:
This MR will fix `credential_provider` compatibility issue with lower version (<5.3.0) of redis. only pass `credential_provider` to connection param when its available. 

Close #2419

### DIFF
--- a/t/unit/transport/test_redis.py
+++ b/t/unit/transport/test_redis.py
@@ -452,26 +452,6 @@ class test_Channel:
         # the on_disconnect counter should be incremented
         assert chan.on_disconect_count == 1
 
-    def test_connparams_credential_provider(self):
-        # Case 1: credential_provider is None
-        self.channel.client.transport_options = {}
-        self.channel.connection.client.credential_provider = None
-        self.channel._prepare_virtual_host = Mock(return_value=0)
-        connparams = self.channel._connparams(asynchronous=True)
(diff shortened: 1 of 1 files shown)


kombu/transport/redis.py
```diff
@@ -1174,7 +1174,6 @@ def _connparams(self, asynchronous=False):
             'virtual_host': conninfo.virtual_host,
             'username': conninfo.userid,
             'password': conninfo.password,
-            'credential_provider': conninfo.credential_provider,
```
auvipy, line 1177: won‘t this break existing code?

alaminopu: no, because we conditionally set it inside `_process_credential_provider`. This line was actually breaking lower version of redis which doesn't support `credential_provider`, and we were trying pass `credential_provider` argument with default `None`.

auvipy: got it

PR #2426, opened 2025-12-02
Fix SentinelChannel to pass ACL credentials to master_for()

anthonykuzmich7:
When using Redis Sentinel with ACL authentication (username + password), the SentinelChannel failed to pass credentials to the Redis master connection.

The _connparams() method correctly extracts username and password, but they were only passed to sentinel.Sentinel() (for sentinel authentication), NOT to master_for() (for Redis master authentication).

This fix ensures that ACL credentials are forwarded to master_for() so that Celery workers can connect to Redis Sentinel brokers that use ACL authentication (Redis 6.0+ feature).

Related Issues
https://github.com/celery/celery/issues/6301

### DIFF
--- a/t/unit/transport/test_redis.py
+++ b/t/unit/transport/test_redis.py
@@ -2038,6 +2038,17 @@ class test_RedisSentinel:
 
             connection.channel()
 
+            patched.assert_called_once_with(
+                [
+                    ('localhost', 65532),
+                ],
+                connection_class=ANY, db=0, max_connections=10,
+                min_other_sentinels=0, password='mypassword',
+                sentinel_kwargs=None,
+                socket_connect_timeout=None, socket_keepalive=None,
+                socket_keepalive_options=None, socket_timeout=None,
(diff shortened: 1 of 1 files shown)


kombu/transport/redis.py
```diff
@@ -1477,9 +1477,15 @@ def _sentinel_managed_pool(self, asynchronous=False):
                 "'master_name' transport option must be specified."
             )
 
+        master_kwargs = {
+            k: additional_params[k]
```
auvipy, line 1481: do we need any documentation update for this? or this is just fixing a documented feature?

anthonykuzmich7: Here we are just fixing a documented feature. The Sentinel connection string format `sentinel://[USER:PASSWORD@]...` was already documented in the module, implying that credentials should work for authentication. The bug was that credentials were only passed to sentinel.Sentinel() but not to master_for(), so ACL auth with the Redis master didn't work.

PR #2437, opened 2026-01-02
Implement missing _delete() method of filesystem.Channel

WojciechMigda:
As of now, `Channel` class of `filesystem` transport does not have a `_delete` method. It prevents queues from being removed from an associated exchange. This PR intents to remedy that.


WojciechMigda:
I'll try to open another PR. #2470

PR #2464, opened 2026-02-12
Fix active_queues to preserve queue ordering

### LINKED ISSUE #801: active_queues as set in redis transport does not maintain order
https://github.com/celery/kombu/blob/488ed977e0345eb0f509f020a80bb88f42bb1315/kombu/transport/redis.py#L986-L989

When using the "priority" scheduling cycle with the redis transport, the original order of the list of queues to use is not preserved, due to `_active_queues` being converted from a list to a set and back again.

Is there a reason the `active_queues` property needs to be a set? As far as I can tell, it's only used to check if it's empty, check its length, and to update the queue cycle. Could we change it to just return a list instead? Or if duplicates are an issue, remove them while preserving order?

nairb774:
Change active_queues from a set to an order-preserving list so that priority scheduling cycles respect the intended queue order.

Fixes https://github.com/celery/kombu/issues/801

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -1381,9 +1381,11 @@ class Channel(virtual.Channel):
 
     @property
     def active_queues(self):
-        """Set of queues being consumed from (excluding fanout queues)."""
-        return {queue for queue in self._active_queues
-                if queue not in self.active_fanout_queues}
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
can you please also add unit tests to verify the change?

auvipy (approved):

Issue #2463, opened 2026-02-11
`ignore_errors()` does not catch gevent pool's `ConcurrentObjectUseError` during connection close, causing Celery worker crash


JaeHyuckSa:
Hello!

I use Celery with gevent pool + Django + Amazon MQ (RabbitMQ with TLS/AMQPS).

When a greenlet is blocking on an SSL socket read and another greenlet calls `connection.close()`, gevent raises `ConcurrentObjectUseError`. 

`kombu.common.ignore_errors()` only catches errors in `Transport.connection_errors`, which does not include gevent's `ConcurrentObjectUseError`.  This propagates through `ignore_errors()` and crashes the Celery worker as an "Unrecoverable error".

<details>
<summary>ERROR LOG</summary>

``` bash
Traceback (most recent call last):
  File "gevent/ssl.py", line 439, in read
    return self._sslobj.read(nbytes or 1024)
SSLWantReadError: The operation did not complete (read) (_ssl.c:2710)
Traceback (most recent call last):
  File "celery/worker/worker.py", line 203, in start
    self.blueprint.start(self)
  File "celery/bootsteps.py", line 116, in start
    step.start(parent)
  File "celery/bootsteps.py", line 365, in start
    return self.obj.start()
  File "celery/worker/consumer/consumer.py", line 346, in start
    blueprint.start(self)
  File "celery/bootsteps.py", line 116, in start
    step.start(parent)
  File "celery/worker/consumer/events.py", line 36, in start
    prev = self._close(c)
  File "celery/worker/consumer/events.py", line 62, in _close
    ignore_errors(c, dispatcher.connection.close)
  File "kombu/common.py", line 325, in ignore_errors
    return fun(*args, **kwargs)
  File "kombu/connection.py", line 400, in release
    self._close()
  File "kombu/connection.py", line 366, in _close
    self._do_close_self()
  File "kombu/connection.py", line 359, in _do_close_self
    self.transport.close_connection(self._connection)
  File "kombu/transport/pyamqp.py", line 212, in close_connection
    connection.close()
  File "amqp/connection.py", line 602, in close
    return self.send_method(
  File "amqp/abstract_channel.py", line 79, in send_method
    return self.wait(wait, returns_tuple=returns_tuple)
  File "amqp/abstract_channel.py", line 99, in wait
    self.connection.drain_events(timeout=timeout)
  File "amqp/connection.py", line 526, in drain_events
    while not self.blocking_read(timeout):
[...]

PR #2478, opened 2026-03-05
Fix/2463 gevent concurrent error

ChickenBenny:
## Problem

When using Celery with the **gevent pool** and an **AMQPS (TLS) broker**
(e.g. Amazon MQ / RabbitMQ), a race condition occurs during connection shutdown:

1. **Greenlet A** is blocking on an SSL socket read (waiting for the next AMQP frame).
2. **Greenlet B** calls `connection.close()`, which triggers an SSL `CloseOk` read on the
   **same socket**.
3. gevent detects two greenlets accessing the same socket IO-watcher concurrently and raises
   `ConcurrentObjectUseError` from `gevent.exceptions`.
4. `ignore_errors()` only suppresses exceptions listed in
   `Transport.connection_errors` and `Transport.channel_errors`.
   `ConcurrentObjectUseError` is not in either tuple, so it propagates up and Celery
   treats it as an unrecoverable error, **crashing the worker**.

Reproduction repo: https://github.com/JaeHyuckSa/kombu-gevent-issue  
Issue: #2463

---

## Why the straightforward fix breaks Celery's test suite

Two simpler approaches were tried and discarded:

### Approach 1 — module-level import + eager tuple outside `try`

```python
# kombu/utils/compat.py
try:
    from gevent.exceptions import ConcurrentObjectUseError
except ImportError:
    ConcurrentObjectUseError = None

# kombu/common.py
@contextmanager
def _ignore_errors(conn):
    errors = conn.connection_errors + conn.channel_errors   # ← EAGER
    if ConcurrentObjectUseError:
        errors += (ConcurrentObjectUseError,)
    try:
        yield
    except errors:
        pass
```

**Why it broke Celery CI:**

Celery's `ConsumerStep.shutdown()` calls `ignore_errors(connection, consumer.channel.close)`.
In Celery's test suite the connection is a plain `Mock()`, so
`connection.connection_errors` and `connection.channel_errors` are both `Mock` instances —
not tuples. The line

```python
errors = conn.connection_errors + conn.channel_errors
```

executes **every time** `_ignore_errors` is entered (even when no exception is ever raised),
producing:

```
TypeError: unsupported operand type(s) for +: 'Mock' and 'Mock'
```

This caused a cascade of Celery CI failures across its entire test suite.

### Approach 2 — module-level guard `if 'gevent' in sys.modules:`

```python
# evaluated once, at kombu import time
if 'gevent' in sys.modules:
    from gevent.exceptions import ConcurrentObjectUseError
else:
    ConcurrentObjectUseError = None
```

**Why it never actually worked in production:**

Real applications import kombu **before** calling `gevent.monkey.patch_all()`.
The guard evaluates `False` at import time and permanently sets
`ConcurrentObjectUseError = None`, so the fix silently does nothing for every
standard deployment.

---

### `kombu/utils/compat.py` — lazy function

```python
def get_gevent_concurrent_error():
    """Lazily return gevent's ConcurrentObjectUseError, or None.

    Evaluated at call time (not import time) so that it correctly handles
    the common startup ordering where kombu is imported before gevent's
    monkey-patching takes place.
    """
    if 'gevent' in sys.modules:
        try:
            from gevent.exceptions import ConcurrentObjectUseError
            return ConcurrentObjectUseError
        except ImportError:
            pass
    return None
```

Placed in `kombu/utils/compat.py` — the established module that owns all
optional runtime-environment dependencies (`register_after_fork`,
`detect_environment`, etc.).

### `kombu/common.py` — fully lazy `except` clause

```python
@contextmanager
def _ignore_errors(conn):
    try:
        yield
    except conn.connection_errors + conn.channel_errors + (
        (conc_err,) if (conc_err := get_gevent_concurrent_error()) is not None else ()
    ):
        pass
```

The **entire** tuple expression — including `conn.connection_errors + conn.channel_errors` —
now lives inside the `except` clause.

Python only evaluates `except` expressions when an exception is actually raised.
During a normal (no-exception) shutdown, the mock attributes are never added together,
so `Mock + Mock` is never attempted and the `TypeError` never occurs.

---

## Key properties of the fix

| Property | Detail |
|---|---|
| Zero cost for non-gevent users | `get_gevent_concurrent_error()` returns `None` immediately when gevent is not in `sys.modules` |
| Correct import-order handling | Evaluated at call time, not at kombu import time; works even when kombu is imported before `monkey.patch_all()` |
| No Celery CI breakage | Tuple composition only happens on exception path; plain `Mock()` connections are never touched during normal shutdown |
| Covers both API forms | Both `with ignore_errors(conn):` and `ignore_errors(conn, fun)` go through `_ignore_errors` |

---

## Files changed

| File | Change |
|---|---|
| `kombu/utils/compat.py` | Added `get_gevent_concurrent_error()` lazy function |
| `kombu/common.py` | Imported `get_gevent_concurrent_error`; moved entire `except` tuple inside `except` clause |
| `t/unit/test_common.py` | 4 new tests for `_ignore_errors` / `ignore_errors` |
| `t/unit/utils/test_compat.py` | 4 new tests for `get_gevent_concurrent_error()` |

---

### DIFF
--- a/kombu/common.py
+++ b/kombu/common.py
@@ -16,6 +16,7 @@ from amqp import ChannelError, RecoverableConnectionError
 from .entity import Exchange, Queue
 from .log import get_logger
 from .serialization import registry as serializers
+from .utils.compat import get_gevent_concurrent_error
 from .utils.uuid import uuid
 
 __all__ = ('Broadcast', 'maybe_declare', 'uuid',
@@ -288,7 +289,9 @@ def _ensure_errback(exc, interval):
 def _ignore_errors(conn):
     try:
         yield
-    except conn.connection_errors + conn.channel_errors:
+    except conn.connection_errors + conn.channel_errors + (
+        (conc_err,) if (conc_err := get_gevent_concurrent_error()) is not None else ()
+    ):
         pass
 
 
--- a/kombu/utils/compat.py
+++ b/kombu/utils/compat.py
@@ -22,6 +22,22 @@ except ImportError:  # pragma: no cover
         register_after_fork = None
 
 
+def get_gevent_concurrent_error():
+    """Lazily return gevent's ConcurrentObjectUseError, or None.
+
+    Evaluated at call time (not import time) so that it correctly handles
+    the common startup ordering where kombu is imported before gevent's
+    monkey-patching takes place.
+    """
+    if 'gevent' in sys.modules:
+        try:
+            from gevent.exceptions import ConcurrentObjectUseError
+            return ConcurrentObjectUseError
+        except ImportError:
+            pass
+    return None
+
+
 _environment = None
 
 
--- a/t/unit/test_common.py
+++ b/t/unit/test_common.py
@@ -55,6 +55,53 @@ def test_ignore_errors():
             raise KeyError()
 
 
+def test_ignore_errors_catches_concurrent_object_use_error_when_gevent_available():
+    """ignore_errors() must suppress ConcurrentObjectUseError when gevent is installed."""
+
+    class _ConcurrentObjectUseError(Exception):
+        """Stand-in for gevent.exceptions.ConcurrentObjectUseError."""
+
+    connection = Mock()
+    connection.channel_errors = ()
+    connection.connection_errors = ()
+
+    with patch('kombu.common.get_gevent_concurrent_error', return_value=_ConcurrentObjectUseError):
+        # context-manager form
+        with ignore_errors(connection):
+            raise _ConcurrentObjectUseError()
+
+        # function-call form
+        def raising():
+            raise _ConcurrentObjectUseError()
+
+        ignore_errors(connection, raising)
+
+
+def test_ignore_errors_reraises_when_gevent_not_available():
+
+    class _ConcurrentObjectUseError(Exception):
+        pass
+
+    connection = Mock()
+    connection.channel_errors = ()
+    connection.connection_errors = ()
+
+    with patch('kombu.common.get_gevent_concurrent_error', return_value=None):
+        with pytest.raises(_ConcurrentObjectUseError):
+            with ignore_errors(connection):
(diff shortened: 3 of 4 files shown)


auvipy (changes requested):
thanks for the newer approach. what about slightly modifying the celery tests as well?

auvipy (approved):

kombu/common.py
```diff
@@ -288,7 +289,9 @@ def _ensure_errback(exc, interval):
 def _ignore_errors(conn):
     try:
         yield
-    except conn.connection_errors + conn.channel_errors:
+    except conn.connection_errors + conn.channel_errors + (
+        (conc_err,) if (conc_err := get_gevent_concurrent_error()) is not None else ()
+    ):
```
jayaddison, line 294: This pull request / error handling generally looks good to me.

This few lines of code seem a little inelegant, and I have an idea: would emitting the exception(s), if any, using a generator function allow this code to be clearer?

I think it could become something like:

```python
    except conn.connection_errors + conn.channel_errors + tuple(concurrency_errors()):
```

Where the `concurrency_errors` function, renamed from `get_gevent_concurrent_error`, would be adjusted to something like:

```python
def concurrency_errors():
    """updated docstring here...
    """
    if 'gevent' in sys.modules:
        try:
            from gevent.exceptions import ConcurrentObjectUseError
            yield ConcurrentObjectUseError
        except ImportError:
            pass
```

(no `return` statement required, no walrus-operator or `is not None` magic required -- and the function would be easier to extend to support other concurrency engines like `greenlet`, if relevant)

Edit: fixup: rename the function to the intended name in the code suggestion

auvipy: are you up for it?

jayaddison: Yep, I think so.  I'll try to develop and open a refactor pull request for this soon.

PR #2481, opened 2026-03-06
Fix RabbitMQ 4.x crash: global QoS not supported on classic queues

### LINKED ISSUE #2469: Workers crash on startup with RabbitMQ 4.x — "540 NOT_IMPLEMENTED - queue does not support global qos"
Workers using the `pyamqp` transport crash immediately on
  startup when connecting to RabbitMQ 4.x with:

  540 NOT_IMPLEMENTED - queue 'your-queue' does not support global qos
 

 ### Root Cause

  `qos_semantics_matches_spec` in `kombu/transport/pyamqp.py` (and `librabbitmq.py`)
  only has a special case for RabbitMQ < 3.3:

  ```python
  def qos_semantics_matches_spec(self, connection):
      props = connection.server_properties
      if props.get('product') == 'RabbitMQ':
          return version_string_as_tuple(props['version']) < (3, 3)
      return True
  ```

  This returns False for all RabbitMQ >= 3.3, including 4.x. Celery then calls
  basic.qos(global=True) (channel-wide prefetch), which RabbitMQ 4.0 removed
  entirely from classic queues.

  RabbitMQ 4.0 removed global QoS prefetch on classic queues entirely. See:
  - [Deprecation announcement (2021)](https://www.rabbitmq.com/blog/2021/08/21/4.0-deprecation-announcements)
  - [Classic queues docs](https://www.rabbitmq.com/docs/classic-queues)

  ### Fix

  I'm wondering if this patch would be already enough.

  ```
  def qos_semantics_matches_spec(self, connection):
      props = connection.server_properties
      if props.get('product') == 'RabbitMQ':
          version = version_string_as_tuple(props['version'])
          # RabbitMQ < 3.3: per-channel QoS (return True)
          # RabbitMQ 3.3–3.x: global QoS semantics (return False)
          # RabbitMQ 4.0+: global QoS removed on classic queues (return True)
          return version < (3, 3) or version >= (4, 0)
      return True
  ```

  Environment

  - Kombu: 5.6.x
  - RabbitMQ: 4.0+
  - Celery: 5.6.x

Azd325:
RabbitMQ 4.0 removed support for global QoS prefetch (`basic.qos` with `global=true`) on classic queues. When Kombu connected to a RabbitMQ 4.x broker, `qos_semantics_matches_spec` returned `False` (because the version was not < 3.3), causing Celery to set `global=True` in its QoS call, which RabbitMQ 4.x immediately rejected with:

  540 NOT_IMPLEMENTED - queue does not support global qos

This crashed workers on startup before they could consume any tasks.

Fix the version boundary check in both `pyamqp` and `librabbitmq` transports to return `True` (per-channel QoS) for RabbitMQ >= 4.0:

  Before: version < (3, 3)
  After:  version < (3, 3) or version >= (4, 0)

This restores the correct behavior:
  - RabbitMQ < 3.3:   True  (spec-compliant per-channel QoS)
  - RabbitMQ 3.3–3.x: False (global QoS semantics)
  - RabbitMQ 4.0+:    True  (global QoS removed, per-channel only)

RabbitMQ deprecation announcement (2021):
  https://www.rabbitmq.com/blog/2021/08/21/4.0-deprecation-announcements

RabbitMQ 4.x classic queues docs (global QoS removal):
  https://www.rabbitmq.com/docs/classic-queues

Closes: #2469

### DIFF
--- a/kombu/transport/librabbitmq.py
+++ b/kombu/transport/librabbitmq.py
@@ -175,7 +175,14 @@ class Transport(base.Transport):
             warnings.warn(UserWarning(W_VERSION))
         else:
             if props.get('product') == 'RabbitMQ':
-                return version_string_as_tuple(props['version']) < (3, 3)
+                version_str = props.get('version')
+                if not version_str:
+                    return True
+                version = version_string_as_tuple(version_str)
+                # RabbitMQ < 3.3: per-channel QoS (return True)
+                # RabbitMQ 3.3–3.x: global QoS semantics (return False)
+                # RabbitMQ 4.0+: global QoS removed on classic queues (return True)
+                return version < (3, 3) or version >= (4, 0)
         return True
 
     @property
--- a/kombu/transport/pyamqp.py
+++ b/kombu/transport/pyamqp.py
@@ -224,7 +224,14 @@ class Transport(base.Transport):
     def qos_semantics_matches_spec(self, connection):
         props = connection.server_properties
         if props.get('product') == 'RabbitMQ':
-            return version_string_as_tuple(props['version']) < (3, 3)
+            version_str = props.get('version')
+            if not version_str:
+                return True
+            version = version_string_as_tuple(version_str)
+            # RabbitMQ < 3.3: per-channel QoS (return True)
+            # RabbitMQ 3.3–3.x: global QoS semantics (return False)
+            # RabbitMQ 4.0+: global QoS removed on classic queues (return True)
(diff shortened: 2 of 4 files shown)


auvipy (commented):
we should keep support for both old and new versions for easier migrations

auvipy (changes requested):
also can you improve test coverage?

auvipy (approved):

PR #2488, opened 2026-03-12
Set TIMEOUT and CONNECTTIMEOUT on pycurl handles in CurlClient

jlmorton:
The CurlClient._setup_request() method configures many pycurl options but never sets TIMEOUT or CONNECTTIMEOUT on the curl handle. The Request class already defines request_timeout (30s) and connect_timeout (30s), but these values are never applied to the underlying curl transfer.

Without TIMEOUT, pycurl will wait indefinitely for a response on a socket that remains open but stops delivering data. This happens during network partitions, proxy restarts, or sidecar terminations — the TCP connection stays established (no RST, no FIN) but the remote end stops sending. The in-flight HTTP request hangs forever, which permanently breaks the SQS async polling loop: the promise callback from _get_from_sqs never fires, so _loop1 is never re-invoked and the consumer stops polling for messages. The worker process stays alive (event loop keeps running, liveness probes pass) but never consumes another message.

With TIMEOUT set, pycurl aborts the stalled transfer after request_timeout seconds and returns an error. The existing error handling in _process/_process_pending_requests picks this up, the transport reconnects, and the polling loop resumes normally.


auvipy (changes requested):
dont we need any unit tests for the changes proposed?

PR #2511, opened 2026-04-21
Update Redis version requirements in redis.txt

auvipy:

PR #2512, opened 2026-04-21
pin redis to >=5.3.0,<7.1

simonepozzoli-pix4d:
Celery removed some parameters from the calls to `get_connection` https://github.com/celery/celery/pull/10036. This change was included in Celery 5.6.3, but it broke compatibility with `redis<5.3.0` where those parameters are mandatory.

This PR pins the version of Redis to be >=5.3.0, which is the version where the parameters were made optional and deprecated.


auvipy:
https://github.com/celery/kombu/pull/2511

auvipy:
you can come with a separate pr ti add pyredis 7.4 support

PR #2530, opened 2026-04-27
fix(pidbox): default to exclusive queues for RabbitMQ 4.3.0 compatibility

Nusnus:


Nusnus:
Attempted fix instead: https://github.com/celery/kombu/pull/2531

Issue #379, opened 2014-07-16
[Django] ERROR: Unrecoverable error: KeyError(21,)


nimrodm:
I use in the following apps:
kombu==3.0.15
celery==3.1.11
django-celery==3.1.10

Anyone know the error below?  What should I do in order to solve this issue?

Traceback (most recent call last):

File "/opt/app/virtualenvs/app_core/local/lib/python2.7/site-packages/celery/worker/**init**.py", line 206, in start
  self.blueprint.start(self)
File "/opt/app/virtualenvs/app_core/local/lib/python2.7/site-packages/celery/bootsteps.py", line 123, in start
  step.start(parent)
File "/opt/app/virtualenvs/app_core/local/lib/python2.7/site-packages/celery/bootsteps.py", line 373, in start
  return self.obj.start()
File "/opt/app/virtualenvs/app_core/local/lib/python2.7/site-packages/celery/worker/consumer.py", line 278, in start
  blueprint.start(self)
File "/opt/app/virtualenvs/app_core/local/lib/python2.7/site-packages/celery/bootsteps.py", line 123, in start
  step.start(parent)
File "/opt/app/virtualenvs/app_core/local/lib/python2.7/site-packages/celery/worker/consumer.py", line 569, in start
  replies = I.hello(c.hostname, revoked._data) or {}
File "/opt/app/virtualenvs/app_core/local/lib/python2.7/site-packages/celery/app/control.py", line 111, in hello
  return self._request('hello', from_node=from_node, revoked=revoked)
File "/opt/app/virtualenvs/app_core/local/lib/python2.7/site-packages/celery/app/control.py", line 70, in _request
  timeout=self.timeout, reply=True,
File "/opt/app/virtualenvs/app_core/local/lib/python2.7/site-packages/celery/app/control.py", line 306, in broadcast
  limit, callback, channel=channel,
File "/opt/app/virtualenvs/app_core/local/lib/python2.7/site-packages/kombu/pidbox.py", line 300, in _broadcast
  channel=chan)
File "/opt/app/virtualenvs/app_core/local/lib/python2.7/site-packages/kombu/pidbox.py", line 339, in _collect
  self.connection.drain_events(timeout=timeout)
File "/opt/app/virtualenvs/app_core/local/lib/python2.7/site-packages/kombu/connection.py", line 279, in drain_events
  return self.transport.drain_events(self.connection, **kwargs)
[...]

PR #2561, opened 2026-06-20
Fix KeyError(<fileno>) crash in Redis transport on_readable/handle_event

ronm-cye:
## Summary

`MultiChannelPoller.on_readable()` (and the `ERR` branch of `handle_event()`) look up the firing fd with a bare `self._fd_to_chan[fileno]`. Under connection churn the event loop can report a fd as readable *after* its channel has already been removed from `_fd_to_chan` (connection dropped + reconnected within the same poll cycle; the stale reader on the hub isn't removed in lockstep). The lookup then raises `KeyError(fileno)`.

That exception escapes the event loop and crashes the consumer:

```
Unrecoverable error: KeyError(35)
  File ".../celery/worker/loops.py", line 111, in asynloop
    next(loop)
  File ".../kombu/asynchronous/hub.py", line 377, in create_loop
    cb(*cbargs)
  File ".../kombu/transport/redis.py", line 1394, in on_readable
    self.cycle.on_readable(fileno)
  File ".../kombu/transport/redis.py", line 577, in on_readable
    chan, type = self._fd_to_chan[fileno]
                 ~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 35
```

`KeyError` is not in `connection.connection_errors`, so celery's loop error guard (`hub.propagate_errors`) does not recover from it — the worker dies.

## Reproduction (deterministic, against current `main`)

```python
from kombu.transport.redis import MultiChannelPoller

p = MultiChannelPoller()
# epoll reported fd 35 readable, but its channel was just torn down,
# so 35 is no longer in _fd_to_chan:
p.on_readable(35)   # -> KeyError: 35   (crashes the consumer in production)
```

## Fix

Guard both bare `_fd_to_chan[fileno]` lookups. When the fd is no longer mapped, log at debug and return: the event is for a channel that no longer exists, so there is nothing for this transport to read. Any in-flight message on the dropped connection was never acked and is redelivered by the broker, so no message is lost.

This matches the early-return-on-stale-state pattern already used elsewhere in this transport (e.g. `maybe_check_subclient_health`).

## Context

This is a long-standing, intermittently reported crash that has been closed before without a code fix (hard to reproduce reliably in the field):

- celery/kombu#379 — "Unrecoverable error: KeyError(21,)"
- celery/celery#2827 — "redis backend Unrecoverable error: KeyError(9,)"

The integer in the error is just the OS file descriptor, which is why it varies per occurrence and is unrelated to message payloads.

## Tests

Added `test_on_readable_ignores_unmapped_fd` and `test_handle_event_ignores_unmapped_fd` to `test_MultiChannelPoller`. Both fail on `main` with `KeyError: 35` and pass with this change. Full `t/unit/transport/test_redis.py` suite (153 tests) passes.

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -601,7 +601,10 @@ class MultiChannelPoller:
                     return
 
     def on_readable(self, fileno):
-        chan, type = self._fd_to_chan[fileno]
+        chan_type = self._fd_to_chan.get(fileno)
+        if chan_type is None:
+            return
+        chan, type = chan_type
         if chan.qos.can_consume():
             chan.handlers[type]()
 
@@ -609,7 +612,10 @@ class MultiChannelPoller:
         if event & READ:
             return self.on_readable(fileno), self
         elif event & ERR:
-            chan, type = self._fd_to_chan[fileno]
+            chan_type = self._fd_to_chan.get(fileno)
+            if chan_type is None:
+                return
+            chan, type = chan_type
             chan._poll_error(type)
 
     def get(self, callback, timeout=None):
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -603,18 +603,20 @@ class MultiChannelPoller:
     def on_readable(self, fileno):
         chan_type = self._fd_to_chan.get(fileno)
         if chan_type is None:
+            logger.debug('on_readable: fd %r not mapped, ignoring', fileno)
             return
         chan, type = chan_type
         if chan.qos.can_consume():
             chan.handlers[type]()
 
     def handle_event(self, fileno, event):
+        chan_type = self._fd_to_chan.get(fileno)
+        if chan_type is None:
+            logger.debug('handle_event: fd %r not mapped, ignoring', fileno)
+            return
         if event & READ:
             return self.on_readable(fileno), self
         elif event & ERR:
-            chan_type = self._fd_to_chan.get(fileno)
-            if chan_type is None:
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
please cross check the open suggestions

ronm-cye:
Thanks for the review — all three points addressed in the latest push:

1. **`handle_event()` returning truthy for a stale fd** (good catch): the `READ` branch previously returned `(None, self)` even for an unmapped fd, which would make `MultiChannelPoller.get()`'s `if ret: return` fire and end the drain cycle early. Moved the `_fd_to_chan.get()` guard to the top of `handle_event()`, so a stale fd now returns `None` (falsey) before the `READ` branch and the poll loop continues to `raise Empty()` as usual. Truthiness for real events is unchanged.
2. **Debug logging**: added `logger.debug(...)` on the stale-fd path in both `on_readable()` and `handle_event()` so the churn/race is diagnosable.
3. **Test**: `test_handle_event_ignores_unmapped_fd` now asserts the unmapped-fd case returns a falsey value (`is None`) for both `READ` and `ERR`, not just that it doesn't raise.

Full `t/unit/transport/test_redis.py` suite (153 tests) passes.

ronm-cye:
@auvipy all open suggestions have been cross-checked and addressed in the latest push. Full redis suite (153 tests) green. Would appreciate a re-review when you get a chance 🙏

auvipy (approved):

PR #2567, opened 2026-07-16
Wrong unit and mem leak in curl timeout

DrewRubin:
### Issue
Hi, this is a small change that address two issues:

1. There is no cancellation of previous timers for the curl timeout, so if the SQS server becomes unavailable, it schedules a large number of them.
2. The units are mismatched, which worsens the issue as they are scheduled far in the future.

I reproduced this locally by stopping a running ElasticMQ server and then saw memory growth and high CPU usage. 

```
self.hub.call_later(msecs, self._timeout_check)
```

Leads to 

```
def call_later(self, delay, callback, *args):
    return self.timer.call_after(delay, callback, args)
```

Which calls `call_after` in seconds

```
def call_after(self, secs, fun, args=(), kwargs=None, priority=0):
    kwargs = {} if not kwargs else kwargs
    return self.enter_after(secs, self.Entry(fun, args, kwargs), priority)
```

### Fix

We should probably only keep the latest timer, so use return value instead of creating new, and cancels previous if it exists. Did unit conversion in-function to maintain contract.


auvipy (changes requested):
we need tests to verify this change

DrewRubin:
@auvipy Added a test for the subsequent timer cancelation

auvipy:
any relevant old issues or discussions for this fix?

DrewRubin:
@auvipy Not that I saw, it's just something I noticed and monkey patched in our system, so figured I would try to fix upstream.

auvipy (changes requested):
can you please cross check this pr https://github.com/celery/kombu/pull/2488 and see if we can incorporate the approached followed there?

PR #2572, opened 2026-07-26
Add transport-aware batch publishing with Redis pipelines

shivaam:
## Summary

Add a public, transport-neutral `Producer.batch()` context and implement it for
the Redis transport with `client.pipeline(transaction=False)`.

Normal `Producer.publish()` calls still own serialization, message preparation,
declarations, routing, queue priority selection, queue expiration, and fanout.
Only the final Redis `LPUSH`, `PEXPIRE`, and `PUBLISH` commands are deferred.
Transports that do not implement batching retain immediate publication.

This is motivated by [Apache Airflow #8854](https://github.com/apache/airflow/issues/8854).
Airflow may publish hundreds of Celery tasks in one scheduler iteration. Its
local prototype demonstrated the performance benefit, but had to replace
Kombu's private `_put()` and `_put_fanout()` methods.

## Public API

```python
with producer.batch(max_size=500) as batch:
    producer.publish(message_1, ...)
    producer.publish(message_2, ...)
    batch.flush()  # optional; the context remains usable
    producer.publish(message_3, ...)
```

- A successful outermost exit flushes pending operations.
- Nested contexts share the outer batch and its `max_size`.
- An exception leaving any nested context aborts the batch and discards
  operations that have not already been flushed.
- Empty contexts create no transport session.
- Unsupported transports publish immediately.
- `producer.supports_batch_publish` advertises actual transport support.

## Design decision

The selected design combines public `Producer.batch()` ergonomics with a
transport-owned channel session created by `create_publish_batch()`.

This keeps the API usable by Celery, which already passes a Kombu producer into
`Task.apply_async()`, while leaving broker command ownership with the transport.

Alternatives considered:

1. A producer-only implementation was rejected because the producer does not
   own Redis queue names, priority keys, wire encoding, or fanout topics.
2. Exposing only a channel/transport context was rejected because application
   callers would still need transport knowledge and Celery publishes through
   the producer API.
3. `publish_many(requests)` was rejected because it would duplicate the large
   `Producer.publish()` argument surface and force callers to construct a full
   list instead of using their existing publication loop.

## Failure semantics

Redis pipeline execution is non-transactional. If the response to
`pipeline.execute()` is lost, Redis may have accepted none, some, or all of the
commands.

Kombu raises `BatchPublishError` and does not automatically replay the batch,
including when an individual publication used `retry=True`. Applications may
start a new batch and retry explicitly when at-least-once delivery and possible
duplicates are acceptable.

`max_size` bounds buffered Redis commands. Long-running contexts can use
`batch.flush()` for an elapsed-time boundary, and Redis `socket_timeout` bounds
network blocking.

## Validation

- Full unit suite: `1611 passed, 178 skipped`
- Real Redis 8.6.1 integration suite: `64 passed`
- New real-service coverage runs with and without `global_keyprefix` and
  exercises direct routing, topic routing, priorities/FIFO, and fanout Pub/Sub.
- Pre-commit: all hooks passed, including flake8, isort, codespell, and mypy.
- Celery 5.6.3 smoke test: three `Task.apply_async(..., producer=producer)`
  calls remained deferred inside the context and all three appeared in Redis
  on successful exit.
- Non-strict Sphinx HTML build passed. The existing strict `-W` build still
  fails on unrelated warnings already present on `main`.

The included benchmark publishes 100 messages with 10 ms simulated per-request
latency over seven alternating rounds:

| Mode | Median |
| --- | ---: |
| Ordinary publication | 1.2290 s |
| Redis batch | 0.0164 s |

Observed speedup: **75.1x**.

## Compatibility

- Existing publication is unchanged unless `Producer.batch()` is used.
- Redis TLS and Sentinel share the same batch-capable channel code; their
  inheritance/capability paths have unit coverage.
- Standard Redis is covered with a real server.
- TLS and Sentinel were not tested against real services.
- Redis Cluster is not currently an upstream Kombu transport and is not
  supported or tested here.

### DIFF
--- a/docs/userguide/producers.rst
+++ b/docs/userguide/producers.rst
@@ -107,6 +107,8 @@ empty string, and set the routing key to be the name of the queue:
 Batch publishing
 ----------------
 
+.. versionadded:: 5.7
+
 :meth:`~kombu.Producer.batch` groups normal :meth:`~kombu.Producer.publish`
 calls so a supporting transport can send their final broker operations
 together:
@@ -172,7 +174,11 @@ The standard Redis, Redis TLS, and Redis Sentinel transports share this batch
 implementation. Standard Redis is covered by integration tests. TLS and
 Sentinel use the same channel implementation but are not covered by
 real-service batch integration tests. Redis Cluster is not currently an
-upstream Kombu transport and is not supported or tested by this API.
+upstream Kombu transport and is not supported or tested by this API. Redis
+publishing also remains immediate when the underlying connection pool enables
+automatic retries, including ``retry_on_timeout`` or a custom retry policy,
+because replaying a pipeline after an ambiguous failure can publish duplicate
+messages.
 
 Serialization
 =============
--- a/docs/userguide/producers.rst
+++ b/docs/userguide/producers.rst
@@ -104,6 +104,76 @@ empty string, and set the routing key to be the name of the queue:
     ...     routing_key=task_queue.name,
     ... )
 
+Batch publishing
+----------------
+
+:meth:`~kombu.Producer.batch` groups normal :meth:`~kombu.Producer.publish`
+calls so a supporting transport can send their final broker operations
+together:
+
+.. code-block:: python
+
+    with producer.batch(max_size=500) as batch:
+        producer.publish(
+            {'task': 1},
+            exchange='',
+            routing_key='tasks',
+            declare=[task_queue],
+        )
+        producer.publish(
+            {'task': 2},
+            exchange='',
+            routing_key='tasks',
+        )
+
+        # Optional: send everything accumulated so far and keep batching.
+        batch.flush()
+
+Message preparation, serialization, declarations, and routing still use the
+normal publishing path. Declarations and routing operations whose results are
+needed immediately are not buffered.
+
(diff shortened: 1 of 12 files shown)


docs/userguide/producers.rst
```diff
@@ -104,6 +104,76 @@ empty string, and set the routing key to be the name of the queue:
     ...     routing_key=task_queue.name,
     ... )
 
+Batch publishing
+----------------
```
auvipy, line 108: versionadd 5.7 annotation is needed, please cross check that. also unit test coverage could be improved as well

auvipy (changes requested):

docs/userguide/producers.rst
shivaam: Addressed in fd947e23 after rebasing onto current main. I added the `.. versionadded:: 5.7` annotation and expanded unit coverage for batch lifecycle/cleanup, channel revival, Redis retry-policy fallback, pipeline initialization/command failures, and multi-command size boundaries. I also added real-Redis coverage for `retry_on_timeout` fallback and manual flush/continue. Local results: 1,699 unit tests passed (169 skipped), 68 Redis integration tests passed, and pre-commit, configured mypy, and the HTML docs build passed.

auvipy (approved):

PR #2580, opened 2026-08-14
fix: Queue.from_dict no longer drops queue options

### LINKED ISSUE #1236: Queue.from_dict arguments
I need set max_length of dynamically created queue, but I found that it is impossible, because that argument isapsent in create method (https://github.com/celery/kombu/blob/master/kombu/entity.py#L847-L857).
What a reason don't use all Queue arguments in Queue from_dict method?
It is a typo or is it design?

VXNCXNX:
Closes #1236.

## The problem

`Queue.from_dict` passed a hardcoded list of kwargs to `Queue(...)`, so every option outside that list was silently dropped. A queue round-tripped through `as_dict()` / `from_dict()` came back missing most of its configuration:

```
LOST alias:            'al'  -> None
LOST no_declare:       True  -> None
LOST expires:          30.0  -> None
LOST message_ttl:      20.0  -> None
LOST max_length:       100   -> None
LOST max_length_bytes: 1000  -> None
LOST max_priority:     9     -> None
```

No error, no warning — the queue is just declared without the limits you configured. This is the shape of bug that keeps regressing, since every option added to `Queue.attrs` since the list was written has been forgotten again.

@auvipy said a PR was welcome on this one.

## The fix

Keep the explicit handling for the five fields that have prefix-resolution semantics (`name`, `exchange`, `routing_key`, `durable`, `auto_delete` — these read `queue_durable`/`exchange_durable`/`binding_key` and can't be forwarded blindly), then forward every *other* entry of `Queue.attrs` that is present in `options`.

Options added to `attrs` in future are then preserved automatically, which is what stops this from happening a third time.

Two details worth flagging, since both look like they could bite:

**Omitting absent keys is safe.** The old code always passed `exclusive=options.get('exclusive')` etc., i.e. `None` when absent; the new code omits the key. `Object.__init__` (`kombu/abstract.py:44-52`) treats a `None` value and an absent key identically — both fall through to the class default. I verified all six affected fields: `exclusive` None→False vs omitted→False, `no_ack` False/False, the three `*_arguments` None/None, and `bindings` `set()`/`set()` since `Queue.__init__` does `set(bindings or [])`. No behavior change.

**Type coercion still happens.** `Object.__init__` applies each attr's coercion function itself, so forwarding via `**kwargs` keeps it: `expires='30'` still becomes `30.0`, `no_declare='yes'` still becomes `True`.

## Known adjacent gap, not addressed here

`as_dict(recurse=True)` emits a nested Exchange dict, while `from_dict` expects the flat form (`exchange` as a name string plus `exchange_type`, `exchange_durable`, ...). That asymmetry predates this change and is a different bug; the test bridges it in two lines. Similarly, recursive `bindings` still don't round-trip (dicts are unhashable, so `set(bindings or [])` raises). Happy to take either on separately if you want them.

## Tests

`test_as_dict_from_dict_roundtrip` asserts every option survives. I checked it is load-bearing by stashing only the `entity.py` change and re-running it:

```
>   assert getattr(q2, attr) == getattr(q, attr), attr
E   AssertionError: alias
E   assert None == 'fooalias'
FAILED t/unit/test_entity.py::test_Queue::test_as_dict_from_dict_roundtrip
```

`test_from_dict_does_not_override_defaults` is a pure regression guard (it passes on unpatched code) pinning that an absent key doesn't overwrite a class default with `None`.

```
baseline t/unit:  1321 passed, 185 skipped
final    t/unit:  1323 passed, 185 skipped
flake8:           exit 0
```

### DIFF
--- a/kombu/entity.py
+++ b/kombu/entity.py
@@ -833,6 +833,12 @@ class Queue(MaybeChannelBound):
             expiring_queue = False
         return not expiring_queue and not self.auto_delete
 
+    #: Attributes handled explicitly by :meth:`from_dict`, every other
+    #: attribute in :attr:`attrs` is forwarded to :class:`Queue` as-is.
+    _from_dict_explicit_attrs = frozenset({
+        'name', 'exchange', 'routing_key', 'durable', 'auto_delete',
+    })
+
     @classmethod
     def from_dict(cls, queue, **options):
         binding_key = options.get('binding_key') or options.get('routing_key')
@@ -854,10 +860,6 @@ class Queue(MaybeChannelBound):
             q_auto_delete = options.get('auto_delete')
 
         e_arguments = options.get('exchange_arguments')
-        q_arguments = options.get('queue_arguments')
-        b_arguments = options.get('binding_arguments')
-        c_arguments = options.get('consumer_arguments')
-        bindings = options.get('bindings')
 
         exchange = Exchange(options.get('exchange'),
                             type=options.get('exchange_type'),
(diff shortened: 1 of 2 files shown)


auvipy (commented):
just to be sure, this change do not introduce any backward incompatibility, right?

auvipy (approved):

PR #2590, opened 2026-08-27
Redis: drop half-open pub/sub connections the subclient health check can't see

goingforstudying-ctrl:
Been following celery/celery#10502, where workers on managed Redis (ElastiCache/Valkey over TLS) silently stop consuming fanout messages and stop answering `inspect ping` after running for a while. Process alive, broker reachable, no traceback, restart fixes it. kokhlo narrowed it down in that thread: the pub/sub connection issues no commands while idle, it just waits to become readable, so if the TCP path dies silently (failover, NAT connection-table turnover) nothing on it ever times out.

#2498 added the `maybe_check_subclient_health` timer that calls redis-py's `check_health()`, which is the right idea, but I don't think it actually catches this case. `PubSub.check_health()` only *sends* the PING:

```python
if conn.health_check_interval and time.monotonic() > conn.next_health_check:
    conn.send_command("PING", self.HEALTH_CHECK_MESSAGE, check_health=False)
    self.health_check_response_counter += 1
```

On a half-open socket the write lands in the kernel send buffer and succeeds, so no error is ever raised and kombu never finds out. The timer fires every interval, the counter ticks up, and the worker stays catatonic until someone restarts it. (redis-py decrements the counter when a PONG gets parsed, so on a healthy connection it hovers at 0/1.)

So this makes the timer look at the other side of the exchange: once `health_check_response_counter` reaches 2, meaning two full health-check intervals passed with no PONG, the connection is treated as dead and dropped. The existing reconnect machinery takes it from there: `_on_connection_disconnect` prunes the fd, next `on_poll_start` re-registers and resubscribes. With the default `health_check_interval` of 25s that bounds detection to ~50s instead of riding on TCP keepalive defaults (7200s on Linux) or a manual container restart.

Two things I went back and forth on:

Threshold of 2, not 1. One outstanding PING can just mean the event loop was stalled for an interval (long callback, GC pause); two in a row means the path is gone. False reconnects on busy workers seemed worse than one extra interval of detection time.

And resetting the counter when we drop. `clean_health_check_responses()` only drains responses that actually arrive, so a counter inflated by the dead connection survives the reconnect and would pin us at the threshold forever, reconnect loop. Zeroing it mirrors what `PubSub.reset()` does.

If `health_check_interval` is 0 redis-py never sends pings, the counter never increments, and the new check stays inert. Older redis-py without the counter attribute falls back to the current send-only behavior.

Tests: new cases in `t/unit/transport/test_redis.py` covering the drop at threshold, sub-threshold pass-through, the counter reset, the missing-attribute fallback, and a stale counter with no connection object. Also ran the scenario against a real `redis.client.PubSub` with a fake half-open connection to make sure the counter interop works as expected, not just with mocks. Full `t/unit` passes (1152 passed, 186 skipped; the SQS/azure/gcpubsub collection errors in my env are missing optional deps, unrelated).

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -128,6 +128,11 @@ DEFAULT_DB = 0
 
 DEFAULT_HEALTH_CHECK_INTERVAL = 25
 
+#: Unanswered subclient PINGs tolerated before the pub/sub connection is
+#: treated as half-open and dropped.  Two full intervals must pass without
+#: a PONG, so a stalled event loop alone cannot trigger a reconnect.
+SUBCLIENT_MAX_MISSED_HEALTH_CHECKS = 2
+
 #: How often (in seconds) to flush pending streaming re-authentication
 #: tokens onto the long-lived BRPOP and pub/sub connections.  See
 #: :meth:`Channel.maybe_reauth` for why this is needed.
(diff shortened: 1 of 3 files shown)


auvipy (commented):
thanks for finding it out and handling it.... it would be great if you can also find a way to add some integration tests

goingforstudying-ctrl:
Added integration tests in the latest commit (t/integration/test_redis.py, test_RedisSubclientHealthCheck).

Two scenarios against a real Redis server: a healthy subclient survives the health check and keeps delivering fanout messages, and a connection pinned at the missed-PONG threshold gets dropped, resets its counter, and the next poll cycle reconnects, resubscribes, and receives fanout messages again. Both pass locally with redis-server 7.0 (72 integration tests in the file green, plus the 214 redis unit tests).

PR #2592, opened 2026-08-30
fix(redis): treat stale-fd KeyError as a connection error

### LINKED ISSUE #2582: redis transport: KeyError on stale fd in `_fd_to_chan` after a broker restart kills the worker
## What happens

Restarting the redis/valkey broker under a settled celery worker kills the worker every time.
The worker reconnects successfully, then dies with an unrecoverable error the moment anything
polls the hub:

```
[celery.worker] CRITICAL: Unrecoverable error: KeyError(39)
  File ".../celery/worker/worker.py", line 203, in start
    self.blueprint.start(self)
  ...
  File ".../celery/worker/consumer/mingle.py", line 54, in send_hello
    replies = inspect.hello(c.hostname, our_revoked._data) or {}
  File ".../celery/app/control.py", line 785, in broadcast
    return self.mailbox(conn)._broadcast(
  File ".../kombu/pidbox.py", line 398, in _collect
    self.connection.drain_events(timeout=timeout)
  File ".../kombu/transport/virtual/base.py", line 997, in drain_events
    get(self._deliver, timeout=timeout)
  File ".../kombu/transport/redis.py", line 601, in get
    ret = self.handle_event(fileno, event)
  File ".../kombu/transport/redis.py", line 583, in handle_event
    return self.on_readable(fileno), self
  File ".../kombu/transport/redis.py", line 577, in on_readable
    chan, type = self._fd_to_chan[fileno]
                 ~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 39
```

A file descriptor is still registered with the poller after being removed from `_fd_to_chan`,
so the readable callback dereferences a key that is gone.

## It is not specific to mingle

Running with `--without-mingle` does not avoid it — the identical `KeyError` then arrives from
the main event loop instead, on a different fd:

```
  File ".../celery/worker/loops.py", line 96, in asynloop
  File ".../kombu/asynchronous/hub.py", line 377, in create_loop
  File ".../kombu/transport/redis.py", line 1394, in on_readable
  File ".../kombu/transport/redis.py", line 577, in on_readable
KeyError: 35
```

Whatever polls the hub first after the broker returns hits the stale descriptor. Disabling
bootsteps only relocates the trigger.

## Reproduction

Roughly 60 seconds, 100% reproducible for us across every attempt.

```bash
# worker must be fully settled first — see the confound below
sleep 60
docker compose restart valkey      # or: redis-cli shutdown / systemctl restart redis
sleep 30
# worker has exited 1
```

**Confound worth stating:** a worker bounced within ~10s of its own start *survives* — the
startup retry path is tolerant. Only a settled worker reproduces this. An early test of ours
looked like a fix had worked when the worker had simply never died.

## Environment

| | |
|---|---|
| celery | 5.6.2 |
| kombu | **5.6.2** (latest at time of writing) |
| redis-py | 7.1.0 |
| billiard | 4.2.4 |
| Python | 3.12 |
| Broker | valkey 8 (redis protocol), `redis://` transport |
| Platform | Linux aarch64, containers |

Worker launched as:

```
celery -A <app> worker -n <name> -l info -Q <9 queues> -E --max-tasks-per-child 1
```

## Things that do not fix it

- **`broker_channel_error_retry = True`** — verified applied (`app.conf.broker_channel_error_retry`
  reads `True` in the running worker). No effect, which makes sense: `KeyError` is neither a
  `connection_error` nor a `channel_error`, so neither retry class covers it.
- **`--without-gossip`** — no effect.
- **`--without-mingle`** — no effect, as above.

Not memory pressure: 24 GB total with 18 GB available, no kernel OOM events, container not
OOM-killed, no memory limit set.

## Why it is easy to miss

The failure is invisible in any application whose logging config disables celery's loggers.
Ours did (a Django `LOGGING` dict with `disable_existing_loggers: True` and no `celery`
entry), and the worker died with a **completely empty log** for months of runtime. The
traceback above only appeared after explicitly declaring `celery.worker` as a logger. Worth
knowing when triaging reports of "the worker just disappears".

## Suggested direction

Treat a missing key in `_fd_to_chan` as a stale-registration signal rather than a fatal error —
unregister the fd from the poller and continue, since the channel it belonged to is already
gone. Related historically: #379 (`Unrecoverable error: KeyError(21,)`), closed.

We are happy to test a patch against the reproduction above.

kokhlo:
## Summary

A broker restart under a settled worker still kills the consumer fatally (#2582) even after the #2561 poller guard: the traceback there shows a KeyError escaping the event loop, and KeyError is not in the redis transport's connection_errors, so celery's loop error guard cannot recover — it re-raises as an unrecoverable error and the worker dies instead of reconnecting.

This adds KeyError to the redis transport connection_errors so any stale-fd KeyError that still escapes the poller guard is treated as a connection error: the consumer reconnects and the unacked messages are redelivered by the broker, same as any other connection loss.

Two layers now cover #2582:
- the #2561 guard (main, unreleased) makes the common path a debug log,
- this change makes anything the guard misses recoverable instead of fatal.

Note for the reporter: the traceback in #2582 is from 5.6.2, where the #2561 guard is absent — upgrading once a release lands will also address the direct crash.

Fixes #2582


kokhlo:
Both points fixed: the connection_errors tuple now closes correctly (d9e5e36), and the regression test resolves the transport the way its sibling already does — `redis.Transport.connection_errors` at module level (f65e4f4a). Syntax re-verified on both files before pushing.

auvipy (approved):

t/unit/transport/test_redis.py
```diff
@@ -2921,6 +2921,11 @@ def test_handle_event(self):
 
         p.handle_event(13, ~(redis.READ | redis.ERR))
 
+
```
auvipy:
```suggestion
```

PR #2602, opened 2026-09-02
Add GitHub Actions workflow for Python package publishing

auvipy:


tavallaie:
Do you need any help with trusted publisher? we have a very good CI/CD on [pgmq-py](https://github.com/pgmq/pgmq-py) 
maybe it can help you

auvipy:
I will improve it soon, may be early next week. no we do not have issues with trusted publishers. I am just wrapping other remaining things for upcoming v5.7 pre releases

PR #2628, opened 2026-09-07
fix(redis): replace deprecated ZREVRANGEBYSCORE with ZRANGE BYSCORE REV

### LINKED ISSUE #2050: Updating deprecated zrevrangebyscore
Due to the new licensing issues with Redis, I'm currently testing out Microsoft Garnet as an open source caching alternative for the Celery result backend (which uses kombu for messaging) of Apache Superset. However, as Microsoft Garnet is quite new, it does not support deprecated Redis commands like "zrevrangebyscore", which is used in kombu/transport/redis.py and finally leads to the crashing of Garnet because it cannot handle the command. Are there any plans to upgrade the command to "zrange" as stated in https://redis.io/docs/latest/commands/zrevrangebyscore in the future?

MehrazRumman:
Closes #2050

 This is a fresh implementation on current `main` that also covers `global_keyprefix`, which
the earlier attempt did not, and resolves the open review questions.

### Problem

`QoS.restore_visible()` called `ZREVRANGEBYSCORE`, which Redis deprecated in
6.2.0. Redis-compatible servers such as Microsoft Garnet do not implement it,
so restoring unacked messages crashed the broker there.

### Fix

* `restore_visible()` now issues the documented replacement,
  `ZRANGE <key> <ceil> 0 BYSCORE REV [LIMIT <offset> <num>] WITHSCORES`.
  The `(ceil, 0)` bound order is intentionally unchanged: with `REV`, Redis
  treats the first bound as the highest score, so it maps 1:1 onto the old
  `ZREVRANGEBYSCORE <key> <ceil> 0` call.
* `GlobalKeyPrefixMixin.PREFIXED_SIMPLE_COMMANDS` lists `ZRANGE` in place of
  `ZREVRANGEBYSCORE`. Without this, `global_keyprefix` users would have had
  `restore_visible` query an unprefixed `unacked_index` key and never restore
  anything.

### Why the bound order is correct

Verified against a live Redis 8.6.2 server:
`ZRANGE key ceil 0 BYSCORE REV LIMIT 0 10 WITHSCORES` returns exactly the
same rows, in the same order, as
`ZREVRANGEBYSCORE key ceil 0 LIMIT 0 10 WITHSCORES`, with and without
`LIMIT`. The inverted `(0, ceil)` order returns an empty list, which would
silently disable restoration. The wire arguments redis-py emits are:

    ('ZREVRANGEBYSCORE', key, ceil, 0, 'LIMIT', 0, 10, 'WITHSCORES')
    ('ZRANGE', key, ceil, 0, 'BYSCORE', 'REV', 'LIMIT', 0, 10, 'WITHSCORES')

### Tests

* `t/unit/transport/test_redis.py`
  * three existing `restore_visible` tests now mock `zrange`
  * `test_qos_restore_visible_uses_zrange_byscore_rev`: patches `time()` and
    asserts the full call, key and `(ceil, 0)` bounds included, and that
    `zrevrangebyscore` is never called
  * `test_qos_restore_visible_without_limit`: no `LIMIT` when `num=None`
    (the `unacked_restore_limit` default); redis-py rejects `offset`
    without `num`
  * `test_global_keyprefix_restore_visible`: the command reaching redis-py
    is `('ZRANGE', 'foo_unacked_index', ceil, 0, 'BYSCORE', 'REV', 'LIMIT',
    0, 10, 'WITHSCORES')`
* `t/integration/test_redis.py`
  * `test_RedisRestoreVisible`: publishes a message, fetches it without
    acking, confirms nothing is restored inside `visibility_timeout`, then
    confirms the message is restored with the `redelivered` header and
    removed from the unacked hash and index once the timeout elapses. Runs
    with and without `global_keyprefix`, and uses private unacked keys so it
    cannot disturb messages other tests hold unacked under `-n auto`.

Verified locally: the six changed/new unit tests fail on `main` and pass
with the fix; `t/unit/transport/test_redis.py` passes in full; the
integration test passes against Redis 8.6.2; flake8, pydocstyle and isort
are clean.

### Compatibility note

`ZRANGE ... BYSCORE REV` requires Redis 6.2.0 (February 2021) or a
compatible server that implements it (Valkey, KeyDB 6.3+, Dragonfly,
Garnet). CI already runs integration tests against the latest `redis`
image. Kombu does not currently document a minimum Redis server version;
happy to add one to the transport docs if maintainers want it stated.


MehrazRumman:
@auvipy you can review it !

auvipy:
Minimum supported redis py version is 5.3.1 now. But we can bump it a bit for upcoming v5.7 release of kombu and celery

kombu/transport/redis.py
```diff
@@ -472,9 +472,16 @@ def restore_visible(self, start=0, num=10, interval=10):
...
+                    # not implemented by every Redis-compatible server
+                    # (#2050); ``ZRANGE ... BYSCORE REV`` is the documented
+                    # replacement.  With ``REV`` the first bound is the
+                    # highest score, so ``(ceil, 0)`` keeps the same order
+                    # and returns the same rows as the old command.
+                    visible = client.zrange(
                         self.unacked_index_key, ceil, 0,
-                        start=num and start, num=num, withscores=True)
+                        desc=True, byscore=True,
+                        offset=num and start, num=num, withscores=True)
```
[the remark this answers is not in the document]
auvipy: @MehrazRumman maybe you can update the redis.Version as well?

auvipy (approved):

kombu/transport/redis.py
```diff
@@ -1758,9 +1767,10 @@ def _get_pool(self, asynchronous=False):
         return redis.ConnectionPool(**params)
 
     def _get_client(self):
-        if redis.VERSION < (3, 2, 0):
+        # Keep in sync with requirements/extras/redis.txt.
+        if redis.VERSION < (5, 3, 1):
```
auvipy:
```suggestion
        if redis.VERSION < (6, 1, 0):
```

kombu/transport/redis.py
```diff
@@ -1758,9 +1767,10 @@ def _get_pool(self, asynchronous=False):
...
             raise VersionMismatch(
-                'Redis transport requires redis-py versions 3.2.0 or later. '
+                'Redis transport requires redis-py versions 5.3.1 or later. '
```
auvipy:
```suggestion
                'Redis transport requires redis-py versions 6.1.0 or later. '
```

t/unit/transport/test_redis.py
```diff
@@ -1671,6 +1725,14 @@ def test_get_client(self):
             R.VERSION = (2, 4, 0)
             with pytest.raises(VersionMismatch):
                 redis.Channel._get_client(self.channel)
+
+            # The floor matches requirements/extras/redis.txt: anything
+            # below 5.3.1 is refused, 5.3.1 itself is accepted.
```
auvipy:
```suggestion
            # below 6.1.0 is refused, 6.1.0 itself is accepted.
```

t/unit/transport/test_redis.py
```diff
@@ -1671,6 +1725,14 @@ def test_get_client(self):
...
+            R.VERSION = (5, 3, 0)
+            with pytest.raises(VersionMismatch, match='5.3.1 or later'):
```
auvipy:
```suggestion
            with pytest.raises(VersionMismatch, match='6.1.0 or later'):
```

t/unit/transport/test_redis.py
```diff
@@ -1671,6 +1725,14 @@ def test_get_client(self):
...
+                redis.Channel._get_client(self.channel)
+            R.VERSION = (5, 3, 1)
```
auvipy:
```suggestion
            R.VERSION = (6, 1, 0)
```

auvipy (changes requested):

MehrazRumman:
Thanks for the suggentions @auvipy !

auvipy (approved):

PR #2630, opened 2026-09-08
Fix Redis consumer registration cleanup after disconnect

### LINKED ISSUE #2629: Redis consumer stops receiving after reconnect before the next poll tick
I can reproduce a Redis consumer losing its event-loop registration on current `main` (`fddc84af8bfe7a316c92b1c3771b4d991dbf8537`), with Python 3.12 and redis-py 6.4.0. This is a follow-up to #2492.

If the Redis connection reconnects before the next poll tick, the consumer remains marked as registered but has no reader mapping. Messages stay queued. The disconnect callback also evaluates the cached `subclient` property on channels that have never used PubSub.

We found this while investigating a Celery 5.6.3 worker that stopped receiving tasks but still answered inspection requests. That worker used a backport of #2492. The reproduction below establishes the transport bug; it does not establish which network event triggered our production incident.

## Environment

- Linux x86_64, Python 3.12.10
- Kombu `main` at `fddc84af8bfe7a316c92b1c3771b4d991dbf8537` (the source still reports version 5.6.2)
- redis-py 6.4.0, Redis server 7.0.15
- amqp 5.3.1, vine 5.1.0
- The reproducer uses Kombu directly; no Celery worker, Django application, TLS, or broker credentials are needed.

## Reproduction

The [regression test](https://github.com/celery/kombu/blob/ce86970a66f9c2452b12e9f8b37b1d4f3348d2db/t/integration/test_redis.py) uses a real Redis server and the Kombu hub:

1. Create a consumer and register it with the event loop.
2. Run the first poll tick to register the BRPOP socket.
3. Call `disconnect()` and then `connect()` on that Redis connection before another poll tick.
4. Publish a message through a separate connection and poll the hub for up to five seconds.

Expected: the next poll tick registers the reconnected socket and the consumer receives the message.

Actual on `main`: no message arrives. `_fd_to_chan` is empty, while `_chan_to_sock` still contains the registration. `_client_registered()` therefore returns true and prevents re-registration. An unused `subclient` is also created during the disconnect scan.

The test repeats this sequence three times and runs both with and without `global_keyprefix`. Both cases fail at the first delivery on unmodified `main`; both pass with our correction. It forces reconnect-before-tick ordering explicitly, rather than injecting a retry inside redis-py's `send_command()`.

<details>
<summary>Commands to run the regression test against unpatched main</summary>

Use a disposable checkout and broker. Start Redis in one terminal:

```sh
redis-server --bind 127.0.0.1 --port 16379 --save "" --appendonly no
```

In another terminal:

```sh
git clone https://github.com/celery/kombu.git kombu-reconnect-repro
cd kombu-reconnect-repro
git checkout fddc84af8bfe7a316c92b1c3771b4d991dbf8537
git fetch origin refs/pull/2630/head
# Import only the test; keep the upstream transport implementation unchanged.
git restore --source=ce86970a66f9c2452b12e9f8b37b1d4f3348d2db -- t/integration/test_redis.py
python3.12 -m venv .venv
.venv/bin/python -m pip install -e . redis==6.4.0 pytest==8.4.2
REDIS_HOST=127.0.0.1 REDIS_6379_TCP=16379 .venv/bin/python -m pytest -q -E redis \
  t/integration/test_redis.py::test_event_loop_consumes_after_connection_reconnect
```

Both cases fail at `assert received == list(range(sequence + 1))`: `received` is `[]`, expected `[0]`.

Apply only the transport correction and repeat the same pytest command:

```sh
git restore --source=ce86970a66f9c2452b12e9f8b37b1d4f3348d2db -- kombu/transport/redis.py
```

Both cases pass. Stop the disposable Redis process with Ctrl-C when finished.

</details>

## Related reports

This follows #2492. I also checked #2582 and its fixes (#2561 and #2592): current `main` already includes those changes, but the delivery stall still reproduces. The missing-reader guards prevent a stale-fd exception; they do not clear the stale `_chan_to_sock` registration. The older PubSub report #1365 used a different reproduction and was closed after #1378.

## Suggested correction

Use the existing `(channel, client, mode)` registrations to identify the disconnected connection. Remove its entries from both indexes and remove its hub readers. Keep `on_poll_start` registered and preserve the channel's other BRPOP/LISTEN connection. This also avoids evaluating cached client properties during cleanup.

[The patch against current `main`](https://github.com/celery/kombu/pull/2630) is ready for review and passes 290 Redis/Hub unit tests plus both real Redis reconnect cases.

Could this be included in an official release, with a 5.6.x backport if that branch is still accepting fixes? We'd like to move off our temporary fork once a release passes the reconnect test.

PavelPancocha:
A Redis connection can reconnect before the next event-loop tick and leave its consumer unable to receive messages. Disconnect cleanup removes `_fd_to_chan` entries but leaves `_chan_to_sock` populated, so `_client_registered()` skips registration of the new socket. The cleanup scan can also create an unused PubSub client by evaluating the cached `subclient` property.

Identify the disconnected connection through the existing `(channel, client, mode)` registrations. Remove both registration indexes and the corresponding hub readers, while retaining `on_poll_start` and unrelated BRPOP/LISTEN registrations.

Replace the old disconnect tests with assertions against real poller state. Add a Redis integration test that reconnects before the next poll tick, verifies message delivery over three cycles with and without a global key prefix, and checks normal-close cleanup.

This follows up on #2492. The patch is based on current `main` and has no package version change.

Validation on Python 3.12 with redis-py 6.4.0:

- All nine new disconnect unit cases fail before the fix and pass after it.
- Redis transport and Hub unit suites: 290 passed.
- Real Redis integration: both prefix variants fail at the first delivery before the fix; both pass three reconnect/delivery cycles and close cleanup after it.
- All configured pre-commit hooks passed, including mypy.

The integration test forces reconnect-before-tick ordering. It does not simulate every redis-py retry path or TLS failure.

Closes #2629.


t/unit/transport/test_redis.py
```diff
@@ -1817,259 +1817,6 @@ def test_register_with_event_loop__non_positive_interval_falls_back(
             call(25, transport.cycle.maybe_check_subclient_health),
         ])
 
-    @pytest.mark.parametrize('fds', [{12: 'LISTEN', 13: 'BRPOP'}, {}])
```
auvipy, line 1820: why so many existing unit tests have to be removed?

kombu/transport/redis.py
```diff
@@ -1880,63 +1880,19 @@ def register_with_event_loop(self, connection, loop):
         on_readable = self.on_readable
 
         def _on_disconnect(connection):
-            if connection._sock:
```
auvipy, line 1913: this seems too risky to me to delete so much code.... but lets see

auvipy:
@antonio-mello-ai can you please check this pr?

auvipy:
@PavelPancocha can you please review a redis related pr https://github.com/celery/kombu/pull/2631

antonio-mello-ai:
**Português**

Oi auvipy, verifiquei aqui e essa PR faz sentido. Ela melhora o que eu tinha feito, preservando a correção original.

Fiz uma comparação em memória das duas versões: essa limpa os registros e remove o reader sem criar um cliente PubSub desnecessário. Também confirmei que os dois novos testes de integração Redis passaram no CI.

---

**English version translated using AI**

Hi auvipy, I checked this PR and it makes sense. It improves on what I had done while preserving the original fix.

I compared the two versions in memory: this one clears the registrations and removes the reader without creating an unnecessary PubSub client. I also confirmed that the two new Redis integration tests passed in CI.

t/unit/transport/test_redis.py
PavelPancocha: I replaced the repeated setup with a shared fixture. There are still nine unit-test cases, now checking both registration maps. The socket-specific branches are gone because cleanup no longer reads the socket. I did lose the explicit separate-channel case; that is worth keeping.

kombu/transport/redis.py
PavelPancocha: The change is limited to the disconnect callback. The old code leaves `_chan_to_sock` behind, so after reconnecting Kombu can think the consumer is still registered. The replacement clears both maps and removes the reader, while keeping `on_poll_start`. The real Redis test fails before the fix and passes after it, through three reconnects with and without a key prefix.

PavelPancocha:
Thanks @antonio-mello-ai for checking. Your original fix stays intact.

@auvipy I also reviewed #2631 and left [a comment about the Sentinel cleanup](https://github.com/celery/kombu/pull/2631#issuecomment-5589141796).

auvipy (approved):

auvipy:
To be extra cautious, you could add **one more safety test** that old tests didn't cover:

```python
def test_disconnect_with_both_client_and_subclient_same_connection():
    """Edge case: channel has both client and subclient on same disconnected connection"""
    client = self.channel.client
    subclient = self.channel.__dict__['subclient'] = Client()
    
    # Force both to use the same connection pool
    subclient.connection = client.connection
    
    self.cycle._register(self.channel, client, 'BRPOP')
    self.cycle._register(self.channel, subclient, 'LISTEN')
    
    # Disconnect the shared connection
    client.connection._sock = None
    self.cycle._on_connection_disconnect(client.connection)
    
    # Both registrations should be removed
    assert not self.cycle._chan_to_sock
    assert not self.cycle._fd_to_chan
```

This edge case isn't explicitly tested in the old or new suite, and it validates the most complex scenario the fix handles. or something that is little bit better then this...

PavelPancocha:
Added coverage for shared connections, repeated disconnects, and separate channels. The reconnect integration test now covers PubSub too. I also made the Sentinel test compatible with older redis-py versions. The full unit suite and all six integration suites passed locally on Python 3.12.

auvipy:
pyredis 6.1.0+ compatibility is enough for now

auvipy (approved):

auvipy:
the pypy failures are not related to this pr

PR #2631, opened 2026-09-08
transport/redis: close Sentinel node connections in _disconnect_pools (fixes #1108)

### LINKED ISSUE #1108: socket leak in redis-sentinel
I have been experiencing an issue where kombu-sentinel support is leaking the socket opened to redis-sentinel nodes.
Kombu-sentinel keeps reconnecting the redis sentinel nodes and leaving the connected sockets behind, If there is network glitch or some other known-unknown issue which is preventing kombu-sentinel to connect to redis master node 
 
Reproduction: when celery is configured with different "mastername" in comparison to what redis-sentinel setups are configured with.  
For ex: Redis sentinel setup has master as "masternode" but the celery broker is configured as "masternode1". This way kombu-sentinel support will not be able to find the redis master node and keep reconnecting and leaving the connected sentinel sockets behind. 

Redis sentinel will reach to "max connected client" eventually and the complete system will be dead.

jexbow:
## Description

Fixes #1108: when a Sentinel node goes down (restart, network partition), kombu's redis-sentinel transport never closes the sockets it opened to the Sentinel nodes themselves, leaving connections stuck in `CLOSE_WAIT` until the file-descriptor limit is reached.

Root cause: `SentinelChannel._sentinel_managed_pool()` creates a `redis.Sentinel(...)` instance but never keeps a reference to it. `Channel._disconnect_pools()` only calls `pool.disconnect()` on the master/slave `SentinelConnectionPool`; the per-node connection pools owned by the `Sentinel` instance are unreachable and can never be closed.

redis-py >= 8.1.0 now provides `Sentinel.close()` / `Sentinel.aclose()` (redis/redis-py#4184, released 2026-07-30), so the closing API we need is available.

## Changes

- `Channel.__init__`: initialize `self._sentinel_manager` and `self._async_sentinel_manager` to `None`.
- `SentinelChannel._sentinel_managed_pool()`: keep a reference to the created `Sentinel` instance (async / sync respectively).
- `Channel._disconnect_pools()`: after disconnecting the master/slave pools, also close the Sentinel-node connections:
  - `manager.close()` when the attribute exists (guarded by `hasattr`);
  - as a best effort, schedule `async_manager.aclose()` on the running event loop (the redis transport has no async teardown hook yet);
  - references are cleared so repeated calls are idempotent.
- Tests: save-reference assertion, `close()` called exactly once and idempotent, safe no-op for redis-py < 8.1.0 (no `close` attribute).

## Compatibility

- kombu's redis transport requires redis-py >= 6.1.0.
- `Sentinel.close()` / `aclose()` only exist since redis-py 8.1.0 (PR redis/redis-py#4184) — all calls are feature-detected with `hasattr()`, so older redis-py versions keep the previous behaviour (no node-connection teardown, no exceptions).

## Testing

- `pytest t/unit/transport/test_redis.py` — 241 passed (33 Sentinel-related tests, including the 3 new ones).
- Manual reproduction of #1108: stop a sentinel node, repeat publish/consume, confirm no `CLOSE_WAIT` accumulation.

### DIFF
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -1053,22 +1053,17 @@ class Channel(virtual.Channel):
 
         self._sentinel_manager = self._async_sentinel_manager = None
 
-        if manager is not None and hasattr(manager, 'close'):
-            # redis-py >= 8.1.0 provides Sentinel.close()
-            # (redis/redis-py#4184); older versions have no such API and
-            # keep the previous behaviour.
-            manager.close()
-
-        if async_manager is not None and hasattr(async_manager, 'aclose'):
-            # ``redis.asyncio.sentinel.Sentinel.aclose`` is a coroutine
-            # and kombu's redis transport has no async teardown hook yet,
-            # so schedule it on the currently running loop as a
-            # best-effort cleanup.
-            try:
-                import asyncio
-                asyncio.get_running_loop().create_task(async_manager.aclose())
-            except RuntimeError:  # pragma: no cover - no running loop
-                pass
+        for sentinel_manager in (manager, async_manager):
+            if sentinel_manager is not None and hasattr(sentinel_manager, 'close'):
+                # redis-py >= 8.1.0 provides Sentinel.close()
+                # (redis/redis-py#4184); older versions have no such API
+                # and keep the previous behaviour.  Both managers hold the
+                # same kind of synchronous ``redis.sentinel.Sentinel``:
+                # ``asynchronous=True`` only switches the master/slave
+                # connection class via ``_connparams`` and never creates a
+                # ``redis.asyncio`` Sentinel, so ``close()`` is the API to
+                # use for both (see review on celery/kombu#2631).
+                sentinel_manager.close()
 
     def _on_connection_disconnect(self, connection):
         if self._in_poll is connection:
--- a/kombu/transport/redis.py
+++ b/kombu/transport/redis.py
@@ -984,6 +984,11 @@ class Channel(virtual.Channel):
         if not self.ack_emulation:  # disable visibility timeout
             self.QoS = virtual.QoS
         self._registered = False
+        #: redis.Sentinel instance(s) created by :meth:`_sentinel_managed_pool`,
+        #: kept so their per-sentinel-node connections can be closed again in
+        #: :meth:`_disconnect_pools` (celery/kombu#1108).
+        self._sentinel_manager = None
+        self._async_sentinel_manager = None
         self._expires = {}
         self._queue_cycle = cycle_by_name(self.queue_order_strategy)()
         self.Client = self._get_client()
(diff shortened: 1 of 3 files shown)


PavelPancocha:
@auvipy I had a look at #2631 and found one thing I'd fix before merging.

[`_sentinel_managed_pool()`](https://github.com/celery/kombu/blob/bb6c2ca4f5cf8bc37d70d7ab40978b8458645936/kombu/transport/redis.py#L2122-L2134) creates a synchronous `redis.sentinel.Sentinel` for both pools. `asynchronous=True` here is for Kombu's event loop; it doesn't create a `redis.asyncio` client. These managers have `close()`, not `aclose()`, so the current check skips cleanup of `_async_sentinel_manager`, which is used by `channel.client`.

I checked `_disconnect_pools()` with real Sentinel instances from redis-py 8.1.0. The sync manager's `close()` was called once. The consumer manager's `close()` wasn't called at all. All 241 Redis unit tests pass, so they don't catch this case.

Could we use the guarded `close()` call for both managers? A regression test with real Sentinel instances, or mocks limited to the actual Sentinel API, should catch this.

This is separate from #2630. This PR closes Sentinel discovery connections; ours fixes stale consumer socket registrations after reconnecting. We use a direct Redis connection, so I haven't tested a live Sentinel failover.

jexbow:
@pavelpancocha You are right — thank you for the careful review.

I went through the code again and can confirm your analysis: `_sentinel_managed_pool()` always builds the plain synchronous `redis.sentinel.Sentinel`. The `asynchronous` flag is only forwarded into `_connparams()`, where it swaps the master/slave connection class (and wires disconnects to kombu's own poller); it never creates a `redis.asyncio` object. So the "async" manager stored on the channel is a synchronous `Sentinel`, which only has `close()` (redis-py >= 8.1.0) and no `aclose()`. My previous `hasattr(async_manager, 'aclose')` branch could therefore never fire, and the sentinel-node connections were silently skipped during cleanup — the exact leak celery/kombu#1108 is about.

I have updated the fix so `_disconnect_pools()` closes both managers with `close()`:

```python
for sentinel_manager in (manager, async_manager):
    if sentinel_manager is not None and hasattr(sentinel_manager, 'close'):
        # redis-py >= 8.1.0 provides Sentinel.close()
        # (redis/redis-py#4184); older versions have no such API
        # and keep the previous behaviour.  Both managers hold the
        # same kind of synchronous redis.sentinel.Sentinel:
        # asynchronous=True only switches the master/slave connection
        # class via _connparams and never creates a redis.asyncio
        # Sentinel, so close() is the API to use for both.
        sentinel_manager.close()
```

I also added a regression test that mocks the manager with `spec=redis.sentinel.Sentinel` (the real API, which has `close()` but no `aclose()`) and asserts the async manager is closed exactly once — a regression back to `aclose()` would fail this test.

Test results: `242 passed` on `t/unit/transport/test_redis.py` (previously 241), including all existing Sentinel tests; negative check confirmed the new test fails against the old `aclose()` branch.

auvipy (changes requested):
what about adding additional integration test to verify this?

jexbow:
@auvipy Thanks for the review — integration test added.

I added `test_SentinelManagerClose` in `t/integration/test_redis.py`. The integration environment runs no Sentinel service, so the test patches `redis.sentinel.Sentinel` with a sentinel-like class whose `master_for` returns a real `redis.Redis` connection to the in-container Redis (enabling full channel setup and lifecycle), while exposing `close()` so we can assert both `_sentinel_manager` and `_async_sentinel_manager` are closed exactly once during `_disconnect_pools()`. This covers the fix in the integration suite without requiring a live Sentinel cluster.

Regarding a **full Sentinel integration test** (with real failover, node kill, and `CLOSE_WAIT` verification): I've prepared an offline manual verification plan using docker-compose with a 2-node Sentinel cluster. The blocker for automating this in CI is that kombu's current `integration-redis` environment only provisions a plain Redis container (`[docker:redis]` in `tox.ini`). Adding a Sentinel service would need:

1. `[docker:sentinel]` stanza in `tox.ini` with healthcheck and port mapping
2. Sentinel-to-Redis network orchestration (sentinels must see the master)
3. Fault injection plumbing (container kill / network partition) — inherently flaky on shared CI runners

I verified the fix works correctly in this offline setup (no `CLOSE_WAIT` after `connection.close()` with the fix; sockets linger without it). I'm happy to open a follow-up PR that adds automated Sentinel integration tests once we agree on the CI infrastructure approach — I didn't want to bloat the current bug-fix PR with tox/workflow changes.

Please let me know if the patch-based integration test is sufficient for 5.7.0, or if you'd prefer the full Sentinel CI setup to land together with this PR.

auvipy:
if you want you can start a different sentinel docker service in tox in this pr as well

auvipy (approved):

auvipy (approved):

PR #2640, opened 2026-09-10
fix(redis): catch ValueError in _on_connection_disconnect

jexbow:
## Problem

`redis-py >= 5.x` unconditionally sets `connection._sock = None` inside `disconnect()`. After that, kombu's `MultiChannelPoller._on_connection_disconnect()` calls `poller.unregister(connection._sock)` with `None`.

`kombu/utils/eventio.py:201` (`fileno(None)`) raises `ValueError: file descriptor is None`. The existing `except (AttributeError, TypeError)` in `_on_connection_disconnect` does **not** catch `ValueError`, so the exception propagates and short-circuits the `sentinel_manager.close()` cleanup added in PR #2631.

## Impact

- In redis-py 8.1.0 (and 7.4.1, 5.0.8 — all versions behave the same way), the async pool disconnect path crashes before reaching PR #2631's sentinel cleanup.
- Sentinel connections leak on every master failover + channel close.
- Users see repeated `ValueError: file descriptor is None` when using Redis Sentinel with Celery/Kombu.

## Fix

Add `ValueError` to the exception tuple so the disconnect callback swallows the harmless error and allows the rest of the cleanup (including `sentinel_manager.close()`) to proceed.

## Verification

A minimal sentinel environment was built and tested against:
- Baseline (no PR #2631): sentinel connections monotonically accumulate (`EST=4→7`).
- With PR #2631 + this patch: `Sentinel.close()` called 12× over repeated publish/failover cycles, zero connections left behind.
- 238 existing unit tests in `t/unit/transport/test_redis.py` all pass after this change.

Refs: kombu issue #1108, kombu PR #2631


auvipy (changes requested):
the codsov is showing zero coverage. can you please cross check if it is a false positive or not?

auvipy:
also can you please cross check the unit tests here if they create any conflict with this pr https://github.com/celery/kombu/pull/2630/changes

auvipy:
please cross check and fix the merge conflicts. that pr rewrites the unit tests so you might have to adjust as well

jexbow:
Thanks for the review @auvipy.

I just checked and PR #2630 ("Fix Redis consumer registration cleanup after disconnect") has already been merged into main, which rewrote the disconnect cleanup logic and includes the  handling that this PR was adding.

Since upstream already covers the fix, I'll close this PR to avoid duplication.

jexbow:
Thanks for the review @auvipy.

I just checked and PR #2630 ("Fix Redis consumer registration cleanup after disconnect") has already been merged into main. It rewrote the disconnect cleanup logic and includes the ValueError handling in _on_connection_disconnect that this PR was adding, along with comprehensive tests.

Since upstream already covers the fix, I'll close this PR to avoid duplication.

auvipy:
great! thanks

jexbow:
Hi @auvipy — reopening this PR. My earlier close was based on the assumption that #2630 already covered this fix, but testing the merged main showed that is **not** the case.

**Why #2630 doesn't cover this path:**
- #2630 rewrote the local `_on_disconnect` inside `Transport.register_with_event_loop`, which is assigned as `cycle._on_connection_disconnect = _on_disconnect`. It does **not** touch `MultiChannelPoller._on_connection_disconnect` itself (`kombu/transport/redis.py`).
- When `async_pool.disconnect()` runs (before `sentinel_manager.close()` in the sentinel cleanup added by #2631), redis-py >= 5.x clears `connection._sock` to `None` and the **class-level** `MultiChannelPoller._on_connection_disconnect` calls `poller.unregister(None)`, which raises `ValueError: file descriptor is None`.

**Verified end-to-end** with celery + kombu + sentinel (failover and sentinel restart scenarios):
- On merged main without a workaround, that `ValueError` propagates **before** `sentinel_manager.close()` is reached, so the #2631 fix is short-circuited and sentinel connections are never closed.
- With the one-line `ValueError` added to the except tuple, cleanup runs and remaining connections drop to zero.

This PR is now merged with the latest main (conflict with #2630 resolved), with the unit test rewritten to target `MultiChannelPoller._on_connection_disconnect` directly so it actually covers the changed line. 245 unit tests in `test_redis.py` pass.

auvipy:
ok great

auvipy (approved):

PR #2642, opened 2026-09-10
Added checks to reject URLs containing dotted schemes

### LINKED ISSUE #2641: Untrusted URLs with dotted scheme can force import of arbitrary module
## Description

A behaviour in Kombu may be undesired.

Kombu allows specifying a custom transport by passing a string such as `kombu.transport.pyamqp:Transport` as the `transport` parameter to the `Connection()` constructor. This is a documented, intended feature.

The transport can also be derived from the URL passed as the `hostname` parameter, e.g. `amqp://guest:guest@localhost:5672//`. In this case the URL **scheme** is used as the transport name: it is looked up in the alias-to-class mapping `TRANSPORT_ALIASES`. However, if the scheme is **not** an alias but contains a `.` or `:`, `resolve_transport()` (`kombu/transport/__init__.py`) does not reject it — it treats it as a fully-qualified class path and passes it straight to `symbol_by_name()`, which calls `importlib.import_module()`.

```python
# kombu/transport/__init__.py  (resolve_transport)
try:
    transport = TRANSPORT_ALIASES[transport]
except KeyError:
    if '.' not in transport and ':' not in transport:
        raise KeyError(...)          # plain unknown name -> rejected
    # dotted / colon'd scheme falls through unrejected ...
return symbol_by_name(transport)     # -> importlib.import_module()
```

`Connection.__init__` (`kombu/connection.py`) extracts the scheme via `urllib.parse.urlparse` and resolves it.

## Impact

There is a case — which may be the unintended composition of the two separate behaviours above — that is a problem when an **untrusted** URL is passed to the constructor.

A URL whose scheme is a dotted module path forces `importlib.import_module()` on any module name resolvable on `sys.path`, e.g.:

```
evilmodpoc.transport://localhost/
```

This yields code execution when a module with import-time side effects is reachable on `sys.path`. It is not arbitrary file loading from a path — only already-importable modules can be named — but the import of an attacker-plantable or side-effecting module is sufficient. 

Broker URLs are normally operator-controlled and trusted. This becomes relevant where the URL originates from a less-trusted layer — e.g. a multi-tenant configuration store, a Kubernetes secret populated by a different principal, environment-variable injection, or an SSRF/config-write primitive.

Dotted schemes are valid per RFC 3986 (they are used, for example, by OAuth native-app redirect schemes and mobile deep links), but they are never used as kombu transport names — so constraining them is a kombu-specific decision, not a URL-validity check.

## Possible solutions

- **If the behaviour is desired**, either:
  - keep it as-is and document that the URL passed to `Connection()` must be trusted input, and what an untrusted one can lead to; or
  - add a security mechanism (this may introduce a breaking change, or be opt-in and therefore leave an unsafe default).

- **If the behaviour is not desired** (no support for URLs with dotted schemes), add a check for a `.` in the scheme **derived from the URL**. A dotted class path remains supported via the `transport` keyword argument, so the documented custom-transport feature is unaffected.

## Example patch

The check is applied only to transports derived from the URL (both the plain scheme and the `prefix+scheme://` split); the `transport` keyword argument is left untouched.

```diff
diff --git a/kombu/connection.py b/kombu/connection.py
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -211,9 +211,21 @@ class Connection:
                 # e.g. sqla+mysql://root:masterkey@localhost/
                 params['transport'], params['hostname'] = \
                     hostname.split('+', 1)
+                # a dotted scheme taken from the URL would be imported as an
+                # arbitrary module by resolve_transport(); refuse it.
+                if '.' in params['transport']:
+                    raise ValueError(
+                        f"Invalid transport scheme: {params['transport']!r}")
                 self.uri_prefix = params['transport']
             elif '://' in hostname:
-                transport = transport or urlparse(hostname).scheme
+                if not transport:
+                    transport = urlparse(hostname).scheme
+                    # a dotted URL scheme would be imported as an arbitrary
+                    # module by resolve_transport(); refuse it.  Pass the
+                    # `transport` keyword argument to use a custom class.
+                    if '.' in transport:
+                        raise ValueError(
+                            f'Invalid transport scheme: {transport!r}')
                 if not get_transport_cls(transport).can_parse_url:
                     # we must parse the URL
                     url_params = parse_url(hostname)
```

ndaprela:
This commit introduces two checks meant to prevent untrusted URLs containing dots in their scheme to be abused to load arbitrary modules from sys.path. Fixes #2641

### DIFF
--- a/kombu/connection.py
+++ b/kombu/connection.py
@@ -211,9 +211,16 @@ class Connection:
                 # e.g. sqla+mysql://root:masterkey@localhost/
                 params['transport'], params['hostname'] = \
                     hostname.split('+', 1)
+                if '.' in params['transport']:
+                    raise ValueError(
+                        f"Invalid transport scheme: {params['transport']!r}")
                 self.uri_prefix = params['transport']
             elif '://' in hostname:
-                transport = transport or urlparse(hostname).scheme
+                if not transport:
+                    transport = urlparse(hostname).scheme
+                    if '.' in transport:
+                        raise ValueError(
+                            f'Invalid transport scheme: {transport!r}')
                 if not get_transport_cls(transport).can_parse_url:
                     # we must parse the URL
                     url_params = parse_url(hostname)
--- a/t/unit/test_connection.py
+++ b/t/unit/test_connection.py
@@ -134,6 +134,34 @@ class test_connection_utils:
         assert conn.as_uri(include_password=True) == self.pg_url
 
 
+class test_Connection_transport_scheme:
+    """A transport scheme derived from the URL must not be a dotted import
+    path (which resolve_transport() would import as an arbitrary module)."""
+
+    def test_dotted_url_scheme_rejected(self):
+        with pytest.raises(ValueError, match='Invalid transport scheme'):
+            Connection('some.evil.module://localhost')
+
+    def test_dotted_scheme_via_uri_prefix_rejected(self):
+        with pytest.raises(ValueError, match='Invalid transport scheme'):
+            Connection('some.evil.module+amqp://localhost')
+
+    def test_alias_scheme_not_rejected(self):
+        assert Connection('memory://').transport_cls == 'memory'
+
+    def test_uri_prefix_alias_not_rejected(self):
+        # documented `prefix+scheme://` form with a non-dotted prefix
+        assert Connection('sqla+mysql://localhost').uri_prefix == 'sqla'
+
+    def test_dotted_transport_keyword_not_rejected(self):
+        # the documented way to use a custom transport class is the
+        # `transport` keyword argument, which must keep working.
+        from kombu.transport.memory import Transport as MemoryTransport
+        conn = Connection(
+            'memory://', transport='kombu.transport.memory:Transport')
+        assert isinstance(conn.transport, MemoryTransport)
+
+
 class test_Connection:
 
     def setup_method(self):


auvipy (changes requested):
please also add unit tests to verify this changes

MehrazRumman (changes requested):
Thanks for picking this up. The two checks are correct as far as they go, and the tests for the `transport=` keyword are the right thing to protect. Requesting changes because, at `17c2e32`, two URL-derived paths still reach `symbol_by_name()` and therefore `importlib.import_module()`, so the PR closes #2641 only partially.

**Bypass 1: a `:` in the `prefix+scheme://` split.** The note on the issue that "RFC 3986 schemes cannot contain `:`" is true for `urlparse(...).scheme`, so the `://` branch is fine with a dot check alone. But the `+` branch is not a scheme parse; it is `hostname.split('+', 1)[0]`, which is raw text. `resolve_transport()` lets anything containing `.` **or** `:` through to `symbol_by_name()`, and the new check only looks for `.`:

```python
Connection('baitmod:x+amqp://localhost')   # passes the new check
# -> transport_cls = 'baitmod:x'
# -> get_transport_cls() -> resolve_transport() -> symbol_by_name('baitmod:x')
# -> importlib.import_module('baitmod'); getattr(module, 'x')
```

I ran kombu's own `resolve_transport()` and `symbol_by_name()` code with a bait module on `sys.path`: `'baitmod:x'` imports the module and returns its attribute. The import is lazy here (on first `.transport` access rather than in `__init__`), but it is the same primitive.

**Bypass 2: failover alternates and `switch()`.** Only `alt[0]` goes through the guarded branches in `__init__`. Every other alternate (from `;`-separated URLs, a list `hostname`, or `alternates=`) is handed to `switch()`, which does:

```python
conn_params = parse_url(conn_str) if "://" in conn_str else {"hostname": conn_str}
self._init_params(**dict(self._initial_params, **conn_params))
```

`parse_url()` returns `transport=scheme`, so `self.transport_cls` becomes the raw scheme with no check, and `_do_close_transport()` has already reset `_transport`, so the next `.transport` access resolves it. `Connection('amqp://h1;evil.mod://h2')` is accepted, and the import fires on the first failover.

**Suggested shape.** One helper applied at all three URL-derived sites:

```python
def _check_url_transport(name):
    if '.' in name or ':' in name:
        raise ValueError(f'Invalid transport scheme: {name!r}')
```

- `+` branch: `_check_url_transport(params['transport'])`
- `://` branch: `_check_url_transport(transport)` (unchanged in effect, just shared)
- `switch()`: `if 'transport' in conn_params: _check_url_transport(conn_params['transport'])`

It may also be worth validating every entry of `alt` eagerly in `__init__`, so a bad alternate fails at construction rather than mid-failover. No alias in `TRANSPORT_ALIASES` contains `.` or `:` (I checked all 27), so rejecting both characters for URL-derived names cannot break a legitimate scheme, and the `transport=` keyword keeps accepting dotted and colon paths exactly as your `test_dotted_transport_keyword_not_rejected` pins down.

**Tests to add:** `Connection('x:y+amqp://localhost')` raises; `Connection('amqp://h1;a.b://h2')` (or `.switch('a.b://h2')`) raises; and a positive check that `switch('redis://h2')` still works.

Everything else looks good: `ValueError` is the right type, the two existing tests cover the two guarded branches, and CI is green.

ndaprela:
I added the requested changes.

auvipy:
@kokhlo here is the pr to review

auvipy:
@GangEunzzang would be great if you can do a final check? do we need additional integration tests for this? since connection is a very critical part. @ndaprela what is your take on adding integration test? can you explore the scope?

GangEunzzang (commented):
@auvipy LGTM (including Redis failover and a real Celery worker); I'd add a failover integration test to keep the alternate-URL path covered.

auvipy:
> I'd add a failover integration test to keep the alternate-URL path covered.

@ndaprela can you please revisit?

auvipy (approved):

PR #2645, opened 2026-09-11
Prevent path traversals in kombu.transport.filesystem

### LINKED ISSUE #2644: Filesystem transport: exchange name is joined into a path without sanitisation
## Component
`kombu.transport.filesystem`

## Behaviour

The filesystem transport persists each exchange's binding table as a file inside `control_folder`, building the path directly from the exchange name:

```python
# kombu/transport/filesystem.py
file = self.control_folder / f"{exchange}.exchange"
```

This appears at three sites — `get_table()` (read), `_queue_bind()` (write/create), and `_delete()` (read/write). The `exchange` string is not sanitised, so a name containing `..` or a path separator (or an absolute path) escapes `control_folder`:

```
get_table('../../etc/foo')       -> reads  <control_folder>/../../etc/foo.exchange
_queue_bind('../../tmp/foo', …)  -> writes <control_folder>/../../tmp/foo.exchange
```

The write content is the JSON binding table (`[["<routing_key>","","<queue>"]]`), and the filename always carries a `.exchange` suffix — so it's a constrained write, not free-form file creation.

## Scope / reachability

Exchange names are generally trusted developer configuration, though the exposed behaviour is relevant in two ways:

- It violates the reasonable invariant that the filesystem transport stays confined to its configured folders — a hazard for any consumer that derives an exchange name from lesser-trusted input.
- In Celery it becomes wire-reachable: the `add_consumer` remote-control command forwards an attacker-supplied `exchange` to `add_task_queue()` and on into this transport, so on a worker configured with `broker_url='filesystem://…'` a broker/pidbox writer can drive `_queue_bind` with a traversal name.

## Suggested fix

Reject exchange names that contain path separators or are equal to either one of `.` and `..`:
```diff
diff --git a/kombu/transport/filesystem.py b/kombu/transport/filesystem.py
index e0dfd2a3..13ce3460 100644
--- a/kombu/transport/filesystem.py
+++ b/kombu/transport/filesystem.py
@@ -161,8 +161,14 @@ class Channel(virtual.Channel):
 
     supports_fanout = True
 
+    def _exchange_file(self, exchange):
+        if (exchange in {".", ".."} or os.sep in exchange
+                or (os.altsep and os.altsep in exchange)):
+            raise ChannelError(f"Invalid exchange name: {exchange!r}")
+        return self.control_folder / f"{exchange}.exchange"
+
     def get_table(self, exchange):
-        file = self.control_folder / f"{exchange}.exchange"
+        file = self._exchange_file(exchange)
         try:
             f_obj = file.open("r")
             try:
@@ -178,7 +184,7 @@ class Channel(virtual.Channel):
             raise ChannelError(f"Cannot open {file}")
 
     def _queue_bind(self, exchange, routing_key, pattern, queue):
-        file = self.control_folder / f"{exchange}.exchange"
+        file = self._exchange_file(exchange)
         self.control_folder.mkdir(exist_ok=True)
         queue_val = exchange_queue_t(routing_key or "", pattern or "",
                                      queue or "")
@@ -265,7 +271,7 @@ class Channel(virtual.Channel):
     def _delete(self, queue, exchange, routing_key, pattern, *args, **kwargs):
         super()._delete(queue, exchange, routing_key, pattern, *args, **kwargs)
 
-        file = self.control_folder / f"{exchange}.exchange"
+        file = self._exchange_file(exchange)
         queue_val = exchange_queue_t(routing_key or "", pattern or "",
                                      queue or "")
         f_obj = None

```

ndaprela:
This PR adds a centralized check in kombu/transport/filesystem.py to verify that paths built interpolating the exchange name do not contain path traversals payloads. The PR also adds the relative unit tests. Fixes #2644 .

### DIFF
--- a/kombu/transport/filesystem.py
+++ b/kombu/transport/filesystem.py
@@ -161,8 +161,14 @@ class Channel(virtual.Channel):
 
     supports_fanout = True
 
+    def _exchange_file(self, exchange):
+        if (exchange in {".", ".."} or os.sep in exchange
+                or (os.altsep and os.altsep in exchange)):
+            raise ChannelError(f"Invalid exchange name: {exchange!r}")
+        return self.control_folder / f"{exchange}.exchange"
+
     def get_table(self, exchange):
-        file = self.control_folder / f"{exchange}.exchange"
+        file = self._exchange_file(exchange)
         try:
             f_obj = file.open("r")
             try:
@@ -178,7 +184,7 @@ class Channel(virtual.Channel):
             raise ChannelError(f"Cannot open {file}")
 
     def _queue_bind(self, exchange, routing_key, pattern, queue):
-        file = self.control_folder / f"{exchange}.exchange"
+        file = self._exchange_file(exchange)
         self.control_folder.mkdir(exist_ok=True)
         queue_val = exchange_queue_t(routing_key or "", pattern or "",
                                      queue or "")
@@ -265,7 +271,7 @@ class Channel(virtual.Channel):
     def _delete(self, queue, exchange, routing_key, pattern, *args, **kwargs):
         super()._delete(queue, exchange, routing_key, pattern, *args, **kwargs)
 
-        file = self.control_folder / f"{exchange}.exchange"
+        file = self._exchange_file(exchange)
         queue_val = exchange_queue_t(routing_key or "", pattern or "",
                                      queue or "")
         f_obj = None
--- a/t/unit/transport/test_filesystem.py
+++ b/t/unit/transport/test_filesystem.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import contextlib
+import os
 import shutil
 import tempfile
 from queue import Empty
@@ -11,6 +12,7 @@ import pytest
 
 import t.skip
 from kombu import Connection, Consumer, Exchange, Producer, Queue
+from kombu.exceptions import ChannelError
 from kombu.transport.virtual import Channel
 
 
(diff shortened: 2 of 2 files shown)


auvipy:
@MehrazRumman please cross check this

MehrazRumman:
working on it @auvipy !

MehrazRumman (changes requested):
## Review

I checked out the branch and ran `t/unit/transport/test_filesystem.py` locally: **18 passed**. The fix is correct on POSIX, the three call sites from the issue are all covered, and routing the construction through one `_exchange_file()` helper is the right shape. Centralising it means a future fourth call site is far more likely to inherit the check.

I probed the guard against a few inputs the tests do not cover. One gap is worth fixing before merge.

### Should be fixed: the guard misses Windows drive-relative paths

The transport explicitly supports Windows (there is an `os.name == 'nt'` branch using `win32file` for locking), so the guard has to hold there too. It does not. A drive-relative name such as `D:evil` contains neither `os.sep` nor `os.altsep`, so it passes, and `pathlib` then discards the control folder entirely:

```
Simulating the guard on Windows (os.sep='\\', os.altsep='/'):

  'D:evil'         guard_allows=True  -> 'D:evil.exchange'   <== escapes control_folder
  'C:evil'         guard_allows=True  -> 'C:\app\control\evil.exchange'
  'CON'            guard_allows=True  -> 'C:\app\control\CON.exchange'   <== reserved device name
```

`C:evil` happens to stay inside only because the anchor drive matches; if `control_folder` sits on `D:`, the two swap roles. Reserved device names (`CON`, `NUL`, `COM1`) also survive the suffix on Windows and address the device rather than a file.

**This is invisible to CI.** Every test class in the module carries `@t.skip.if_win32`, including the new one, so the `os.altsep` half of the guard is never executed by the suite. That is consistent with existing convention, so I am not asking you to change the skip, but it does mean the Windows behaviour has to be reasoned about rather than tested.

**Suggested fix:** verify containment of the constructed path instead of blocklisting the characters that could break it. Something along the lines of:

```python
def _exchange_file(self, exchange):
    file = self.control_folder / f"{exchange}.exchange"
    if file.parent != self.control_folder:
        raise ChannelError(f"Invalid exchange name: {exchange!r}")
    return file
```

I ran this against the same inputs: it rejects `D:evil`, `../escape` and `sub/child`, and accepts `good.name`. It needs the explicit `exchange in {".", ".."}` case kept alongside it, since `".."` yields `"...exchange"`, whose parent is still the control folder. Allowlisting the accepted characters would work equally well and would also exclude the reserved device names, at the cost of rejecting some exotic but legitimate names.

### Notes, not blocking

1. **Queue names reach a path too, and are not covered.** `_put()` interpolates the queue name into a filename under `data_folder_out`. I tried to exploit it and could not: the name is appended after a `<timestamp>_<uuid4>.` prefix, so any traversal needs an intermediate directory named with a random UUID, which an attacker cannot create. So there is no bug to fix, but it is the obvious next question a reviewer will ask, and the reasoning is worth a sentence in the PR description so nobody has to re-derive it.

2. **A latent bug surfaced while probing that one.** When `open()` fails in `_put()`, the `finally` block runs `unlock(f)` before `f` was ever bound, so the intended `ChannelError` is masked by `UnboundLocalError`. Pre-existing and out of scope here, but worth its own issue.

3. **Behaviour change for legitimate names.** AMQP permits `/` in exchange names, so anyone using one on the filesystem transport moves from silently writing outside the control folder to a hard `ChannelError`. That is the correct trade, and it deserves a Changelog line so the failure is recognisable on upgrade.

4. **Test quality is good.** `test_relative_name_cannot_alias_another_exchange` is the one I would have asked for, since table poisoning through an alias is the more interesting consequence of this bug than the file write itself. Nice to see it already there.

Fix the Windows case and I think this is ready.

auvipy:
@GangEunzzang would you mind looking into it?

GangEunzzang (commented):
Ran a Celery worker on the filesystem transport against this branch: tasks, pidbox and `inspect ping` all work, and the `../escaped` bind that writes outside `control_folder` on `main` is rejected here. LGTM.

auvipy (approved):

kombu/transport/filesystem.py
```diff
@@ -155,14 +156,63 @@ def unlock(file):
 exchange_queue_t = namedtuple("exchange_queue_t",
                               ["routing_key", "pattern", "queue"])
 
+#: Characters accepted in an exchange name.  The name is interpolated into a
+#: filename under ``control_folder``, so rather than blocklisting the
+#: constructs that can redirect that path, only characters that cannot are
+#: allowed.  This rejects path separators, ``..``, Windows drive-relative
+#: prefixes (``D:evil``) and UNC prefixes on every platform, rather than only
+#: on the one the tests happen to run on.
+EXCHANGE_NAME_RE = re.compile(r'\A[A-Za-z0-9._-]+\Z')
```
jayaddison, line 165: Maybe nitpicky: this comment seems slightly misleading, in my opinion; particulary the line:

> #: allowed.  This rejects path separators, ``..``, Windows drive-relative

This regex _does_ accept `..`.  There _is_ code in `_is_valid_exchange_name` that rejects when `exchange.strip(".")` is falsy, but that is separate to this regex.

Edit: boolean inversion fixup

auvipy: you can come with a new pr

PR #2647, opened 2026-09-11
fix(filesystem): match queue name exactly, not as a substring

### LINKED ISSUE #2646: filesystem transport: queue name matched as a substring, so dotted queue names cross-deliver
If two queues share a spool directory and one queue's name is a dotted suffix of the other's
(`b` and `a.b`, `tasks` and `priority.tasks`), the shorter queue silently consumes, counts and
purges the longer queue's messages.

## Cause

`_put` writes spool files as `<ts>_<uuid4>.<queue>.msg` (`filesystem.py:210`):

```python
filename = '{}_{}.{}.msg'.format(int(round(monotonic() * 1000)),
                                 uuid.uuid4(), queue)
```

The readers match that filename with a substring test instead of comparing the queue field
(`filesystem.py:227,234`, and the same pair at `299,306` in `_purge` and `325,331` in `_size`):

```python
queue_find = '.' + queue + '.msg'
...
if filename.find(queue_find) < 0:
    continue
```

A file written for queue `a.b` is `…_<uuid>.a.b.msg`, which contains `.b.msg`, so a consumer of
`b` accepts it.

## Reproduce

```python
import tempfile, kombu

spool = tempfile.mkdtemp()
conn = kombu.Connection('filesystem://', transport_options={
    'data_folder_in': spool, 'data_folder_out': spool, 'control_folder': spool})

victim = kombu.Queue('a.b', kombu.Exchange('a.b', 'direct'), routing_key='a.b')
thief = kombu.Queue('b', kombu.Exchange('b', 'direct'), routing_key='b')
got = []

with conn.channel() as chan:
    kombu.Producer(chan).publish({'secret': 'for a.b'}, exchange=victim.exchange,
                                 routing_key=victim.routing_key, declare=[victim, thief])
    print('_size("a.b") =', chan._size('a.b'), '  _size("b") =', chan._size('b'))
    with kombu.Consumer(chan, [thief], accept=['json'],
                        callbacks=[lambda b, m: (got.append(b), m.ack())]):
        try:
            conn.drain_events(timeout=1)
        except TimeoutError:
            pass
print('consumer of "b" received:', got)
```

On kombu 5.6.2, CPython 3.14.4, Linux:

```
_size("a.b") = 1   _size("b") = 1
consumer of "b" received: [{'secret': 'for a.b'}]
```

`_purge` behaves the same way: with three messages published to `a.b`, `chan._purge('b')` returns
`3` and leaves `chan._size('a.b') == 0`.

## Impact

Messages are delivered to the wrong worker and processed there. The queue they were routed to
never sees them, and `_size` reports a backlog it does not own, so the misdelivery is not visible
from queue depth. `_purge` on one queue deletes another queue's backlog. 

At the moment, no security boundary is crossed: a queue name is a
substring of a filename in a shared directory and anything that can consume one
queue already holds the directory access needed to read every other queue's files outright. 

There is one shape in which the same defect becomes a confidentiality and integrity issue. It
needs all three of:

1. a component that holds the only filesystem access to the spool and consumes on behalf of
   callers, so the caller never touches `data_folder_in` itself;
2. the consumed queue name derived from untrusted input; and
3. a victim queue on the same spool whose spool filenames contain the attacker-chosen
   `.<name>.msg` needle — `b` against a victim `a.b`, and so on.

Given those three, a caller who supplies a matching name reads messages routed to the victim queue
without ever holding access to the spool, or supplies that name to a purge and deletes the victim's
backlog.

## Suggested fix

The prefix `_put` writes (an `int` timestamp and a `uuid4`) contains no `.`, so the queue name is
exactly `filename.partition('.')[2]` minus the extension. Comparing that for equality fixes all
three sites:

```diff
     def _get(self, queue):
         """Get next message from `queue`."""
-        queue_find = '.' + queue + '.msg'
+        queue_find = f'{queue}.msg'
         folder = os.listdir(self.data_folder_in)
         folder = sorted(folder)
         while len(folder) > 0:
             filename = folder.pop(0)
 
             # only handle message for the requested queue
-            if filename.find(queue_find) < 0:
+            if filename.partition('.')[2] != queue_find:
                 continue
```

The same two-line change applies in `_purge` and `_size`.

Note that a suffix test (`filename.endswith(queue_find)`) is not enough: `'…uuid.a.b.msg'` ends
with `'.b.msg'`, so it leaves exactly the case above unfixed.

## Credits

Found by [OSTIF](https://ostif.org/) using AI with their own harness.  
Manually verified, fixed, and reported by [Shielder](https://www.shielder.com/).

mindlaess:
Fixes #2646

## Problem

`_get`, `_purge` and `_size` in the filesystem transport select spool files with a
substring test rather than by comparing the queue field:

```python
queue_find = '.' + queue + '.msg'
...
if filename.find(queue_find) < 0:
    continue
```

`_put` writes `<ts>_<uuid4>.<queue>.msg`, so a file belonging to queue `a.b` is
`…_<uuid>.a.b.msg` — which contains `.b.msg`. A consumer of `b` therefore accepts it.
Where two queues share a spool and one name is a dotted suffix of the other, the shorter
queue silently consumes, counts and purges the longer queue's messages. `_purge` on one
queue deletes the other's backlog.

## Fix

The prefix `_put` writes (an `int` timestamp and a `uuid4`) contains no `.`, so the queue
field is exactly `filename.partition('.')[2]` minus the extension. Comparing that for
equality anchors the match at the start of the queue field, at all three sites:

```python
queue_find = f'{queue}.msg'
...
if filename.partition('.')[2] != queue_find:
    continue
```

A suffix test (`filename.endswith('.' + queue + '.msg')`) is **not** sufficient —
`…uuid.a.b.msg` ends with `.b.msg`, leaving the reported case unfixed.

No filename format change, so existing spools stay readable.

## Verification

Behaviour corpus of 10 `(queue consumed, queue written to)` pairs driven through
`Channel._put` / `Channel._size` on a real spool:

| Build | Wrong results / 10 |
|---|---|
| unpatched | 5 |
| `endswith` | 3 |
| this patch | 0 |

`('b','b')`, `('a.b','a.b')` and `('b.msg','b.msg')` still match, so the change does not
over-restrict.

The added regression test fails on unpatched `main` (`assert 1 == 0`) and passes with the fix.

Unit suite `python -bb -m pytest t/unit -q`: no regressions, +1 passing test.

### DIFF
--- a/kombu/transport/filesystem.py
+++ b/kombu/transport/filesystem.py
@@ -224,14 +224,14 @@ class Channel(virtual.Channel):
 
     def _get(self, queue):
         """Get next message from `queue`."""
-        queue_find = '.' + queue + '.msg'
+        queue_find = f'{queue}.msg'
         folder = os.listdir(self.data_folder_in)
         folder = sorted(folder)
         while len(folder) > 0:
             filename = folder.pop(0)
 
             # only handle message for the requested queue
-            if filename.find(queue_find) < 0:
+            if filename.partition('.')[2] != queue_find:
                 continue
 
             if self.store_processed:
@@ -296,14 +296,14 @@ class Channel(virtual.Channel):
     def _purge(self, queue):
         """Remove all messages from `queue`."""
         count = 0
-        queue_find = '.' + queue + '.msg'
+        queue_find = f'{queue}.msg'
 
         folder = os.listdir(self.data_folder_in)
         while len(folder) > 0:
             filename = folder.pop()
             try:
                 # only purge messages for the requested queue
-                if filename.find(queue_find) < 0:
+                if filename.partition('.')[2] != queue_find:
                     continue
 
                 filename = os.path.join(self.data_folder_in, filename)
@@ -322,13 +322,13 @@ class Channel(virtual.Channel):
         """Return the number of messages in `queue` as an :class:`int`."""
         count = 0
 
-        queue_find = f'.{queue}.msg'
+        queue_find = f'{queue}.msg'
         folder = os.listdir(self.data_folder_in)
         while len(folder) > 0:
             filename = folder.pop()
 
             # only handle message for the requested queue
-            if filename.find(queue_find) < 0:
+            if filename.partition('.')[2] != queue_find:
                 continue
 
             count += 1
--- a/t/unit/transport/test_filesystem.py
+++ b/t/unit/transport/test_filesystem.py
@@ -195,6 +195,30 @@ class test_FilesystemTransport(WithJanitorMixin):
             self.q2(consumer_channel).purge()
             assert self.q2(consumer_channel).get() is None
 
+    def test_dotted_queue_name_not_matched_by_suffix(self):
+        producer_channel = self._add_channel(self.p.channel())
+        consumer_channel = self._add_channel(self.c.channel())
+        producer = Producer(producer_channel, self.e)
+        # 'b' is a dotted suffix of 'a.b', the two must not share messages
+        dotted = Queue('a.b', exchange=self.e, routing_key='a.b')
+        suffix = Queue('b', exchange=self.e, routing_key='b')
+        with (
+            managed_consumer(consumer_channel, dotted),
+            managed_consumer(consumer_channel, suffix),
+        ):
+            dotted(consumer_channel).declare()
+            suffix(consumer_channel).declare()
+
+            producer.publish({'foo': 1}, routing_key='a.b')
+
+            assert suffix(consumer_channel).queue_declare(
+                passive=True).message_count == 0
+            assert suffix(consumer_channel).get() is None
+
+            # purging 'b' must leave the message routed to 'a.b' in place
+            suffix(consumer_channel).purge()
+            assert dotted(consumer_channel).get()
+
 
 @t.skip.if_win32
 class test_FilesystemFanout(WithJanitorMixin):


auvipy (changes requested):
can we please improve the test coverage?

MehrazRumman (commented):
The fix is correct and complete. Notes from reviewing at `50e7d89`:

**Premise holds.** `_put` writes `'{}_{}.{}.msg'.format(int(round(monotonic() * 1000)), uuid.uuid4(), queue)`, so the part before the first `.` is an integer, `_`, and a UUID with hyphens. It can never contain a dot, which makes `filename.partition('.')[2]` exactly `<queue>.msg`. `_put_fanout` goes through `_put`, so fanout spools use the same format and stay readable. All three scan sites (`_get`, `_purge`, `_size`) are patched consistently; there is no fourth one.

I ran the three predicates over every `(consumer, writer)` pair from `['b', 'a.b', 'b.msg', 'ab', 'b.c', 'xb', 'a.b.c']`:

| Predicate | Wrong pairs |
| --- | --- |
| `find('.' + q + '.msg') >= 0` (current `main`) | `('b','a.b')`, `('b','b.msg')`, `('b.c','a.b.c')` |
| `endswith('.' + q + '.msg')` | `('b','a.b')`, `('b.c','a.b.c')` |
| `partition('.')[2] == f'{q}.msg'` (this PR) | none |

So the description's point that a suffix test would not be enough is right, and the exact comparison does not over-restrict.

**On @auvipy's coverage request, the gap is one specific branch.** Codecov's single partial line is the `if` in `_size()`: the new test asserts `message_count == 0` for `b`, and no test in the file ever asserts a non-zero count, so the `count += 1` side of that branch is never executed. Adding one assertion to the new test closes it and also proves the positive case:

```python
            producer.publish({'foo': 1}, routing_key='a.b')

            assert dotted(consumer_channel).queue_declare(
                passive=True).message_count == 1
            assert suffix(consumer_channel).queue_declare(
                passive=True).message_count == 0
```

Optional, cheap, and worth having: publish a second message to `b` in the same test and check `a.b` still reports one and `b` reports one after the purge/get sequence, so both directions of the pair are pinned rather than only "shorter must not see longer".

Happy to approve once that assertion is in.

auvipy:
@GangEunzzang and this, please?

GangEunzzang (commented):
@auvipy Checked with real prefork workers on `b` and `a.b`, plus count/purge isolation; looks good to me.

auvipy (approved):

PR #2649, opened 2026-09-12
Install the pgmq extra only in the envs that exercise it (fixes PyPy integration CI)

MehrazRumman:
## Problem

Since #2559 (PGMQ transport) the three PyPy integration jobs (`pypy3.11` × `py-amqp`, `py-redis`, `py-mongodb`) fail at the pip install step.

#2559 added `-r extras/pgmq.txt` to `requirements/test-ci.txt`, which every tox env installs. `pgmq` hard-depends on `orjson`, and orjson refuses to build on PyPy in its own build script, so a Rust toolchain (#2634) cannot help:

### DIFF
--- a/t/integration/test_pgmq.py
+++ b/t/integration/test_pgmq.py
@@ -5,13 +5,14 @@ import socket
 from contextlib import closing
 from time import sleep, time
 
-import psycopg
 import pytest
 
 import kombu
 from kombu.exceptions import OperationalError
 
 from .common import BaseExchangeTypes, BaseMessage, BasicFunctionality
+
+psycopg = pytest.importorskip('psycopg')
 
 # connect() does not wrap the transport error as kombu OperationalError.
 _CONN_FAIL = (OperationalError, psycopg.OperationalError, psycopg.InterfaceError)
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -117,8 +117,6 @@ jobs:
                   allow-prereleases: true
                   cache: 'pip'
                   cache-dependency-path: '**/setup.py'
-          -   name: Install Rust toolchain
-              uses: dtolnay/rust-toolchain@stable
           -   name: Install tox
               run: python -m pip install --upgrade pip wheel tox tox-docker
           -   name: >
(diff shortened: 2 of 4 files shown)


t/integration/test_pgmq.py
```diff
@@ -5,14 +5,15 @@
...
 
-import psycopg
 import pytest
 
 import kombu
 from kombu.exceptions import OperationalError
 
 from .common import BaseExchangeTypes, BaseMessage, BasicFunctionality
 
+psycopg = pytest.importorskip('psycopg')
```
auvipy, line 15: why this import or skip?

auvipy:
@tavallaie actually the pgmq most probably introduced the pypy failures

t/integration/test_pgmq.py
MehrazRumman: @auvipy  The `-E` option is not a pytest plugin, it is implemented in our root `conftest.py`. It works in `pytest_runtest_setup`: it looks at the `@pytest.mark.env(...)` marker on each collected test and skips the ones whose env was not passed with `-E`. That happens *after* collection, so pytest still imports every module under `t/integration/` no matter which `-E` value is given.

Before this PR that was harmless, because `test-ci.txt` installed pgmq (and therefore psycopg) in every env. With this PR, pgmq/psycopg are only installed in the pgmq envs, so the plain module-level `import psycopg` in `test_pgmq.py` now breaks collection for every other integration env, PyPy and CPython alike:

```
t/integration/test_pgmq.py:8: in <module>
import psycopg
E   ModuleNotFoundError: No module named 'psycopg'
ERROR t/integration/test_pgmq.py
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
```

(reproduced locally with the unchanged file in a `pypy3.11-linux-integration-py-amqp` env)

`psycopg = pytest.importorskip('psycopg')` turns that into "1 skipped" in the envs that don't have psycopg, and is a no-op in the pgmq envs where it is installed, so the pgmq integration tests still run there. It is the same pattern `t/unit/transport/test_pgmq.py` already uses with `pytest.importorskip("pgmq")`.

auvipy (approved):

PR #2299, opened 2025-05-16
Add support for NATS JetStream as a transport

joeriddles:
closes #2103

Implement support for [NATS](https://nats.io/) using its built-in persistence layer [JetStream](https://docs.nats.io/nats-concepts/jetstream).

### Example

You can test using either:
- Installing the [NATS server](https://docs.nats.io/running-a-nats-service/introduction/installation#getting-the-binary-from-the-command-line) locally and running it with JetStream enabled: `nats-server --js`
- Using the demo NATS server at `demo.nats.io`

The transport can be test using `examples/nats_receive.py` and `examples/nats_send.py`:

1. Start the local NATS server (optional)
2. Run `python -m examples.nats_receive` (append `--demo` if using the demo server)
4. In another window, run `python -m examples.nats_send` (append `--demo` if using the demo server)

In the receive window, you should see something like:
```
Received message: 'hello world'
  properties:
{   'body_encoding': 'base64',
    'delivery_info': {'exchange': 'exchange', 'routing_key': 'messages'},
    'delivery_mode': 2,
    'delivery_tag': 'fa01f1a9-f666-46a0-af71-35e5164e1ad9',
    'priority': 0}
  delivery_info:
{'exchange': 'exchange', 'routing_key': 'messages'}
```

PR #2653, opened 2026-09-15
Add NATS JetStream transport to Kombu

### LINKED ISSUE #2103: NATS support
Hi, thanks for the amazing projects kombu & celery! 

I want to use kombu with [NATS](https://docs.nats.io/), are there any plans to support it?

emattiza:
# PR Draft — NATS JetStream transport with hardening

## Title

NATS JetStream transport (hardened): shared event loop, subscription caching, error improvements

## Description

This PR adds a NATS JetStream transport to Kombu, based on the original
work by @joeriddles (PR #2299) and @maxyloon (PR #2541), with additional
hardening from manual testing:

- Pull-subscription caching (removes an API round-trip per message)
- Shared event loop + NATS client across channels (one loop thread + 3 TCP
  connections instead of N threads and 3N connections)
- Integration test isolation (unique queue names per test, teardown cleanup)
- Clear closed-channel errors (NotBoundError instead of AttributeError)
- New NATSError(OperationalError) with cause chaining for stream/consumer
  creation failures

Closes #2103


MehrazRumman (commented):
Thanks for pushing this forward, a NATS transport has been requested for a long time. I pulled the branch, ran the unit suite locally (177 passed with nats-py 2.15.0) and read through `kombu/transport/nats.py`. A few things below block merging, and a handful of correctness issues I was able to reproduce with the mocked channel. Line numbers refer to ae31c8f.

## Blockers

**1. Unresolved merge conflict in `README.rst` (lines 74-78).**
`<<<<<<< HEAD` / `=======` / `>>>>>>> celery/main` markers are committed. Both the NATS and PGMQ link targets should be kept.

**2. `README.rst` transport table is misaligned (line 109).**
The new `*NATS*` row is 110 chars while every other row/border is 109 (extra space before the last `|`). RST grid tables are strict about this, so the docs build will fail.

**3. `Transport.connection_errors` never matches what nats-py actually raises (lines 100-104).**
The tuple is built from `nats.aio.errors.ErrConnectionClosed` / `ErrTimeout` / `ErrNoServers`. In nats-py 2.x those legacy classes are *subclasses* of `nats.errors.ConnectionClosedError` / `TimeoutError` / `NoServersError`, not aliases, and the client raises the modern ones:

```python
>>> isinstance(nats.errors.NoServersError(), Transport.connection_errors)
False
```

So `Connection.ensure()`, `ensure_connection()` and Celery's broker retry loop treat every NATS connection failure as unrecoverable. Should be `nats.errors.ConnectionClosedError, nats.errors.TimeoutError, nats.errors.NoServersError` (and probably `nats.errors.ConnectionReconnectingError`).

**4. Fanout messages cannot be turned into `Message` objects (lines 342, 781-800).**
`Message.__init__` does `payload["subject"]`, `payload["ack"]`, etc., but the fanout branch of `JetStreamChannel._get` returns the plain dict from `nats_body_and_headers_to_message()` without those keys. So the first broadcast delivered through the normal consume path raises `KeyError: 'subject'`:

```python
payload = ch._get('w.celery.pidbox')   # fanout inbox pre-filled
Message(payload, channel=ch)           # KeyError: 'subject'
```

The unit tests only assert on the dict returned by `_get`, so they don't catch it. Even with the keys present, `QoS.ack` would call `channel.ack_msg(msg)` → `msg.nats_ack()`, which a Core NATS `Msg` can't do. Celery's pidbox (`celery inspect`, `control`, revoke broadcasts) goes through exactly this path, so remote control is currently broken. Fanout messages probably need a no-op ack/nak/term and `subject` set from `msg.subject`.

## Correctness issues (reproduced)

**5. `_ensure_stream` timeout fallback never fires (lines 648-659).**
The check wraps `stream_info` in `asyncio.wait_for(..., timeout=1.0)`, but the `except` only catches `nats.errors.TimeoutError`. `asyncio.wait_for` raises the builtin `TimeoutError`, which is a *parent* of the nats one, so it escapes and `_put`/`_new_queue` blow up instead of falling through to `add_stream`. The `add_stream` / `add_consumer` handlers a few lines further down (`except nats.errors.TimeoutError:`) have the same problem. Catching the builtin `TimeoutError` (which covers the nats one too) fixes all of them.

**6. `Transport._get_client` caches a half-initialised client after a failed connect (lines 1109-1136).**
`self._nats_client = Client()` is assigned before `connect()` is awaited. If the first connect raises (server down, bad credentials), `_nats_client` keeps pointing at a never-connected `Client`, and every later channel gets it back without any reconnect attempt:

```
first call raised: NoServersError
second call: connect awaited 1 time total, returns stale client
```

Reset to `None` on failure (or assign only after connect succeeds). Also, this method isn't guarded by `_loop_lock`, so two threads creating channels concurrently (thread pool + producer pool in a Celery worker) can each dial their own client and one leaks. The `if self._nats_client is None: raise RuntimeError` right after `Client()` is dead code.

**7. `_on_reconnect` breaks `nats+core` channels (lines 1138-1146).**
It clears `_subscriptions` on *every* channel. For `CoreNATSChannel` that dict is the source of truth for live subscriptions. nats-py re-establishes Core NATS subscriptions itself after reconnect, so the old subscription keeps delivering while `_get`/`_new_queue` see an empty dict and subscribe again → two subscriptions feeding the same inbox → every message delivered twice after a reconnect. `_has_queue` also returns `False` until re-subscribed. This should be limited to `JetStreamChannel` (and ideally unsubscribe the stale pull subs instead of dropping the references).

**8. `verify_connection` (lines 1218-1236) ignores credentials and only catches `ValueError`.**
`Connection.connected` calls this on every `default_channel` access / `ensure_connection`. Each call spins up a fresh event loop and a fresh TCP + NATS handshake, and it connects with no `user`/`password`, so on an auth-enabled server the connect fails with a nats-py error (not a `ValueError`) that propagates straight out of `connection.connected`. With the shared client already in place, this should probably just return `self._nats_client is not None and self._nats_client.is_connected`.

**9. `Nats-TTL` needs `allow_msg_ttl=True` on the stream (line 769 vs. 660-676).**
Per-message TTL is only honoured (and, on nats-server ≥ 2.11, the publish is rejected otherwise) when the stream has `allow_msg_ttl` enabled. The default `StreamConfig` doesn't set it, so any Celery task published with `expires=` fails unless the user knows to pass `stream_config={'allow_msg_ttl': True}`. Either set it in the default config or document it. (Also worth noting the default is `StorageType.MEMORY`, so messages don't survive a server restart; for a broker default I'd expect FILE, or at least a loud note in the docs.)

## Test/CI gaps

- The 2000-line unit module is skipped in CI: `t/unit/transport/test_nats.py` does `pytest.importorskip('nats')`, but `tox.ini` only installs `requirements/extras/nats.txt` for the `*-linux` factors, not `*-unit`. Add it to the `3.10-unit,...` line like pgmq.
- `integration-nats` is added to tox but not to the matrix in `.github/workflows/python-package.yml`, so it never runs.
- `test_get_fanout_raises_empty_on_timeout` sets `JetStreamChannel.default_wait_time_seconds = 0.01` at class level and never restores it, which leaks into later tests.

## Minor

- `QoS.restore_unacked_once` is a no-op, so unacked messages on shutdown are only redelivered after the consumer's `ack_wait` (30s default). A `nak` here would make worker restarts snappier.
- `celery.fanout.*` / `celery.queue.*` subject prefixes hard-code "celery" into a kombu transport; something like `kombu.` or a configurable prefix would be more appropriate.
- `_queue_bind` keys fanout subscriptions by exchange but inboxes by queue, so a second queue bound to the same fanout exchange on one channel silently gets no inbox and falls back to the JetStream path.
- `_new_queue` creates a JetStream stream even for fanout-only queues (pidbox reply queues), which is wasteful.
- `examples/nats_benchmark.py` at 400+ lines feels heavy for `examples/`; maybe a separate gist/repo.

Happy to re-review once the README conflict and items 3-4 are sorted, since those are the ones that would break a Celery worker on first use.

MehrazRumman (approved):

PR #2655, opened 2026-09-17
fix(azureservicebus): keep renewal receiver out of messages

### LINKED ISSUE #2654: Azure Service Bus lock renewal makes messages unpickleable in Celery prefork workers
## Summary

Enabling Azure Service Bus lock renewal makes any received messages un-pickleable.

This means that celery prefork workers reserve tasks but can't send them to the child processes, so every concurrency slot becomes occupied and any later tasks remain queued.

The affected Azure Service Bus transport code has not changed since `f304ece19b567fede07189422974c2aa6f146214` (#2545).

## Environment

- Kombu: `main` at `83b8ee0d00c5ff99573bf0c369b689386d09cbd5`
- Celery: `5.6.3`
- azure-servicebus: reproduced with `7.14.2`; `7.14.3` has the same `__getstate__` behavior
- Python: `3.12`
- Worker pool: prefork
- Operating system: Linux

## Steps to reproduce

Configure a Celery application with Azure Service Bus and lock renewal enabled:

```python
broker_transport_options = {
    "max_lock_renewal_duration": 1920,
    "use_lock_renewal": True,
}
```

Start a prefork worker and enqueue tasks. The worker logs each task as received, but the task does not start in a child process.

The serialization failure can also be demonstrated directly:

```python
import pickle
import threading

from azure.servicebus import ServiceBusReceivedMessage

message = object.__new__(ServiceBusReceivedMessage)
message._receiver = object()
message._uamqp_message = object()
message._kombu_renewal_receiver = threading.Lock()

pickle.dumps(message)
```

This raises:

```text
TypeError: cannot pickle '_thread.lock' object
```

## Cause

`Channel._register_for_renewal()` stores the dedicated renewal receiver on the received message:

```python
message._kombu_renewal_receiver = renewal_obj.receiver
```

`ServiceBusReceivedMessage.__getstate__()` copies the complete instance dictionary and clears only `_receiver` and `_uamqp_message`. It does not clear the dynamically added `_kombu_renewal_receiver` attribute. The dedicated receiver contains a thread lock, so the message can no longer be pickled.

Celery copies the message properties into the task request that it serializes for the prefork pool. The Azure message is present under `properties.delivery_info.azure_message`, which exposes this failure during task dispatch.

Relevant code:

- https://github.com/celery/kombu/blob/main/kombu/transport/azureservicebus.py#L434-L439
- https://github.com/Azure/azure-sdk-for-python/blob/azure-servicebus_7.14.3/sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py#L804-L808
- https://github.com/celery/celery/blob/v5.6.3/celery/worker/request.py#L157-L171

## Expected behavior

Enabling lock renewal must not prevent received messages from being serialized and dispatched to prefork worker processes.

## Proposed fix

Keep the dedicated renewal receiver, but do not attach it to the received message. Pass it to a per-registration lock-renewal failure callback instead, for example with `functools.partial`. The retry callback can then reuse the dedicated receiver without adding an unpickleable object to the message.

A regression test can register a received message for renewal and verify that the message remains pickleable.

jerbob:
Fixes #2654.

## Problem

When Azure Service Bus lock renewal is enabled, Kombu attaches the dedicated renewal receiver to each `ServiceBusReceivedMessage`. That receiver contains thread locks, so Celery prefork workers cannot pickle the task request and dispatch it to a child process.

## Root cause

`ServiceBusReceivedMessage.__getstate__()` clears the SDK's `_receiver` and `_uamqp_message` attributes, but it does not know about Kombu's dynamically added `_kombu_renewal_receiver`. The added receiver therefore remains in the serialized state and raises `TypeError: cannot pickle '_thread.lock' object`.

## Fix

Keep the dedicated renewal receiver out of the message. Pass it to a per-registration `on_lock_renew_failure` callback with `functools.partial`, so retry registration still uses the same receiver without making the message unpickleable.

## Testing

The added regression test uses a real `ServiceBusReceivedMessage` and a thread lock. It fails on unmodified `main` and passes with this change.

```text
python -bb -m pytest -q t/unit/transport/test_azureservicebus.py
55 passed

flake8 kombu/transport/azureservicebus.py t/unit/transport/test_azureservicebus.py
isort --check-only kombu/transport/azureservicebus.py t/unit/transport/test_azureservicebus.py
```

Tested with Python 3.13.12 and `azure-servicebus==7.14.3`.


auvipy:
@MehrazRumman

MehrazRumman (changes requested):
Nice fix — the diagnosis in #2654 is right and the approach is the right shape.

Attaching `_kombu_renewal_receiver` to the `ServiceBusReceivedMessage` was always going to break prefork, since `ServiceBusReceivedMessage.__getstate__()` only knows to strip the SDK's own `_receiver` / `_uamqp_message`. Carrying the receiver in a `functools.partial` on the per-registration `on_lock_renew_failure` callback is better than the alternatives (overriding `__getstate__`, or a side table keyed on the message, which would be leak-prone since these messages aren't reliably hashable or weak-referenceable). It also stops kombu from mutating an SDK object, which is a win on its own. Diff is small, the unrelated churn was cleaned up in the follow-up commits, and coverage holds.

A few things I'd like addressed before merge:

**1. Add a comment explaining *why* the partial exists.** This is my main ask. `_register_message_for_renewal` currently reads as an ordinary indirection, and the obvious "simplification" — moving `on_lock_renew_failure` back onto the `AutoLockRenewer` constructor — silently reintroduces the bug, because the renewer-level callback only receives `(renewable, error)` and you'd need to get the receiver off the message again. Something like:

```python
def _register_message_for_renewal(self, receiver, message):
    # The receiver is passed through the callback rather than attached to
    # the message: it holds thread locks, and anything set on the message
    # survives __getstate__ and makes it unpickleable in prefork workers
    # (see #2654).
```

Same for the regression test — it has no docstring while its neighbours do, and `object.__new__(ServiceBusReceivedMessage)` deserves a line saying it's there to exercise the real `__getstate__`.

**2. `_create_auto_lock_renewer` no longer sets a renewer-level failure callback.** Today `_register_message_for_renewal` is the only registration path so nothing is lost, but the renewer is now fail-open: any future `self.connection._renewer.register(...)` that forgets the kwarg quietly drops the retry logic. Worth a comment on `_create_auto_lock_renewer` noting the callback is per-registration by design.

**3. The tests assert on `partial` internals rather than behaviour.** `callback.func == channel._on_lock_renew_failure` / `callback.args == (renewal_receiver,)` pins the implementation, so switching to a closure or different currying would fail the test without anything actually being wrong. Invoking the callback and checking the re-registration would be more robust:

```python
callback = renewer.register.call_args.kwargs["on_lock_renew_failure"]
callback(msg_a, Exception("transient"))
assert renewer.register.call_args.args == (renewal_receiver, msg_a)
```

Nits:

- `_register_message_for_renewal` and `_on_lock_renew_failure` have no annotations while most of the module does; `receiver: ServiceBusReceiver, message: ServiceBusReceivedMessage) -> None` would match the surrounding style.
- `_register_message_for_renewal` assumes `self.connection._renewer is not None`. The caller guarantees it and the retry path is inside a `try/except`, but there's a pre-existing shutdown race — `Transport.close_connection` sets `_renewer = None` while renewal threads may still be in flight, so a late failure callback produces an `AttributeError` logged via `logger.exception`. Not a regression here, just noise you may want to guard against separately.
- One `partial` is allocated per message rather than per receiver. Negligible, but building it once when the renewal receiver is created would be marginally tidier.
- Three commits; probably worth squashing on merge.

One thing worth double-checking: the per-registration `on_lock_renew_failure` kwarg on `AutoLockRenewer.register()`. It's present across `azure-servicebus` 7.x wherever `register()` takes a receiver first — which the existing code already relies on — so the floor shouldn't move, but confirming against the minimum pinned version in `requirements/` would close that out.

auvipy:
@jerbob can you please address the review comments

auvipy:
@GangEunzzang need your help to verify this as well

auvipy (changes requested):
should we also consider adding integration test to this pr?

GangEunzzang (commented):
@auvipy The fix checks out locally with real SDK messages through a prefork pool; I haven't run it against Azure itself. I'd add an integration test against Azure with `use_lock_renewal=True` and `acks_late=True`, keeping the task running past the initial lock. @jerbob please check

kokhlo:
Independent cross-check of the fix — reproduced the bug on `main` and verified this branch end-to-end with the real SDK, no Azure connection needed.

**Repro on `main` (b45d959, `azure-servicebus==7.14.3` — the exact pin in `requirements/extras/azureservicebus.txt`):** built a real `ServiceBusReceivedMessage` via `object.__new__` with the attributes `AutoLockRenewer.register()` touches (`_received_timestamp_utc`, `_settled`, `_raw_amqp_message.annotations`, future `_expiry`), a `threading.Lock()` as both `_receiver` and the renewal receiver, then called `Channel._register_for_renewal("queue_a", msg)`. Result: `_kombu_renewal_receiver` set on the message and `pickle.dumps` → `TypeError: cannot pickle '_thread.lock' object`. Same setup on this branch: attribute absent, `pickle.loads(pickle.dumps(msg))` round-trips.

**Retry path with the real renewer:** after `_register_for_renewal`, drained the task the real `AutoLockRenewer` queued — `(receiver, renewable, start, duration, on_lock_renew_failure, override)` — confirmed the queued receiver is the renewal receiver and the callback is the `partial`. Swapped `connection._renewer` for a recorder and invoked the failure path (`channel._on_lock_renew_failure(renewal_receiver, msg2, Exception("transient"))`): re-registers exactly once with the same renewal receiver, the same message, and passes `on_lock_renew_failure=` again. So the retry chain survives the removal of the message attribute.

**Version-floor question from @MehrazRumman's review:** `inspect.signature(AutoLockRenewer.register)` at `7.14.3` → `['self', 'receiver', 'renewable', 'max_lock_renewal_duration', 'on_lock_renew_failure']` — the per-registration kwarg exists at the pinned floor, and since the pin is `==7.14.3` there is no older version to worry about.

**Regression test matrix:** the new `test_register_for_renewal_keeps_message_pickleable` fails on unmodified `main` (transport code from `main` + test file from this branch → `assert not hasattr(message, "_kombu_renewal_receiver")` → `assert not True`) and passes here. Full suite on this branch: `python -bb -m pytest -q t/unit/transport/test_azureservicebus.py` → **55 passed**.

No residual writes of `_kombu_renewal_receiver` anywhere in `kombu/` on this branch.

On the integration-test ask: `t/integration/` has no azure service bus harness at all (only kafka/mongodb/pgmq/py-amqp/redis/sqs), so that would mean standing up a new integration module with live Azure credentials rather than extending an existing one — possibly better scoped as a follow-up than a blocker for this fix.

auvipy:
what about improving unit test coverage?

auvipy (approved):

PR #2657, opened 2026-09-18
Reject bool for pool limit

Pitchfork-and-Torch:
## Summary
`bool` subclasses `int`, so `set_limit(True)` / `Pool(limit=True)` previously silently became pool size 1 (and `set_limit(False)` became 0 via `limit or 0`). Reject `bool` explicitly with `TypeError`.

## Test plan
- [x] `t/unit/test_pools.py::test_bool_limit` passes (2 tests)


auvipy:
@GangEunzzang can you please try this and verify that it doesnt regress?

GangEunzzang (commented):
One thing: `resize()` and the `limit` setter still accept a bool, so `pool.limit = True` gets through. Otherwise looks good, Celery only passes ints here.

auvipy:
> One thing: `resize()` and the `limit` setter still accept a bool, so `pool.limit = True` gets through. Otherwise looks good, Celery only passes ints here.

@Pitchfork-and-Torch please cross check the suggestion

Pitchfork-and-Torch:
Checked. resize() and the limit setter still accepted bool. resize() now rejects bool, and the setter calls resize(), so pool.limit = True raises TypeError the same way set_limit() and Resource(limit=True) do. Int limits are unchanged.

auvipy (approved):

PR #2665, opened 2026-09-21
Keep global QoS semantics for RabbitMQ 4.x

GangEunzzang:
#2481 made `qos_semantics_matches_spec()` return `True` for RabbitMQ 4.0 and later, on the assumption that 4.0 removed global QoS from classic queues. It didn't: `basic.qos(global=True)` followed by `basic.consume` still succeeds on a classic queue on RabbitMQ 4.3.5, and the docs only list quorum queues as rejecting it.

The effect on Celery: `qos_global` becomes `False` on every RabbitMQ 4 broker, so the prefetch-count reduction after a connection restart is skipped (celery/celery#9512 logs "Skipping prefetch count reduction ... per-consumer QoS (apply_global=False)"). That is what turned the `t/smoke/tests/test_consumer.py` prefetch tests red on Celery main after the 5.7.0a1 bump: they pass with Kombu 5.6.2 on the same broker image, and they pass again with this branch installed in the worker image. Celery already sets `qos_global=False` itself when it detects quorum queues, so consumers on quorum queues keep working without this heuristic.

This goes back to the version boundary from the spec: per-channel below 3.3, global semantics from 3.3 on. The #2481 unit tests are adjusted accordingly.

Discussion: celery/celery#10679.


auvipy:
tagging it as regression

auvipy (approved):

auvipy:
can you check if anything else is needed to be changed in kombu?

GangEunzzang:
@auvipy Please hold off merging this. I kept digging for the "anything else in Kombu" question and the premise doesn't hold for 4.3+:

- RabbitMQ 4.3.0 moved `global_qos` from `permitted_by_default` to `denied_by_default` (4.3.0 release notes). 4.0–4.2 still permit it.
- When it's denied the broker doesn't reject `basic.qos(global=true)`; `rabbit_channel.erl` rewrites it to `global=false` and only logs. Nothing in `server_properties` tells the client which mode it's in.
- Ran a real worker against 4.3.5 with this behaviour (kombu 5.6.2), 2 tasks running, forced a reconnect: the worker logs "Temporarily reducing the prefetch count to 2", the broker sets the new consumer to 2, and it stays at 2 after the tasks finish; when later tasks make the worker log "Prefetch count has been restored to the maximum of 4", the broker's consumer is still at 2. That's the stuck-after-reconnect bug #9512 fixed. With current main the worker skips the reduction and the broker stays at 4. The Celery smoke tests only assert the log lines, which is why they went green here.
- On 4.0–4.2, where global QoS is still permitted, this branch brings back the 540 "does not support global qos" from #2469 for quorum queues (reproduced on 4.2.9).

So #2481's per-consumer default for 4.x is right and Kombu needs nothing here. What needs changing is on the Celery side: the smoke tests, and the prefetch reduction itself if it should keep working on RabbitMQ 4. I'll close this and follow up on celery/celery#10679.

PR #2669, opened 2026-09-22
Upgrade grpcio and protobuf dependencies

auvipy:
Update grpcio and protobuf versions in requirements.


auvipy (commented):
going to merge this to test with celery 5.7 alpha and betas

